PR #33, opened 2016-06-10
Add Carthage support

dirkoswanepoel:


JKalash:
OneSignal SDK supports carthage as of 2.0

PR #35, opened 2016-06-15
Fix issue when additional data is absent and launch URL is included.

bernikowich:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/35)


bernikowich:
There is case when payload has path `/custom/u/`, but has no `/custom/a/`. This tiny commit will fix it.

PR #56, opened 2016-08-16
2.0.9

JKalash:
- Updated OneSignal.h header explanation for the singleton class
- FIX: No more usableDeviceToken method. Explanation: This method used
  to return null thus not registering any push token if the user did not
grant push permissions. Changed since dev can still send remoteSilent
notifications. This also fixed the no push token issue developers where
experiencing.
- FIX: Remove call to registerForPushNotifications without prompt in
  init when autoPrompt set to false. Only communicate with APNS once dev
calls registerForPushNotifications API. Fixes ambiguity when passing
kOSSettingsKeyAutoPrompt : @NO.
- FIX: Ad Hoc profile no longer need to restart app to properly register
  token
- FIX: Non-initializing the OneSignal SDK no longer blocks swizzled
  UIApplicationDelegate methods from getting called

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/56)

### DIFF
--- a/OneSignal.podspec
+++ b/OneSignal.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "OneSignal"
-  s.version          = "2.0.8"
+  s.version          = "2.0.9"
   s.summary          = "OneSignal push notification library for mobile apps."
   s.homepage         = "https://onesignal.com"
   s.license          = { :type => 'MIT', :file => 'LICENSE' }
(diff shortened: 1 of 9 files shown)


gdeglin:
Revised update notes:

Added links to OneSignal's setup instructions in OneSignal.h.

OneSignal will now correctly fetch a push token even if users have not granted push permission. The push token will now be fetched anyway since it can still be used to send remote silent notifications. This also fixes the "No push token" message on the OneSignal dashboard.

Fix: Apps built with an AdHoc profile no longer need to be restarted to properly register for notifications.

Fix: Not initializing the OneSignal SDK no longer blocks swizzled UIApplicationDelegate methods from getting called

jkasten2:
Reviewed 6 of 9 files at r1, 5 of 5 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/onesignal/onesignal-ios-sdk/56)*

PR #59, opened 2016-08-19
2.0.10

JKalash:
- Swizzling checks for all AppDelegate subclasses to find which proper class to inject to. SDK makes sure the subclass overrides the required method and does not simply inherit it from a superclass.
- SDK parses info.plist to find out if it should request location when in use or always, or log an error if neither description entries are in the property list file.
- Changed CLLocationManagerDelegate from OneSignal’s private shared instance to the existing OneSignalLocation private singleton.
- Time-based location tracking. SDK sends user’s initial location then assigns a refresh rate of 5 min (foreground) or 10min (background).

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/59)

### DIFF
--- a/OneSignal.podspec
+++ b/OneSignal.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "OneSignal"
-  s.version          = "2.0.9"
+  s.version          = "2.0.10"
   s.summary          = "OneSignal push notification library for mobile apps."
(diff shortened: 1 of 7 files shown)


jkasten2:
Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/onesignal/onesignal-ios-sdk/59)*

Issue #68, opened 2016-09-02
Crash iOS 8.3 at [OneSignalLocation internalGetLocation]


jrtapiaO:
Hello. Can you check this?
Thanks.
"CLLocationManager.allowsBackgroundLocationUpdates" is Available in iOS 9.0 and later.

 *** Terminating app due to uncaught exception 'NSUnknownKeyException', reason: '[<CLLocationManager 0x7e5638e0> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key allowsBackgroundLocationUpdates.'
*** First throw call stack:
(
	0   CoreFoundation                      0x02aa4746 __exceptionPreprocess + 182
	1   libobjc.A.dylib                     0x02657a97 objc_exception_throw + 44
	2   CoreFoundation                      0x02aa4361 -[NSException raise] + 17
	3   Foundation                          0x00cdc1ee -[NSObject(NSKeyValueCoding) setValue:forUndefinedKey:] + 282
	4   Foundation                          0x00c39608 _NSSetUsingKeyValueSetter + 115
	5   Foundation                          0x00c3958d -[NSObject(NSKeyValueCoding) setValue:forKey:] + 267
	6   App                      0x00219cae +[OneSignalLocation internalGetLocation:] + 910
	7   App                      0x00219494 __33+[OneSignalLocation getLocation:]_block_invoke + 68

PR #71, opened 2016-09-02
2.1.1

JKalash:
- iOS 10 fixes:
    - In-app alerts will no longer show up alongside native notifications when InAppAlerts are enabled.
    - Changed the way the OneSignal SDK extracts attachments from the notification payload.
- Added a mutableContent property to the OSNotificationPayload for iOS 10.
- Remote silent notifications will no longer show an alert in an app that is in focus when InAppAlerts are enabled.
- Dismiss the in-app webview when coming back to the app after receiving a deep linked URL and kOSSettingsInAppLaunchURL is true.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/71)

### DIFF
--- a/OneSignal.podspec
+++ b/OneSignal.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "OneSignal"
-  s.version          = "2.1.0"
+  s.version          = "2.1.1"
   s.summary          = "OneSignal push notification library for mobile apps."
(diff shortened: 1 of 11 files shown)


jrtapiaO:
Hello. I saw you merged the issue #68 into #70 but i still have the crash withe the v2.1.1. Can you take a look? Thanks. 👍

PR #86, opened 2016-09-20
[OneSignalMobileProvision] Revert check for wildcard and ad hoc provi…

bogren:
…sioning profile

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/86)


bogren:
Why did you remove the check for `ad-hoc` and `wildcard` profiles?

JKalash:
Hi there,

As of Xcode 8 / iOS 10, we check the user's entitlements rather than the provisioning profile itself to this matter.

Thanks

Issue #144, opened 2016-11-27
crash OneSignalHelper.m line 719


malkabani:
hi
i have app with latest onesignal version 2.2.3
and when try to send notification from dashboard exeption appears

OneSignalHelper.m line 719
+[OneSignalHelper enqueueRequest:onSuccess:onFailure:isSynchronous:]

can you check it and help me please
thanks

jkasten2:
@malkabani Do you have `kOSSettingsKeyInFocusDisplayOption` set in your code? Could provide the iOS version you are testing on? Was your app in focus when you receive the notification or does it only crash after you open the notification from the notification center.

Lastly can your provide the full stack trace?

Thanks.

anurlybayev:
```
Crashed: com.apple.main-thread
0  CoreFoundation                 0x18933a6d0 CFRelease + 120
1  CFNetwork                      0x189b44010 <redacted> + 224
2  500px                          0x10028d3e0 +[OneSignalHelper enqueueRequest:onSuccess:onFailure:isSynchronous:] (OneSignalHelper.m:719)
3  500px                          0x10028d05c +[OneSignalHelper enqueueRequest:onSuccess:onFailure:] (OneSignalHelper.m:702)
4  500px                          0x10029ca78 +[OneSignal submitNotificationOpened:] (OneSignal.m:913)
5  500px                          0x10029c0d4 +[OneSignal notificationOpened:isActive:] (OneSignal.m:835)
6  500px                          0x10028f7c0 -[swizzleUNUserNotif onesignalUserNotificationCenter:willPresentNotification:withCompletionHandler:] (UNUserNotificationCenter+OneSignal.m:111)
7  UIKit                          0x18f5b221c <redacted> + 3892
8  UIKit                          0x18f5a00cc <redacted> + 132
9  FrontBoardServices             0x18b03b92c <redacted> + 36
10 FrontBoardServices             0x18b03b798 <redacted> + 176
11 FrontBoardServices             0x18b03bb40 <redacted> + 56
12 CoreFoundation                 0x189412b5c <redacted> + 24
13 CoreFoundation                 0x1894124a4 <redacted> + 524
14 CoreFoundation                 0x1894100a4 <redacted> + 804
15 CoreFoundation                 0x18933e2b8 CFRunLoopRunSpecific + 444
16 GraphicsServices               0x18adf2198 GSEventRunModal + 180
17 UIKit                          0x18f385894 <redacted> + 684
[...]

PR #156, opened 2016-12-09
Prevent OneSignal specific logic if SDK is not initialized

anurlybayev:
* Related issue: https://github.com/OneSignal/OneSignal-iOS-SDK/issues/144
* Num of :+1: needed: 1
* Code Reviewers: @jkasten2, @JKalash 
* Copy: @500px/ios

### Summary

We're in the process of moving away from Parse to OneSignal. However, as we started deploying push notifications with OneSignal SDK we've started seeing crashes which are described in https://github.com/OneSignal/OneSignal-iOS-SDK/issues/144.

The way our app behaves is as follows:
We have a rollout flag `one_signal`, if on app start it is set to true then we call OneSignal SDK's
```Obj-C
+ (id)initWithLaunchOptions:(NSDictionary*)launchOptions
                      appId:(NSString*)appId
 handleNotificationReceived:(OSHandleNotificationReceivedBlock)receivedCallback
   handleNotificationAction:(OSHandleNotificationActionBlock)actionCallback
                   settings:(NSDictionary*)settings
```
if the flag value is false, we fallback to registering with Parse. However, because 
```Obj-C
+ [UIApplication (OneSignal) load]
```
does method swizzling on `UIApplication` load all of the app delegate messages are handled by OneSignal SDK and some of them call system delegate methods that we implement as they being called by the OS.

So if we enable `one_signal` flag and then disable it, the device is registered with two systems. The way our backend is written it will send push request to both Parse and OneSignal which in practice will generate duplicate notifications. The crash happens, when we try to handle OneSignal notification while not registered with OneSignal, meaning without ever calling
```Obj-C
+ (id)initWithLaunchOptions:(NSDictionary*)launchOptions
                      appId:(NSString*)appId
 handleNotificationReceived:(OSHandleNotificationReceivedBlock)receivedCallback
   handleNotificationAction:(OSHandleNotificationActionBlock)actionCallback
                   settings:(NSDictionary*)settings
```
thus `httpClient` is `nil` which doesn't create the `NSURLRequest` and we pass `nil` to `NSURLConnection`.

The temporary patch not to do OneSignal fancy stuff, unless SDK has been properly initialized.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/156)

### DIFF
--- a/iOS_SDK/OneSignal/UNUserNotificationCenter+OneSignal.m
+++ b/iOS_SDK/OneSignal/UNUserNotificationCenter+OneSignal.m
@@ -92,7 +92,7 @@ static NSArray* delegateUNSubclasses = nil;
                   withCompletionHandler:(void (^)(UNNotificationPresentationOptions options))completionHandler {
     // If OneSignal has not been initialized there is no reason to proceed with all the code below this if statement
     // since delaying method swizzling past + [UIApplication load] is dangerous this check is in place.
-    // For more details see: https://github.com/OneSignal/OneSignal-iOS-SDK/pull/155
+    // For more details see: https://github.com/OneSignal/OneSignal-iOS-SDK/pull/156
     if (![OneSignal app_id]) {
         [swizzleUNUserNotif callLegacyAppDeletegateSelector:notification
                                                 isTextReply:false
--- a/iOS_SDK/OneSignal/UNUserNotificationCenter+OneSignal.m
+++ b/iOS_SDK/OneSignal/UNUserNotificationCenter+OneSignal.m
@@ -90,6 +90,19 @@ static NSArray* delegateUNSubclasses = nil;
 - (void)onesignalUserNotificationCenter:(UNUserNotificationCenter *)center
                 willPresentNotification:(UNNotification *)notification
                   withCompletionHandler:(void (^)(UNNotificationPresentationOptions options))completionHandler {
+    // If OneSignal has not been initialized there is no reason to proceed with all the code below this if statement
+    // since delaying method swizzling past + [UIApplication load] is dangerous this check is in place.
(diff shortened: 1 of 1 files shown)


jkasten2:
@anurlybayev Thanks for digging into this. I didn't see `callLegacyAppDeletegateSelector` in the call stack in the #144 issue. The appId check should instead be around [`notificationOpened:isActive`](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/2.3.3/iOS_SDK/OneSignal/UNUserNotificationCenter%2BOneSignal.m#L110) above your changes.

`callLegacyAppDeletegateSelector` needs to be kept regardless if OneSignal is initialized otherwise any notification selector on your AppDelegate will no longer be called. `[OneSignal app_id]` is already checked in [oneSignalRemoteSilentNotification:UserInfo:fetchCompletionHandler:](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/cc7ae438bfd9cb325b63b4f0ae9771751ed4c68f/iOS_SDK/OneSignal/UIApplicationDelegate%2BOneSignal.m#L194-L199) so it is ok if `callLegacyAppDeletegateSelector` fires without any additional checks to it.

jkasten2:
Going make a commit to ensure `httpClient` is initialized from swizzling as well as `initWithLaunchOptions`. In-addition will add checks for a `nil` check of `request`  in `enqueueRequest:onSuccess:onFailure:isSynchronous:` and it's body to ensure this can't attempt to make an invalid network request that could throw.

jkasten2:
Added `nil` checks for `app_id` since this is lower on the stack. This was added in commit c184c4de82e32a2d82a4cb7857d5485b8942493e

PR #189, opened 2017-02-21
tvOS Support [Needs Implementation]

matteocrippa:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/189)

### DIFF
--- a/OneSignal.podspec
+++ b/OneSignal.podspec
@@ -8,7 +8,8 @@ Pod::Spec.new do |s|
   
   s.source           = { :git => "https://github.com/OneSignal/OneSignal-iOS-SDK.git", :tag => s.version.to_s }
   
-  s.platform     = :ios
+  s.ios.deployment_target = '9.0'
+  s.tvos.deployment_target = '9.0'
   s.requires_arc = true
   
   s.ios.vendored_frameworks = 'iOS_SDK/Framework/OneSignal.framework'
--- a/OneSignal.podspec
+++ b/OneSignal.podspec
@@ -13,5 +13,6 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   
   s.ios.vendored_frameworks = 'iOS_SDK/Framework/OneSignal.framework'
+  s.tvos.vendored_frameworks = 'iOS_SDK/Framework/OneSignal.framework'
   s.framework               = 'SystemConfiguration'
 end


matteocrippa:
it is needed to add support to library to tvOS too in order to have this working

jkasten2:
@matteocrippa Sorry about the delay, thanks for the pull request. We haven't tested our SDK with tvOS yet so we haven't added this to our `podspec`.

In your testing is push working for you?

matteocrippa:
@jkasten2 nope I'm stuck, I noticed that the lib lacks of the AppleTV support, if you can add that when you compile your library it should work fine now.

jkasten2:
@matteocrippa  We don't have an Apple TV to test this one. Could you use your fork with cocopods to see if OneSignal is working on your Apple TV after the `podspec` change?

You can use the following in your `Podfile` to try your changes.
```
pod 'OneSignal', :git => 'https://github.com/matteocrippa/OneSignal-iOS-SDK.git'
```

matteocrippa:
@jkasten2 I confirm is not working here the issue that raises

```

Showing Recent Issues
ld: in /Pods/OneSignal/iOS_SDK/Framework/OneSignal.framework/OneSignal(OneSignalAlertViewDelegate.o), building for tvOS, but linking in object file built for iOS, for architecture arm64
```

so the issue is with the library file, you should add `tvOS` as supported architecture in order to proceed

jkasten2:
@matteocrippa 
I see, could you test this with your fork? Open `/Pods/OneSignal/iOS_SDKOneSignal.xcodeproj`, add the `tvOS` architecture, then build the SDK.

You might have to also add `tvOS` to the Run Script that creates the framework.
image

You can the use a local directly as a pod.
```
pod 'OneSignal', :local => '/Pods/OneSignal'
```

If you get stuck I can try to make a branch for you to test with.

matteocrippa:
@jkasten2 I had a quick check, but I think the API is now using some functions that are not available in tvOS (e.g. `currentUserNotificationSettings`) so it fails to compile.

I tried to manage to tweak a little using device check, but it is quite huge to fix

jkasten2:
@matteocrippa Thanks for the details, thinking the best approach to this would be to create another build target and exclude specific source files from the build. Some code will need to be moved into different or new `.m` files to accomplish this however.

matteocrippa:
@jkasten2 I do think so, I managed in the meanwhile to create an `Alamofire` based wrapper to handle the network calls to Onesignal SDK and it works.
I had to do that in order to slip the UI layer that seems to be the one that is providing issue in having the tvOS lib to compile (most of the functions are not available on tvOS)

jkasten2:
@matteocrippa Good to hear you were able to get OneSignal working with REST API calls.

We would like our SDK to be as compatible as possible but we haven't had to much demand for tvOS support. I'll update this thread if we decide to start work on implementing this.

IBacklundKuma:
Apple TV would also be useful for our project.

valeriologiudice:
Hello, is this ever going to be done. Use OneSignal on tvOS would help me very much. Best regards.

matteocrippa:
Gl :package:

valeriologiudice:
> Gl 📦

What does it mean @matteocrippa? 😄

matteocrippa:
Cit. "Lasciate ogni speranza o voi che entrate"

Wizfinger:
Is there any hope for this feature? I need it.

Issue #208, opened 2017-03-16
expose OneSignal SDK version


jamesstout:
Hi 
It would be nice if there was a way to retrieve the SDK version from the SDK. At the moment I have a hack to scan the pod lockfile for the version, put that in my info.plist as a build phase and the read the version during execution.

Thanks
James

jkasten2:
@jamesstout Closing this issue since you now have a pull request open.

PR #209, opened 2017-03-16
Sdk version

jamesstout:
Per my issue #208, I knocked this up. Of course you could have two consts and expose them more simply, this way means you just need to keep the one const.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/209)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -177,18 +177,13 @@ NSObject<OneSignalNotificationSettings>* osNotificationSettings;
 + (NSString*)sdk_semantic_version {
 
 	// examples:
-	// ONESIGNAL_VERSION = @"020402"
-	// ONESIGNAL_VERSION = @"020116"
+	// ONESIGNAL_VERSION = @"020402" returns 2.4.2
+	// ONESIGNAL_VERSION = @"001000" returns 0.10.0
 	// so that's 6 digits, where the first two are the major version
 	// the second two are the minor version and that last two, the patch.
 	// c.f. http://semver.org/
 
-	if(ONESIGNAL_VERSION.length != 6){
-		return @"UNKNOWN";
-	}
-	else{
-		return [ONESIGNAL_VERSION one_getSemanticVersion];
-	}
+	return [ONESIGNAL_VERSION one_getSemanticVersion];
 }
 
 + (NSString*)mUserId {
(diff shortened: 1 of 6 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -144,6 +145,27 @@ + (NSString*)app_id {
...
++ (NSString*)sdk_semantic_version {
+	
+	// examples:
+	// ONESIGNAL_VERSION = @"020402"
+	// ONESIGNAL_VERSION = @"020116"
+	// so that's 6 digits, where the first two are the major version
+	// the second two are the minor version and that last two, the patch.
+	// c.f. http://semver.org/
+	
+	if(ONESIGNAL_VERSION.length != 6){
```
jkasten2: Could we remove the length check here? We don't plan to move off a 6 digit semantic version number and this would be better checked in a unit test.

iOS_SDK/OneSignalSDK/UnitTests/UnitTests.m
```diff
@@ -582,6 +584,87 @@ - (void)testBasicInitTest {
...
+							   @"020303" : @"2.3.3",
+							   @"020304" : @"2.3.4",
+							   @"020305" : @"2.3.5",
+							   @"020306" : @"2.3.6",
+							   @"020307" : @"2.3.7",
+							   @"020400" : @"2.4.0",
+							   @"000400" : @"0.4.0",
+							   @"000000" : @"0.0.0",
+							   @"020401" : @"2.4.1",
+							   @"020402" : @"2.4.2"};
```
jkasten2: Could we limit this dictionary to just a few unique lengths?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
jamesstout: Have moved to a unit test

iOS_SDK/OneSignalSDK/UnitTests/UnitTests.m
jamesstout: We could do. I just grepped the git reflog for all historical values of that const and checked the method worked on all previous values. They are all the same length, what's being tested is that the method works when the substrings start with a zero or not. If we change the lengths in this test, they will fail as the method no longer checks the length.

jkasten2:
@jamesstout Could you rebase your branch? I added some files to the `OneSignalSDK/OneSignal.xcodeproj` that is now getting a merge conflict.

jamesstout:
Branch rebased. Had to fix some conflicts, hence the new commits.

jkasten2:
@jamesstout Thanks, no problem, squashed your commits into one for a clean merge into master.

jamesstout:
Thanks v much

PR #244, opened 2017-05-30
Address two Clang Static Analyzer warnings

beltex:
Hello OneSignal!

### What's in this pull request?

Addressed two warnings via the Clang Static Analyzer:

- Added missing call to `super.viewDidLoad()`
- Added missing `CFRelease` call to prevent possible mem leak

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/244)


jkasten2:
@beltex Thanks for the warning fixes!

PR #254, opened 2017-07-05
Fixed race conditions in UNUserNotificationCenter category

fluidsonic:
# DO NOT MERGE YET
I have identified two issues with the current implementation of `onesignalGetNotificationSettingsWithCompletionHandler:` and `onesignalRequestAuthorizationWithOptions:completionHandler:`:

1. The documentation of [`getNotificationSettings(completionHandler:)`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649524-getnotificationsettings) states that `completionHandler` is called asynchronously but the SDK calls it synchronously [in one case](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/2.5.4/iOS_SDK/OneSignalSDK/Source/UNUserNotificationCenter%2BOneSignal.m#L107).
1. The documentation of [`getNotificationSettings(completionHandler:)`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649524-getnotificationsettings) states that `completionHandler` is called on an arbitrary thread. In addition the documentation of [`UNNotificationCenter`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter#2851211) states the shared notification center can be used from any thread.

The first issue results in unexpected behavior for anyone using `UNNotificationCenter.getNotificationSettings(completionHandler:)`.

The second issue causes [`-onesignalGetNotificationSettingsWithCompletionHandler:`](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/2.5.4/iOS_SDK/OneSignalSDK/Source/UNUserNotificationCenter%2BOneSignal.m#L105) and [`-onesignalRequestAuthorizationWithOptions:completionHandler:`](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/2.5.4/iOS_SDK/OneSignalSDK/Source/UNUserNotificationCenter%2BOneSignal.m#L91) to be highly prone to race conditions.

I've received several logs for **crashes during app startup** with a backtrace like this, which drew my attention:

```
SEGV_ACCERR

0   libobjc.A.dylib   objc_retain + 16
1   App               reabstraction thunk helper from @callee_owned (@owned __ObjC.UNNotificationSettings) -> () to @callee_unowned @convention(block) (@unowned __ObjC.UNNotificationSettings) -> () (StandardApiManager.swift:0)
2   OneSignal         -[OneSignalUNUserNotificationCenter onesignalGetNotificationSettingsWithCompletionHandler:] (UNUserNotificationCenter+OneSignal.m:107)
3   App               App.StandardApiManager.onStart () -> () (StandardApiManager.swift:507)
```

## To-Do
This pull request makes sure that modifications to the SDK's state always happen on the main thread.

When running the Unit Tests though another issue comes up: **plenty of deadlocks**

For some reason, `-[UNNotificationCenter getNotificationSettingsWithCompletionHandler:]` is [forcibly called **synchronously**](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/2.5.4/iOS_SDK/OneSignalSDK/Source/OneSignalNotificationSettingsIOS10.m#L81) from the main thread during SDK initialization. It's called **several times and sometimes even recursively** (call → completion handler → another call → completion handler → etc.)!

This is a very bad approach:

* The current implementation assumes that the `completionHandler` is always called on a background thread but the documentation states that his is not guaranteed.
→ **Whenever the `completionHandler` is called on the main thread (which is legal), the SDK will deadlock the app on app startup!**
* There is a reason the method is asynchronous. By converting it to a synchronous function and then calling it _multiple times_ **the app start can take longer than necessary**.
* It's quite difficult to understand the SDK's initialization routine due to many threads, semaphores and callbacks being involved & nested.

I gave up trying to make the SDK initialization properly thread-safe without using semaphores since this is no easy task, potentially requires a lot of changes and I don't have enough knowledge about the SDK's internals to ensure my implementation is correct.

Please take over from there.

Thanks,
Marc

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/254)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/UNUserNotificationCenter+OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/UNUserNotificationCenter+OneSignal.m
@@ -89,31 +89,39 @@ static BOOL useCachedUNNotificationSettings;
 static UNNotificationSettings* cachedUNNotificationSettings;
 
 - (void)onesignalRequestAuthorizationWithOptions:(UNAuthorizationOptions)options completionHandler:(void (^)(BOOL granted, NSError *__nullable error))completionHandler {
-    OneSignal.currentPermissionState.hasPrompted = true;
-    
-    useCachedUNNotificationSettings = true;
-    id wrapperBlock = ^(BOOL granted, NSError* error) {
-        useCachedUNNotificationSettings = false;
-        OneSignal.currentPermissionState.accepted = granted;
-        OneSignal.currentPermissionState.answeredPrompt = true;
-        completionHandler(granted, error);
-    };
-    
-    [self onesignalRequestAuthorizationWithOptions:options completionHandler:wrapperBlock];
+    [OneSignalHelper dispatch_async_on_main_queue:^{
+        OneSignal.currentPermissionState.hasPrompted = true;
+        
+        useCachedUNNotificationSettings = true;
+        id wrapperBlock = ^(BOOL granted, NSError* error) {
+            [OneSignalHelper dispatch_async_on_main_queue:^{
+                useCachedUNNotificationSettings = false;
+                OneSignal.currentPermissionState.accepted = granted;
+                OneSignal.currentPermissionState.answeredPrompt = true;
+                completionHandler(granted, error);
+            }];
+        };
+        
+        [self onesignalRequestAuthorizationWithOptions:options completionHandler:wrapperBlock];
+    }];
 }
 
 - (void)onesignalGetNotificationSettingsWithCompletionHandler:(void(^)(UNNotificationSettings *settings))completionHandler {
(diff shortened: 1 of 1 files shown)


jkasten2:
@fluidsonic Marc, thanks for looking into this and providing all the details you have. Proprieties on the main OneSignal class in OneSignal.m are loaded lazily which makes the stack hard to follow. Lazy loading was designed to make the SDK more robust to changes however due to the underlying queues and waits with semaphores it is actually currently very fragile to changes on iOS 10. Accessing UNNotificationCenter outside of the OneSignal SDK can create locks depending on when it is executed due to UNNotificationCenter as this class should not be simultaneously accessed from 2 thread as you have linked to.

If you have any test cases that are reproduce able on your fork or from master please add them and I can make sure they are fixed after any changes I make.

PR #259, opened 2017-08-01
Remove +[UIApplication load], Remove -[UIApplication setDelegate:] swizzle

alexhillc:
* Remove `+[UIApplication load]` in OneSignal class implementation in order to prevent swizzling before the library is instantiated (IF it is instantiated at all)
* Remove `-[UIApplication setDelegate:]` swizzle in favor of manually calling `+[OneSignal swizzleSelectors]` during init, which kicks off all required method swizzling. This method also wraps the sensitive code in a `dispatch_once` block in order to prevent multiple swizzles from happening if init is called more than once

The use cases for making these changes:

* Prevent swizzling if the SDK is not going to be instantiated during the app lifecycle
* Ties swizzle methods to the init method rather than depending on the swizzled `-[UIApplication setDelegate:]`. This allows the SDK to lie "dormant" while not in use

Not sure if this will be appreciated, but I figured if I was going to make changes, I may as well PR them!

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/259)


jkasten2:
@alexhillc Thanks for sharing the changes. Some of our wrapping SDKs for other frameworks such as Unity, Cordova, Xamarin, etc might be effected by this. Before merging this in it will need to be tested for these cases. We may need a flag to disabling sizzling with a flag instead depending if this effects them.

PR #263, opened 2017-08-13
add 64-bit support

Hendrik44:
Because of issues with 64-bit support and submitting to iTunes-Connect, i fixed the architecture to default/standard arch. This commit will fix it

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/263)


jkasten2:
@Hendrik44 Thanks for the pull request. This added the `ARCHS = "$(ARCHS_STANDARD)";` flag to `/tmp/OneSignal.dst` and `com.onesignal.OneSignal-Dynamic`. This missing flag on these objects could have had an effect if using Carthage or if the SDK project itself was directly added to your Xcode workspace. 

The main `OneSignal.framework` object already had this flag so projects using Cocoapods or directly including the `OneSignal.framework` into their project already have 64 bit set correctly.

PR #290, opened 2017-10-18
Feature/swift compatibilty nullable attribute

leverdeterre:
Hi, 
i have a issue with an implicit unwrap property.

to avoid a crash on the first start of my app, i must convert userId to a optional (String?)
Xcode show me a warning become this property is not well documented has a nullable one.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/290)

### DIFF
--- a/Examples/ObjectiveCExample/Podfile
+++ b/Examples/ObjectiveCExample/Podfile
@@ -1,7 +1,9 @@
+platform :ios, '10.3'
+
 target 'OneSignalDemo' do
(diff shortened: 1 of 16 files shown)


jkasten2:
@leverdeterre Thanks for the PR. This has been fixed in the 3.0.0 release:
https://github.com/OneSignal/OneSignal-iOS-SDK/blob/3.0.0/iOS_SDK/OneSignalSDK/Source/OneSignal.h#L288

PR #297, opened 2017-11-02
Give user the ability to disable onesignal load

depoon:
Users of one signal (including me) are experiencing crash in Simulator
mode. This commit allows users the option to turn off oneSignal
functionality by using NSProcessInfo arguments.

To disable, use “DISABLE_ONESIGNAL_SWIZZLING” in “arguments passed on launch” in
the scheme setup

Screenshot 2021-03-05 at 10 29 12 AM

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/297)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -1595,6 +1595,11 @@ static NSString *_lastnonActiveMessageId;
 @implementation UIApplication (OneSignal)
 #define SYSTEM_VERSION_LESS_THAN_OR_EQUAL_TO(v)     ([[[UIDevice currentDevice] systemVersion] compare:v options:NSNumericSearch] != NSOrderedDescending)
 + (void)load {
+    
+    if ([self shouldDisableBasedOnProcessArguments]) {
+        [OneSignal onesignal_Log:ONE_S_LL_WARN message:@"OneSignal method swizzling is disabled. Make sure the feature is enabled for production."];
+        return;
+    }
     [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:@"UIApplication(OneSignal) LOADED!"];
     
     // Prevent Xcode storyboard rendering process from crashing with custom IBDesignable Views
@@ -1626,6 +1631,12 @@ static NSString *_lastnonActiveMessageId;
     [OneSignalHelper registerAsUNNotificationCenterDelegate];
 }
 
++(BOOL) shouldDisableBasedOnProcessArguments {
+    if ([NSProcessInfo.processInfo.arguments containsObject: @"DISABLE_ONESIGNAL_SWIZZLING"]) {
+        return YES;
+    }
+    return NO;
+}
 @end
 
 


pouty18:
@emawby can we get this PR merged in or is there anything else you'd like us to add? This is holding up a large portion of our unit-testing bundle. Thanks!

amadeu01:
Why this PR cannot be merged?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1626,6 +1630,12 @@ + (void)load {
     [OneSignalHelper registerAsUNNotificationCenterDelegate];
 }
 
++(BOOL) shouldDisableBasedOnProcessArguments {
+    if ([[[NSProcessInfo processInfo]arguments]containsObject: @"DISABLE_ONESIGNAL"]){
```
jkasten2, line 1634: nit, would be more readable in this format
```Objective-C
if ([NSProcessInfo.processInfo.arguments containsObject: @"DISABLE_ONESIGNAL"]) {
```

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1595,6 +1595,10 @@ + (UNMutableNotificationContent*)serviceExtensionTimeWillExpireRequest:(UNNotifi
 @implementation UIApplication (OneSignal)
 #define SYSTEM_VERSION_LESS_THAN_OR_EQUAL_TO(v)     ([[[UIDevice currentDevice] systemVersion] compare:v options:NSNumericSearch] != NSOrderedDescending)
 + (void)load {
+    
+    if ([self shouldDisableBasedOnProcessArguments]){
```
jkasten2, line 1599: nit, space between `){`

iOS_SDK/OneSignalSDK/Source/OneSignal.m
jkasten2, line 1634: The name `DISABLE_ONESIGNAL` implies it is completely disabling OneSignal. I thinking it needs to be a bit more verbose as it is only disabling swizzling.
@depoon What do you think of `DISABLE_ONESIGNAL_SWIZZLING`?

jkasten2 (changes requested):
Summary of feedback:
* 2 nits, not blockers but would be nice to clean up formatting.
* `DISABLE_ONESIGNAL` name doesn't reflect close enough to its behavior. Getting public names right is important as they are sticky until the next major SDK release.

depoon:
Wow... its been 4 years... where were you ... 😢 

Give me 1+ days to fix this PR... need to refresh myself abit

iOS_SDK/OneSignalSDK/Source/OneSignal.m
depoon: Lets go with your choice of name.

depoon: Yes. This looks nicer

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1626,6 +1630,12 @@ + (void)load {
     [OneSignalHelper registerAsUNNotificationCenterDelegate];
 }
 
++(BOOL) shouldDisableBasedOnProcessArguments {
+    if ([NSProcessInfo.processInfo.arguments containsObject: @"DISABLE_ONESIGNAL_SWIZZLING"]) {
```
jkasten2, line 1634: Is it possible for `processInfo` or `argument` to be `nil`? Possibly if you start the app from opening the app on the device instead of Xcode?

jkasten2, line 1634: We should check the value of `DISABLE_ONESIGNAL_SWIZZLING`, so if someone sets this to `false` it doesn't still disable it which would be confusing. Most people are going set and forget this flag but someone may come back and set this to false to debug something and it won't behave as expected.

jkasten2 (changes requested):
@depoon Awesome, appreciate the fixes here!
Sorry after taking a 2nd pass found two more things:
1. One possible `nil` handling we might have to do, depends on what the device does
2. Need to add a value check.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
depoon: This feature of disabling OneSignalSwizzling is dependent on injecting ProcessInfo variables. So it is reasonable that the `if condition` will become false when the processInfo.arguments is nil or empty.

depoon: hi @jkasten2,  `NSProcessInfo.processInfo.arguments` returns an array and not a dictionary (`NSProcessInfo.processInfo.environment`). The name of the value is self-explanatory, and its confusion to introduce double-negatives with `false`.

I have considered your concern about developers forgetting to remove to the argument. I have added a warning to alert developers.

jkasten2: Ah my mistake, you are right `arguments` is an array so presents is the correct way to do this.

Also warning looks good to me!

jkasten2: I wasn't sure on a real device when starting the app outside of Xcode if `processInfo` or `arguments` could be `nil` `containsObject` to throw. It seems like you have tested this but I'll give this a try as well to confirm as well.

jkasten2: I tested this on iOS 14.4 and this didn't create any issues when starting the app outside of Xcode.

jkasten2 (commented):
@depoon Everything looks good to me, thanks again!
I created a follow up PR for someone else at OneSignal to take a 2nd pass before we merge it into master.
https://github.com/OneSignal/OneSignal-iOS-SDK/pull/898

jkasten2:
This is now merged into master, it will be included in the next release.

Issue #278, opened 2017-09-19
Crash on iOS 11 with class_getSuperclass - SDK 2.5.4 through 2.5.6


aaronbrethorst:
This happens inconsistently, unfortunately. I imagine that turning this from a `do-while` to a straight-up `while` loop would fix the issue. You can check out the project where this is happening at https://github.com/OneBusAway/onebusaway-iphone/

<img width="765" alt="screen shot 2017-09-19 at 12 14 03 am" src="https://user-images.githubusercontent.com/2254/30579923-8ee05bda-9ccf-11e7-8c37-5bbde17fc49b.png">

anushkmittal:
I think it's the same issue as #274. @jkasten2 Can you reproduce the issue with this repo?

aaronbrethorst:
Yeah, this looks like the same issue to me. I'm happy to help out in any way I can.

jkasten2:
@aaronbrethorst Tried building and running the project you linked to above. Launched the app a number of times on iOS 11 on both a simulator and a real device and wasn't able to reproduce the issue. However I agree with your suggestion on preventing any possible nil errors.

The fix will be available in the next release.

aaronbrethorst:
Cool, thank you! Do you have an estimated date for release?

hixfield:
I have build against master but still see the issue popping up (same location as screenshot above). Maybe something went wrong with your push/merging to master?

jkasten2:
@hixfield I have verified the change is on master correctly.
https://github.com/OneSignal/OneSignal-iOS-SDK/blob/master/iOS_SDK/OneSignalSDK/Source/OneSignalSelectorHelpers.m#L103

Can do clean build to ensure you have the change in your project? If you are still seeing a crash can you verify the stack trace isn't different in anyway?

gerbiljames:
If you're using CocoaPods you can't just point it at the master branch in this repo because it will just pull the prebuilt framework down, which doesnt have the fix yet. I worked around this by forking the repo, compiling the framework and pushing it to the fork, then pointing CocoaPods at the fork. I assume it would be the same with other dependency managers too.

hixfield:
I am using cartage

hixfield:
This is the contents of my cart file:

github "OneSignal/OneSignal-iOS-SDK" "master"

I verified in the carthage/checkouts folder and can confirm that the source includes the fix.
[...]

PR #298, opened 2017-11-02
Hotfix random crash ios 11

gbtagliari:
I think it's related to issue #278 

The crash occurs in simulator when the project don't have Notification Service Extension enabled (or configured).

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/298)


jkasten2:
@gbtagliari Thanks for the pull, however I have implemented a fix in pull #301 and have thoroughly tested it.

PR #321, opened 2017-12-22
Fixes small memory leak, fixes buttons not working in iOS 7

Nightsd01:
• Fixes an issue where the app would (mistakenly) print a 'missing App ID parameter' whenever the Get iOS Params request was executed.
• Fixes an issue that caused buttons not to appear in iOS 7 push notification (in app)
• Fixes an issue that caused button press events in (in-app) push notifications not to be attached to the Action completion block correctly.
• Fixes a relatively small memory leak issue that caused 600 bytes to be leaked every time an HTTP request was made.
• Changes the OneSignalClient to use a single instance of NSURLSession instead of creating a new instance for every HTTP request.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/321)


Nightsd01:
Cancelling PR

PR #348, opened 2018-03-09
removed UNLegacyNotificationTrigger

by433:
due to app rejection - non-public APIs

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/348)


Nightsd01:
@by433 Thank you for this PR, we have made this change in #350 but have decided to do it in a slightly different way. Thank you for your contribution!

PR #364, opened 2018-04-19
Swizzling

Nightsd01:
• Ensures that UNNotificationCenter setDelegate: will never get called twice, which is useful if other SDK's are swizzling the same method incorrectly in a way that invites race conditions

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/364)


jkasten2:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/onesignal/onesignal-ios-sdk/364)*

PR #366, opened 2018-04-24
MIME Checking

Nightsd01:
• Adds the ability for the SDK to use attachments with no file extension
• Also adds the ability to specify the file type using the URL query parameter "file_type"
• If a media attachment URL does not contain a file extension, the SDK will now download the file and check the MIME type of the request to make sure it is a valid file type.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/366)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
+++ b/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
@@ -132,6 +132,10 @@
 		CA08FC801FE99B25004C445F /* Requests.m in Sources */ = {isa = PBXBuildFile; fileRef = CA08FC7D1FE99B25004C445F /* Requests.m */; };
 		CA08FC811FE99B25004C445F /* Requests.m in Sources */ = {isa = PBXBuildFile; fileRef = CA08FC7D1FE99B25004C445F /* Requests.m */; };
 		CA08FC871FE99BB4004C445F /* OneSignalClientOverrider.m in Sources */ = {isa = PBXBuildFile; fileRef = CA08FC831FE99BB4004C445F /* OneSignalClientOverrider.m */; };
+		CA36A42C208FDEFB003EFA9A /* NSURL+OneSignal.h in Headers */ = {isa = PBXBuildFile; fileRef = CA36A42A208FDEFB003EFA9A /* NSURL+OneSignal.h */; };
+		CA36A42D208FDEFB003EFA9A /* NSURL+OneSignal.m in Sources */ = {isa = PBXBuildFile; fileRef = CA36A42B208FDEFB003EFA9A /* NSURL+OneSignal.m */; };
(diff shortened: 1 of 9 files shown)


jkasten2:
We should add some additional tests for the priority order of the 3 extension type checks we are doing since this is getting more complex.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m, line 734 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/366#-LAtqg3O0y0_B80Txz_w:-LAtqg3O0y0_B80Txz_x:b-tx097m) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/b286a607df0e7d319f42e13b9ac067afe1e4e8a6/iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m#L734)):*
> ```objc
>     NSArray* components = [url componentsSeparatedByString:@"."];
>     
>     NSString* extension;
> ```

Lets add a comment here about what the order of priority of what extension is used since this code is getting more complex.

---

*[iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m, line 749 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/366#-LAtq25XBqwZVMsrwc2N:-LAtpSo05MrJhtiJ2n1i:bwki0v4) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/b286a607df0e7d319f42e13b9ac067afe1e4e8a6/iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m#L749)):*
> ```objc
>     
>     if (!extension)
>         extension = [[NSURL URLWithString:urlString] valueForQueryParameter:@"file_type"];
> ```

As a workaround we have documented adding `?filename=file.png` to the end of URLs that didn't have a file type.  We should grab the last query parameter here instead of `file_type` so it behaves as it did with the old versions of the SDK.

---

*[iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m, line 773 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/366#-LAtpkhlC8Eqeyiu_ADE:-LAtpkhlC8Eqeyiu_ADF:b-3sjoqm) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/b286a607df0e7d319f42e13b9ac067afe1e4e8a6/iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m#L773)):*
> ```objc
>         
>         if (!extension) {
>             NSString *newExtension = mimeType.fileExtensionForMimeType;
> ```

I think Mime type should override query params if it is a valid type iOS supports. There might be something unrelated in the last key in the query parameters so I think it would be better to use the Mime type as it has a higher chance of being correct

---


*Comments from [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/366#-:-LAtrpP20pnW1SUk11Vg:bqrsalg)*

jkasten2:
Reviewed 3 of 7 files at r1, 6 of 6 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/366)*

PR #372, opened 2018-05-18
Fix HTTP Race Condition

Nightsd01:
• Fixes a relatively rare issue that would be encountered if the SDK attempted to make concurrent HTTP requests during an invalid situation, such as after the user has revoked GDPR consent, or if the SDK hasn't received the app ID yet. 
• The SDK was previously using a semaphore to wait for a batch of HTTP requests to finish.
• However, if the SDK immediately fails the request (synchronously), the request would fail before the SDK started waiting for the request to finish, resulting in a permanently blocked queue.
• Switched to using a dispatch_group which would handle this situation better. Also added a missing GDPR consent check to the onFocus methods

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/372)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalClient.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalClient.m
@@ -117,14 +117,6 @@
         return;
     }
     
-    NSData *requestData = request.request.HTTPBody;
-    
-    if (requestData) {
-        NSString *json = [[NSString alloc] initWithData:requestData encoding:NSUTF8StringEncoding];
-        
-        NSLog(@"Executing request %@ with json: %@", NSStringFromClass([request class]), json);
-    }
-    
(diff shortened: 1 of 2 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignalClient.m
```diff
@@ -116,6 +117,14 @@ - (void)executeRequest:(OneSignalRequest *)request onSuccess:(OSResultSuccessBlo
         return;
     }
     
+    NSData *requestData = request.request.HTTPBody;
+    
+    if (requestData) {
+        NSString *json = [[NSString alloc] initWithData:requestData encoding:NSUTF8StringEncoding];
+        
+        NSLog(@"Executing request %@ with json: %@", NSStringFromClass([request class]), json);
```
jkasten2: Our OneSignal logging method should be used instead of `NSLog`

PR #377, opened 2018-05-29
Bundle ID Improvement

Nightsd01:
• In order to allow communication between the App Extension and the host/primary application, OneSignal SDK uses an app group to perform communication
• For most developers, the name of the app group will simply be group.{your_bundle_id}.onesignal
• The SDK can automatically retrieve this bundle ID, but previously, for the app extension service, this was broken because it was only retrieving the bundle ID for the OneSignalNotificationServiceExtension
• This means the bundle ID's would not have matched and badge count would be inconsistent. This commit fixes the issue by using a method to retrieve the host app's bundle ID.

### Note 1
Users can override this app group name by adding a `OneSignal_app_groups_key` value to their `info.plist` for both the main app and extension service.

### Note 2
This PR also changes the example/demo app in a few ways to avoid confusion. We removed a method that would set the badge count to 11 whenever a notification is received, and we enabled automatic badge clearing.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/377)

### DIFF
--- a/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
+++ b/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
@@ -59,8 +59,6 @@
     
     id notificationReceiverBlock = ^(OSNotification *notification) {
         NSLog(@"Received Notification - %@", notification.payload.notificationID);
-        
-        [[UIApplication sharedApplication] setApplicationIconBadgeNumber:11];
     };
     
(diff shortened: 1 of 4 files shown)


jkasten2:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/377#-:-LEG5HrB2xn7UJ_7_4fV:bnfp4nl)*

PR #398, opened 2018-08-13
Filter Out Non-OneSignal Notifications in Extension Methods

Nightsd01:
• We currently have logic to avoid processing push notifications if they weren't sent through OneSignal, to prevent issues with other providers.
• However we don't currently do any filtering with the extension service methods.
• Adds a check to make sure that the extension doesn't process non-OS notifications.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/398)


Nightsd01:
After discussing this with Josh we decided that the `nonOneSignalPayload` looked a bit confusing and added some comments to explain the logic.

Nightsd01:
Closing this PR because circumstances have changed. Due to linker warnings, Xcode now warns if an app extension depends on a framework that contains any references to UIKit or other extension-unavailable API's.

Because of this, we have decided to split-out our extension-specific code into its own separate smaller framework. Due to this, we will be approaching this problem (filtering out non-onesignal notifications from being processed by the extension service) in a different way.

PR #412, opened 2018-09-10
Dynamic framework

Nightsd01:
• Adds a `OneSignal-Dynamic-Framework` target to the project. This target is an aggregate target (does not directly output a product), it uses a buildscript to merge several Frameworks for different architectures (ie. x86_64 for simulator, arm64 for device) and produces a single fat dynamic framework

• The motivation for this change is that some wrapper SDK's use our iOS SDK as a cocoapod dependency. But in Swift projects that use `use_frameworks!`, no cocoapod dependency can depend on a static framework. Since our Flutter SDK (for example) uses our iOS SDK (which is a static framework), this caused a lot of issues for Swift/Flutter/React-Native/etc. developers

• To fix this, we've added a dynamic framework cocoapod in addition to our traditional cocoapod called `OneSignalDynamic`

• The dynamic framework is located at `{repo}/iOS_SDK/OneSignalSDK/Framework/Dynamic/OneSignal.framework`

• Also, when the project was initially created, the original developer/s named the framework target `OneSignal-Dynamic` but it actually outputs a static framework, for this reason I have renamed the target to `OneSignal-Static-Framework`.

• Another point of confusion to be aware of is that historically it's actually the `OneSignal` target that produces our static framework that gets used by cocoapods, it uses a buildscript to do this as well.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/412)


Nightsd01:
Josh (@jkasten2) is on vacation for a while. Since this PR does not modify our existing release/functionality I can without review for the time being.

PR #427, opened 2018-09-28
Extension framework

Nightsd01:
• The iOS clang linker will emit a warning if an iOS app extension uses a framework that uses UIKit (or some other extension-unavailable API) 
• To fix this (and to improve separation of concerns), we will be splitting out Notification Service Extension functionality into its own separate framework for the next major release (3.0)

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/427)


Nightsd01:
___
*[iOS_SDK/OneSignalSDK/Source/OneSignalExtensionBadgeHandler.m, line 46 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/427#-LNT1gT6COksEtzqUA-O:-LO05ANT9Xmu4xPXUnV0:b-xfwbm6) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/4aeb6d8cda816712411d58324b085ba83685d79f/iOS_SDK/OneSignalSDK/Source/OneSignalExtensionBadgeHandler.m#L46)):*

Why change these from inferred types? If this is due to the pre processors missing can we moved them into a header that can be shared?
</blockquote></details>

Fixed. Discussed this with Josh and decided it's a good idea to keep using let/var auto deduced type macros. Moved the auto-deduce type macros to `OneSignalShared.h` so that it can be shared between the SDK and Extension framework.

PR #488, opened 2019-03-10
Add Display Type Overrider Delegate

Nightsd01:
• This PR adds the ability to customize the display type for individual push notifications.
• This is useful for example in messaging apps, where you do not need to show an alert/notification for messages received if you are already viewing a thread with the same person.

• This PR adds the following method to OneSignal:

`[OneSignal setNotificationDisplayTypeDelegate:delegate]`

• When a notification is received, the following method on the delegate will fire:

```objc
- (void)willPresentInFocusNotificationWithPayload:(OSNotificationPayload *)payload forDisplayType:(OSNotificationDisplayType *)type
```

• The display type is passed in as a pointer that the app can set (or not). If the developer wants to set the `NONE` display type for this specific notification, they would do so like this:

```objc
- (void)willPresentInFocusNotificationWithPayload:(OSNotificationPayload *)payload forDisplayType:(OSNotificationDisplayType *)type {
   *type = OSNotificationDisplayTypeNone;
}
```

• Setting the display type for this notification does not effect any future notifications that may be received, and the SDK will always default to whatever was set with `[OneSignal setInFocusDisplayOption:]`

• Adds a test to ensure that the behavior works correctly
• Adds this delegate to the demo project, it can be tested by sending `overrideDisplayType` in the `additionalData` of a push notification and setting it to `0` (none), `1` (alert), or `2` (notification)

### DIFF
--- a/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.h
+++ b/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.h
@@ -31,7 +31,7 @@
 #import <UIKit/UIKit.h>
 #import <OneSignal/OneSignal.h>
 
-@interface AppDelegate : UIResponder <UIApplicationDelegate, OSPermissionObserver, OSSubscriptionObserver, OSEmailSubscriptionObserver>
+@interface AppDelegate : UIResponder <UIApplicationDelegate, OSPermissionObserver, OSSubscriptionObserver, OSEmailSubscriptionObserver, OSNotificationDisplayTypeDelegate>
 
 @property (strong, nonatomic) UIWindow *window;
 
--- a/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
+++ b/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
@@ -77,6 +77,8 @@
     [OneSignal addSubscriptionObserver:self];
     [OneSignal addEmailSubscriptionObserver:self];
     
+    [OneSignal setInFocusNotificationDisplayTypeDelegate:self];
+    
     NSLog(@"UNUserNotificationCenter.delegate: %@", UNUserNotificationCenter.currentNotificationCenter.delegate);
     
     return YES;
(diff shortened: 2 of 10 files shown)


Nightsd01:
@jkasten2 we could do that, but it would complicate things. I’m using a pointer because I’m guessing that in a lot of cases they’ll just want to use the default display type in which case they don’t need to dereference the pointer or do anything with it. 

If we use a block however, that makes things a bit more complex since we’ll have to do something if they never call it, do we check after a timer? Etc. so I figured this was simpler and faster.

Nightsd01:
@jkasten2 now that I think about it a bit more, I think a callback might actually be best since it will make it easier to use with wrapper SDK’s that can’t do synchronous returns. That should make it easy to implement in the wrapper SDK’s. 

I will update the PR this weekend to use a callback with a 30 second timer - that sound good? The only complexity is that multiple notifications can arrive pretty quickly so I’ll keep that in mind.

jkasten2:
@Nightsd01 It might be worth trying to see what happens when the `withCompletionHandler` does not fire to get the time down for sure, then add a bit of buffer to it.

Also I wonder if iOS will wait until the app has called the `withCompletionHandler` callback before calling `userNotificationCenter:willPresentNotification:withCompletionHandler:` again. If it does do this it will save on the SDK needing to keep track of a callback list. 😄

Nightsd01:
@jkasten2 updated the diff! I’ll be adding a test soon but it should be good to take a look at the code now

Nightsd01:
___
*[iOS_SDK/OneSignalSDK/Source/OneSignal.h, line 222 at r2](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/488#-LaHR-uQ09v9DOsVrfVp:-LaHcn_6EFJfWWZ60qTk:b-30ttyg) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/c8d1a206a42eba06a2d835b6aaba921fca2b84c7/iOS_SDK/OneSignalSDK/Source/OneSignal.h#L222)):*

Thinking we can drop `withDefaultDisplayType` as I think it is something that wouldn't be used. It also wouldn't be different per notification so if it needs to be read it should be something they can read from calling something on the OneSignal class. 
</blockquote></details>

The reason I included the default display type as a parameter here is to make it easy for developers to just call the completion block with the default value if they don't want to override it in some situations.

They could also just call the completion block with `OneSignal.inFocusDisplayType` but I think this is less obvious. What do you think?

PR #491, opened 2019-04-03
Fix prompting for location permission

troy-lamerton:
Fixed use of the prompt boolean.

If the authorization status of location permission is 'when in use' then the always allow prompt was being shown by one signal - even when the prompt parameter is false.

This PR enforces the correct conditions to request location permission.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/491)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
@@ -171,8 +171,9 @@ static OneSignalLocation* singleInstance = nil;
     // Check for location in plist
     if (![clLocationManagerClass performSelector:@selector(locationServicesEnabled)])
         return;
-    
-    if ([clLocationManagerClass performSelector:@selector(authorizationStatus)] == 0 && !prompt)
+    int permissionStatus = [clLocationManagerClass performSelector:@selector(authorizationStatus)];
+    // return if permission not determined and should not prompt
+    if (permissionStatus == 0 && !prompt)
         return;
     
     locationManager = [[clLocationManagerClass alloc] init];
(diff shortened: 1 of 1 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
```diff
@@ -186,15 +187,17 @@ + (void) internalGetLocation:(bool)prompt {
         //Location Always requires: Location Background Mode + NSLocationAlwaysUsageDescription
         NSArray* backgroundModes = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"UIBackgroundModes"];
         NSString* alwaysDescription = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSLocationAlwaysUsageDescription"] ?: [[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSLocationAlwaysAndWhenInUseUsageDescription"];
-        if(backgroundModes && [backgroundModes containsObject:@"location"] && alwaysDescription) {
+        // use background location updates if always permission granted or prompt allowed
+        if(backgroundModes && [backgroundModes containsObject:@"location"] && alwaysDescription && (permissionStatus == 3 || prompt)) {
```
jkasten2: Just to confirm with the `3` value here you are checking for `kCLAuthorizationStatusAuthorized`?
https://developer.apple.com/documentation/corelocation/clauthorizationstatus?language=objc

I believe we should be checking for `kCLAuthorizationStatusAuthorizedWhenInUse` as well.

iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
```diff
@@ -186,15 +187,17 @@ + (void) internalGetLocation:(bool)prompt {
...
             [locationManager performSelector:@selector(requestAlwaysAuthorization)];
```
jkasten2: I think if we move `requestWhenInUseAuthorization` and `requestAlwaysAuthorization` into their own methods and do the required checks to prevent prompting when it isn't enable might be the better approach. We can do this refracting after we merge your PR but feel free to make make larger changes if you are feeling ambitious 😄

iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
```diff
@@ -186,15 +187,17 @@ + (void) internalGetLocation:(bool)prompt {
...
             if (deviceOSVersion >= 9.0) {
                 [locationManager setValue:@YES forKey:@"allowsBackgroundLocationUpdates"];
             }
         }
         
-        else if([[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSLocationWhenInUseUsageDescription"])
-            [locationManager performSelector:@selector(requestWhenInUseAuthorization)];
+        else if([[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSLocationWhenInUseUsageDescription"]) {
+            if (permissionStatus == 0) [locationManager performSelector:@selector(requestWhenInUseAuthorization)];
```
jkasten2: We may also want to check for `kCLAuthorizationStatusRestricted` here.

jkasten2:
@troy-lamerton Thanks for digging in and making this PR! I left some comments about if these are the correct `permissionStatus` values to be checked.

iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
troy-lamerton: That is covered by the check if `prompt` is true: `(permissionStatus == 3 || prompt)`.

If the status is `kCLAuthorizationStatusAuthorizedWhenInUse` then background location updates should only be attempted if `prompt`ing is allowed.

iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
troy-lamerton: Why is that?
> `kCLAuthorizationStatusRestricted`: This app is not authorized to use location services.
The user cannot change this app’s status, possibly due to active restrictions such as parental controls being in place.

The theme I understood from the code is that if the app doesn't have permission, just continue on anyway and the calls to start location tracking will simply do nothing.

iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
troy-lamerton: I just want to fix the issue.

iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
jkasten2: I see, thanks for confirming.

iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
jkasten2: You are right, we only need to check for `0` here. Was thinking the inverse at first.

jkasten2:
I confirmed this fixes the issue of prompting for background location if only the foreground position is accepted when the app is cold started the next time.

PR #507, opened 2019-07-24
initialization method was added with opportunity to set base url

sskurzhanskiy:
This is needed for bypass blocking to you server

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/507)


jkasten2:
@sskurzhanskiy Thanks for the PR however we do not plan to support changing the root endpoint for the SDK.

PR #509, opened 2019-08-04
iOS 13 Token Fix

Nightsd01:
- This PR fixes an issue where the SDK was relying on NSData's `description` field to convert the APNS token to a string
- This worked fine up until iOS 13 where Apple has changed NSData's `description` to be more like Swift, where it includes the byte length in addition to the data being encoded
- Our existing test suite would actually have caught this issue if ran in an iOS 13 environment. However I did add a test to make sure our SDK can handle variable length APNS tokens

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/509)


jkasten2:
@Nightsd01 Thanks for the PR! Tested it back to iOS 9 and still works there too.

PR #511, opened 2019-08-08
Cleaned up payload conversion implementation

jkasten2:
* Also added nonnul to parseWithApns
* Confirmed payload parsing works on both iOS 9.3.6 and iPadOS 13 Beta 6

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/511)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/NSString+OneSignal.h
+++ b/iOS_SDK/OneSignalSDK/Source/NSString+OneSignal.h
@@ -18,7 +18,7 @@
 - (NSString *)supportedFileExtension;
 
 // returns a lower case hex representation of the data
-+ (NSString *)hexStringFromData:(NSData *)data;
++ (nullable NSString *)hexStringFromData:(nonnull NSData *)data;
 
 @end
 #endif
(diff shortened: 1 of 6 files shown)


iOS_SDK/OneSignalSDK/Source/NSString+OneSignal.h
```diff
@@ -18,7 +18,7 @@
 - (NSString *)supportedFileExtension;
 
 // returns a lower case hex representation of the data
-+ (NSString *)hexStringFromData:(NSData *)data;
++ (nullable NSString *)hexStringFromData:(nonnull NSData *)data;
```
Jeasmine: can we use _Nullable _Nonnull only for convention purposes?

jkasten2: We do have a mixture right now, I think the lower case is cleaner and easier to read. These lowercase versions seem to be exactly the same function too.
```
The nullable and nonnull annotations are simplified forms of the _Nullable and _Nonnull annotations
```
https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/designating_nullability_in_objective-c_apis

PR #567, opened 2019-11-25
Improvements

Nightsd01:
# Some various improvements:
- Fixes an issue with Outcomes where we attempted to insert a value into a dictionary that was explicitly marked as nullable, this can result in crashes
- Adds a new common define (`OS_NULL_IF_NUL()`) to make this cleaner
- We had several classes where we apparently added methods to conform to the `NSCoding` protocol, but we never actually marked the classes as conforming to this protocol for some reason
- Fixes a bug where the `OSSessionResult` class' initializers didn't actually initialize
- Changes a few cases where initializers were marked to return `id`, it is preferable to return `instancetype` in Objective-C to avoid unnecessary casting and give the compiler better information to work with

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/567)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSIndirectNotification.h
+++ b/iOS_SDK/OneSignalSDK/Source/OSIndirectNotification.h
@@ -25,7 +25,7 @@
  THE SOFTWARE.
  */
 
-@interface OSIndirectNotification : NSObject
+@interface OSIndirectNotification : NSObject <NSCoding>
 
 @property (nonatomic, readonly) NSString *notificationId;
 @property (nonatomic, readonly) double timestamp; // seconds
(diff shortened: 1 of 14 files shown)


Nightsd01:
My git skills are very rusty :D I will clean up these commits in a bit. 

Also, looks like we should really look into two specific flaky tests, they have been causing test failures for a long time

Nightsd01:
Closing for now, will create a new PR when I have a chance to clean up the commits on my fork, Sourcetree has really messed things up.

jkasten2:
@Nightsd01 Thanks. Looks like some good clean up! I'll wait for your next PR to take a deeper look

PR #569, opened 2019-11-29
Added initDone to prevent double initializations

jkasten2:
* Without this the session and outcomeEvents controllers could be created more than once.
* Removed self check in public init method, not needed since it is static
  - This moved a large nested block a tab back
* Split the notificationBlocks:: selector OneSignalHelper into two selectors
* Corrected outcome cache clearing tests to restart the app
* Other small misc clean up

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/569)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -174,6 +174,9 @@ static BOOL delayedInitializationForPrivacyConsent = false;
 // method can be called the moment the user provides privacy consent.
 DelayedInitializationParameters *delayedInitParameters;
 
+// Ensure we only initlize the SDK once even if the public method is called more.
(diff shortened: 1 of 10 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -174,6 +174,9 @@ @implementation OneSignal
 // method can be called the moment the user provides privacy consent.
 DelayedInitializationParameters *delayedInitParameters;
 
+// Ensure we only initlize the SDK once even if the public method is called more.
+static BOOL intiDone;
```
mikechoch: Fix spelling to `initDone`

jkasten2: Fixed

PR #591, opened 2020-01-09
WIP - Fix subscribing when Background Modes are off and Direct To History is enabled

jkasten2:
## Issue
If Background Modes are off under Xcode > Capabilities AND iOS "Enable iOS 12 Direct To History" is enabled on the OneSignal dashboard settings the player will be created with the following on the dashboard;"Provisioning Profile Missing Push Permission"

The SDK reports `notification_types` as `80` and no identifier.

Also calling `OneSignal.promptForPushNotificationsWithUserResponse` says that it is disabled when it isn't and won't show a prompt.

## TODO
1. `registerForAPNsToken` needs to check for `notifPermissionState.provisional`.
  * Complete in commit 37a4d6f
1. We need to delay calling `registerForAPNsToken` until the download params completes. 
1. Write unit test reproducing issue

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/591)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -878,9 +878,11 @@ void onesignal_Log(ONE_S_LOG_LEVEL logLevel, NSString* message) {
     backgroundModesEnabled = (backgroundModes && [backgroundModes containsObject:@"remote-notification"]);
     
     // Only try to register for a pushToken if:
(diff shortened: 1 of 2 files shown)


jkasten2:
**This PR is approved however incomplete so do not merge it yet!**

emawby:
Prompting for full push now works as intended but the dashboard still shows "Provisioning Profile Missing Push Permission". What should we be displaying if background push is off?

Issue #590, opened 2020-01-09
Crash on app opening


tapkain:
**Description:**

User user app for some time, after it it freezes and crashed, then it starts to crash on the app start.

```
Fatal Exception: NSInvalidArgumentException
*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil
0  CoreFoundation                 0x1c3b2698c __exceptionPreprocess
1  libobjc.A.dylib                0x1c2cff9f8 objc_exception_throw
2  CoreFoundation                 0x1c3aa03f8 _CFArgv
3  CoreFoundation                 0x1c3a110cc -[__NSArrayM insertObject:atIndex:]
4  Runner                         0x102e20188 -[OneSignalSessionManager getIndirectNotificationIds]
5  Runner                         0x102e1f5d8 -[OneSignalSessionManager restartSessionIfNeeded]
6  Runner                         0x102de83fc +[OneSignal registerUserInternal]
7  libdispatch.dylib              0x1c3564a38 _dispatch_call_block_and_release
8  libdispatch.dylib              0x1c35657d4 _dispatch_client_callout
9  libdispatch.dylib              0x1c3540dec _dispatch_lane_serial_drain$VARIANT$armv81
10 libdispatch.dylib              0x1c354192c _dispatch_lane_invoke$VARIANT$armv81
11 libdispatch.dylib              0x1c3549e08 _dispatch_workloop_worker_thread
12 libsystem_pthread.dylib        0x1c3746114 _pthread_wqthread
13 libsystem_pthread.dylib        0x1c3748cd4 start_wqthread
```

**Environment**
1. SDK version OneSignal (2.12.2)
2. cocoapods
3. I'm using also flutter but I'm not sure it's related to the flutter plugin, since the crash occurs in iOS version of it.

PR #592, opened 2020-01-09
Added null check for notificationId to fix issue #590

mikechoch:
* Some notificationIds are coming through as null
* Most likely non-OneSignal notifications slipping through and causing a crash when adding the null notificationId to the received notificationId array
* Added a null check and unit test to make sure no null or blank string notificationIds are tracked for confirmed deliveries and outcomes

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/592)

### DIFF
--- a/iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
+++ b/iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
@@ -263,6 +263,10 @@ static XCTestCase* _currentXCTestCase;
 + (void)receiveNotification:(NSString *)notificationId wasOpened:(BOOL)opened {
     // Create notification content
     UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
+    
+    if (!notificationId)
+        notificationId = @"";
+    
     content.userInfo = [self createNotificationUserInfo:notificationId];
     
     // Create notification request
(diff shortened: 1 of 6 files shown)


iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
```diff
@@ -263,6 +263,10 @@ + (void)answerNotificationPrompt:(BOOL)accept {
 + (void)receiveNotification:(NSString *)notificationId wasOpened:(BOOL)opened {
     // Create notification content
     UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
+    
+    if (!notificationId)
+        notificationId = @"";
```
jkasten2: Why set `notificationId` to a blank string instead of keeping it `nil`?

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
```diff
@@ -271,10 +275,10 @@ + (void)receiveNotification:(NSString *)notificationId wasOpened:(BOOL)opened {
     // Entry point for the NSE
     [OneSignalNotificationServiceExtensionHandler didReceiveNotificationExtensionRequest:request withMutableNotificationContent:content];
     
-    [self handleNotificationReceived:notificationId messageDict:content.userInfo wasOpened:opened];
+    [self handleNotificationReceived:content.userInfo wasOpened:opened];
 }
 
-+ (void)handleNotificationReceived:(NSString*)notificationId messageDict:(NSDictionary*)messageDict wasOpened:(BOOL)opened {
++ (void)handleNotificationReceived:(NSDictionary*)messageDict wasOpened:(BOOL)opened {
```
jkasten2: 👍 Nice param cleanup!

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
jkasten2: ah this is the test. Makes sense why it is needed now too

Issue #545, opened 2019-09-29
Device isn't registered with OneSignal Notifications


mahmoud-amer-m:
**Description:**

I have a single device that always has userId equal nil from `OneSignal.getPermissionSubscriptionState()?.subscriptionStatus.userId`
After setting the log environment to verbose, I found that this request: https://onesignal.com/api/v1/players always got time out

**Environment**

iOS Version: 12.4
OneSignal Version: 2.11.0 (Cocoapods

**Steps to Reproduce Issue:**

1. Install iOS SDK version 2.8.2 with Cocoapods into a project
2. Initialize the SDK in didFinishLaunchingWithOptions:
3. in onOSSubscriptionChanged, the new state always has nil userId

**Anything else:**
Logs:
2019-09-29 22:20:18.450612+0200 SignifyClients[313:10216] VERBOSE: ONESIGNAL setOneSignalDelegate CALLED: (null)
2019-09-29 22:20:18.450694+0200 SignifyClients[313:10216] VERBOSE: ONESIGNAL setOneSignalDelegate CALLED: <SignifyClients.AppDelegate: 0x282fa30a0>
2019-09-29 22:20:18.451155+0200 SignifyClients[313:10216] VERBOSE: updateNotificationTypes called: 15
2019-09-29 22:20:18.451234+0200 SignifyClients[313:10216] VERBOSE: startedRegister: 1
2019-09-29 22:20:18.451295+0200 SignifyClients[313:10216] VERBOSE: getNotificationTypes:mSubscriptionStatus: -1
2019-09-29 22:20:18.452692+0200 SignifyClients[313:10216] VERBOSE: updateNotificationTypes called: 15
2019-09-29 22:20:18.452742+0200 SignifyClients[313:10216] VERBOSE: startedRegister: 1
2019-09-29 22:20:18.452783+0200 SignifyClients[313:10216] VERBOSE: getNotificationTypes:mSubscriptionStatus: -1
2019-09-29 22:20:18.492596+0200 SignifyClients[313:10216] VERBOSE: oneSignalDidRegisterForRemoteNotifications:deviceToken:
2019-09-29 22:20:18.492729+0200 SignifyClients[313:10216] VERBOSE: updateDeviceToken:onSuccess:onFailure:
2019-09-29 22:20:18.492754+0200 SignifyClients[313:10216] VERBOSE: getNotificationTypes:mSubscriptionStatus: -1
[...]

PR #595, opened 2020-01-13
Update API URL to circumvent block in Egypt

### LINKED ISSUE #554: Egypt blocking OneSignal - Can't subscribe user to Onesignal push notification always returns null Help wanted!!
func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        // Override point for customization after application launch.
        
        FirebaseApp.configure()
          OneSignal.add(self as OSSubscriptionObserver)
        
        
        let notificationReceivedBlock: OSHandleNotificationReceivedBlock = { notification in

             print("Received Notification: \(notification!.payload.notificationID)")
          }

          let notificationOpenedBlock: OSHandleNotificationActionBlock = { result in
             // This block gets called when the user reacts to a notification received
             let payload: OSNotificationPayload = result!.notification.payload

             var fullMessage = payload.body
             print("Message = \(fullMessage)")

             if payload.additionalData != nil {
                if payload.title != nil {
                   let messageTitle = payload.title
                      print("Message Title = \(messageTitle!)")
                }

                let additionalData = payload.additionalData
                if additionalData?["actionSelected"] != nil {
                   fullMessage = fullMessage! + "\nPressed ButtonID: \(additionalData!["actionSelected"])"
                }
             }
          }

          let onesignalInitSettings = [kOSSettingsKeyAutoPrompt: false,
             kOSSettingsKeyInAppLaunchURL: true]

          OneSignal.initWithLaunchOptions(launchOptions,
             appId: "Removed app id",
             handleNotificationReceived: notificationReceivedBlock,
             handleNotificationAction: notificationOpenedBlock,
             settings: onesignalInitSettings)

          OneSignal.inFocusDisplayType = OSNotificationDisplayType.notification

        
        
        
     
      OneSignal.promptForPushNotifications(userResponse: { accepted in
        if(accepted){
            OneSignal.setSubscription(true)
        }else{
           
        }
      })
        
       
        
        return true
        
    
    }

This is my observer which fires as soon as the app runs even before i allow notifications giving me null push notification

    func onOSSubscriptionChanged(_ stateChanges: OSSubscriptionStateChanges!) {
        if !stateChanges.from.subscribed && stateChanges.to.subscribed {
           print("Subscribed for OneSignal push notifications!")
        }
        print("SubscriptionStateChange: \n\(stateChanges)")

        //The player id is inside stateChanges. But be careful, this value can be nil if the user has not granted you permission to send notifications.
        if let playerId = stateChanges.to.userId {
           print("Current playerId \(playerId)")
        }
     }

it also gives this error 

SubscriptionStateChange: 
Optional(<OSSubscriptionStateChanges:
from: <OSSubscriptionState: userId: (null), pushToken: (null), userSubscriptionSetting: 1, subscribed: 0>,
to:   <OSSubscriptionState: userId: (null), pushToken: 8924c708393c86a36ec29b4d0aa01d265c40a911258f1395997859f4c6687b21, userSubscriptionSetting: 1, subscribed: 0>

iSWORD:
It appears that subdomains of onesignal.com are not blocked by the Egyptian ISP TE Data (also known as WE)

Fixes #554 & #545

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/595)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalCommonDefines.h
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalCommonDefines.h
@@ -31,8 +31,8 @@
 #import <Foundation/Foundation.h>
 
 // networking
-#define API_VERSION @"api/v1/"
-#define SERVER_URL @"https://onesignal.com/"
+#define API_VERSION @"v1/"
+#define SERVER_URL @"https://api.onesignal.com/"
 
 // NSUserDefaults parameter names
 #define HAS_PROMPTED_FOR_NOTIFICATIONS @"OS_HAS_PROMPTED_FOR_NOTIFICATIONS"


iSWORD:
@mikechoch @jkasten2

Same as https://github.com/OneSignal/OneSignal-Android-SDK/pull/924#issuecomment-574590140

mikechoch:
Are you still experiencing OneSignal being blocked in Egypt? This should be fixed already, no need to modify the URL this would only be a temporary fix.

iSWORD:
It's not fixed yet unfortunately. I understand this is temporary, but it works fine.

Screenshot_20200115_234356_com android chrome

mikechoch:
Well sadly we can't merge a PR if this is simply a temporary fix, but you should not be experiencing any blocking so we will investigate this.

iSWORD:
It doesn't work on any of my devices. This fix might be temporary in a way that Egypt may block subdomains of OneSignal.com at some point, but it's not something that will break any functionality for other people, so merging it shouldn't be a problem.

mikechoch:
I understand, but we will not be changing the URL we use in our SDK. This should remain the same and we will look into getting Egypt up and running again if it is truly blocked.

I will be closing this PR now.

iSWORD:
That's a bummer. I don't see how changing the URL to a first party alias could cause any problems but you're the maintainer so what do I know. I have no other choice but to use SNS to meet our SLAs though because my clients will ask for an action plan with a clear timeline.

Thank you & good luck getting TE Data to unblock your domain :)

Kaabi:
> I understand, but we will not be changing the URL we use in our SDK. This should remain the same and we will look into getting Egypt up and running again if it is truly blocked.
> 
> I will be closing this PR now.

Moving to Firebase directly or Amazon SNS like @iSWORD mentioned is the best solution for a problem exists since August 2019 and only been noticeable in October 2019.

adding an argument to OneSignal init function to change base_url seem to be something you won't even wanna think about it.

muhammad-omar:
@mikechoch We can understand the size and impact proposed here along with the affected users.
However, it is actually affecting almost all of us here in Egypt and it is not 100% consistently blocked. Maybe 2 out of 10 trials passes. Also, ISPs claims that they did not block it, so it is no luck with asking them to allow it. 

We can check possible solutions like passing variables while installing the plugin with default one to be the current one. maybe pass one of predefined list to ensure more secure approach avoiding random URLs

jkasten2:
The change to use `api.onesignal.com` is now included in the release 2.13.0

https://github.com/OneSignal/OneSignal-iOS-SDK/releases/tag/2.13.0

Kaabi:
> The change to use `api.onesignal.com` is now included in the release 2.13.0
> 
> https://github.com/OneSignal/OneSignal-iOS-SDK/releases/tag/2.13.0

this is great news.
when OneSignal-Cordova-SDK will include the new SDK updates?

PR #617, opened 2020-02-19
Clean up views after dismissing webview by swiping down

### LINKED ISSUE #606: Swiping down to close launch URL notification content causes app to become unresponsive
**Description:**

When a push notification is sent with a Launch URL, after the user taps on the notification, our app opens, and a browser view is presented modally with a done button and the URL page displayed (We're not doing anything custom here, this is presented by OneSignal). If the user presses the done button, the browser view closes, and everything works as normal. The user is also able to drag/swipe downward on the modal view, which also closes the browser view, but if they close it using this method, the app becomes unresponsive. The reason for this is an empty view controller persists that sits on top of all the app views.

**Environment**

We are using version 2.12.5 of the OneSignal iOS SDK through cocoapods. We observed this issue on version 2.10.2 and updated to see if that resolved the issue, but we're still able to reproduce.

**Steps to Reproduce Issue:**

1. Install iOS SDK version 2.12.5 with Cocoapods into a project
2. Initialize the SDK in didFinishLaunchingWithOptions:

```
let onesignalInitSettings = [kOSSettingsKeyAutoPrompt: false]
        
OneSignal.initWithLaunchOptions(launchOptions,
                                appId: Config.oneSignalAppId,
                                handleNotificationAction: nil,
                                settings: onesignalInitSettings)

OneSignal.inFocusDisplayType = OSNotificationDisplayType.notification
```

3. App should have a UITabBarController as its rootViewController
4. Put app in background
5. Send a push notification to this device with a Launch URL set
6. Tap on notification to open URL in app
7. After browser modal opens, swipe/drag down to close browser view

Notice at this point the app becomes unresponsive.

**Anything else:**

We tested on various iPhones running iOS 13 flavors and can reproduce 100% of the time on our live app. While testing, after swiping to close the browser window, I paused the execution of the app in Xcode and opened the Debug View Hierarchy tool, and noticed that there is a UIViewController with an empty UIView on top of all the app views that must be blocking all user interaction.

kiyot:
Current implementation cleans up views only when 'done' button is tapped.
iOS 13 modal presentation style has other ways to dismiss view controllers
than tapping 'done' button, so we need to add cleaning up code to them.

This PR will fix #606

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/617)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
@@ -67,10 +67,7 @@ UIViewController *viewControllerForPresentation;
 
 - (void)dismiss:(id)sender {
     [self.navigationController dismissViewControllerAnimated:true completion:^{
-        // Clear web view
-        [_webView loadHTMLString:@"" baseURL:nil];
-        if (viewControllerForPresentation)
-            [viewControllerForPresentation.view removeFromSuperview];
+        [self clearWebView];
     }];
 }
 
@@ -108,6 +105,7 @@ UIViewController *viewControllerForPresentation;
     if (!navController) {
         navController = [[UINavigationController alloc] initWithRootViewController:self];
         navController.modalTransitionStyle = UIModalTransitionStyleCoverVertical;
+        navController.presentationController.delegate = self;
     }
     if (!viewControllerForPresentation) {
         viewControllerForPresentation = [[UIViewController alloc] init];
@@ -133,7 +131,15 @@ UIViewController *viewControllerForPresentation;
     @catch(NSException* exception) { }
 }
 
+- (void)clearWebView {
+    [_webView loadHTMLString:@"" baseURL:nil];
+    if (viewControllerForPresentation)
+        [viewControllerForPresentation.view removeFromSuperview];
+}
 
+- (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController {
+    [self clearWebView];
+}
 
 @end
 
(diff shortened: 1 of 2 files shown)


rgomezp:
Thank you so much for your contribution. We will review it for the next iOS native release

iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
```diff
@@ -133,7 +131,15 @@ - (void)showInApp {
...
 
+- (void)clearWebView {
+    [_webView loadHTMLString:@"" baseURL:nil];
+    if (viewControllerForPresentation)
+        [viewControllerForPresentation.view removeFromSuperview];
+}
 
+- (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController {
+    [self clearWebView];
+}
```
mikechoch, line 142: What is the thought and intention with implementing `UIAdaptivePresentationControllerDelegate`?

Are both methods required (`dismiss` and `presentationControllerDidDismiss`) or is one better to go with than the other? 

Or do they trigger in different scenarios and this solves an edge case?

kiyot: > What is the thought and intention with implementing UIAdaptivePresentationControllerDelegate?

The default `modalPresentationStyle` in iOS 13 is usually `pageSheet`, which can be dismissed by swiping down. Current implementation assumes that user can only dismiss the viewcontroller by tapping `leftBarButtonItem`, and necessary clean-up is done in `dismiss` method. When user dismisses the viewcontroller by swiping down, `dismiss` is not called and the problem in #606 occurs.

By implementing `UIAdaptivePresentationControllerDelegate`, you can receive events of viewcontroller dismissal and perform clean-ups.

> Are both methods required (dismiss and presentationControllerDidDismiss) or is one better to go with than the other?

When viewcontroller is dismissed by `dismissViewControllerAnimated` method, `presentationControllerDidDismiss` is not triggered. So both `dismiss` and `presentationControllerDidDismiss` need to call `clearWebView`.

PR #655, opened 2020-04-22
SDK version 3.0 - Renamed Session type to OSSession to avoid cross-package conflicts

### LINKED ISSUE #632: OneSignal and Turbolinks pods not compatible with each other?
**Description:**

We've been working with the SDK for a while and decided to update the OneSignal pod to solve other minor problems with an application I have been working on.

After updating the pod (from 2.10.0 to 2.12.6), I've encountered the following error:

```
<unknown>:0: error: reference to 'Session' is ambiguous

/Pods/OneSignal/iOS_SDK/OneSignalSDK/Framework/OneSignal.framework/Headers/OneSignal.h:214:29: note: candidate found by name lookup is 'Session'
typedef NS_ENUM(NSUInteger, Session) {

Turbolinks/Turbolinks.framework/Headers/Turbolinks-Swift.h:210:12: note: candidate found by name lookup is 'Session'
@interface Session : NSObject
```

What we understand of the problem is that both the OneSignal and the Turbolinks pods have a **Session** type declared, which Xcode takes as an error we cannot fix.

**Environment**

OneSignal SDK version 2.12.6 (using cocoapods)
Turbolinks version 3.0.0
Xcode 11.3.1
Using Swift

rgomezp:
Closes #632

Will have to go on next major release (3.0)

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/655)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSChannelTracker.h
+++ b/iOS_SDK/OneSignalSDK/Source/OSChannelTracker.h
@@ -33,7 +33,7 @@ THE SOFTWARE.
 
 @interface OSChannelTracker : NSObject
 
-@property (nonatomic, readwrite) Session influenceType;
+@property (nonatomic, readwrite) OSSession influenceType;
 @property (strong, nonatomic, readwrite, nullable) NSString *directId;
 @property (strong, nonatomic, readwrite, nullable) NSArray *indirectIds;
 @property (strong, nonatomic, readonly, nonnull) OSInfluenceDataRepository *dataRepository;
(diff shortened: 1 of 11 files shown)


mikechoch:
This is not necessary as @Jeasmine has a PR up and this will cause merge conflicts with it. Also she already changed the name to `OSInfluence` instead of `Session`

Jeasmine:
> This is not necessary as @Jeasmine has a PR up and this will cause merge conflicts with it. Also she already changed the name to `OSInfluence` instead of `Session`
> 
> I might be wrong about this though so not closing this until confirmed

Yes, with https://github.com/OneSignal/OneSignal-iOS-SDK/pull/656 this will be fixed, Session is renamed to OSInfluence

jkasten2:
This PR is still needed since we could not make a breaking change in PR #656 since it went out before the major release.

Jeasmine:
Agree still needed, but under the major release branch

PR #665, opened 2020-05-11
Check for prompts before entering logic

Jeasmine:
* We want to avoid evaluateMessageDisplayQueue being call if no prompts are available

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/665)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
@@ -539,7 +539,8 @@ static BOOL _isInAppMessagingPaused = false;
     if (action.clickUrl)
         [self handleMessageActionWithURL:action];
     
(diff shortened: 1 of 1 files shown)


iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
```diff
@@ -539,7 +539,8 @@ - (void)messageViewDidSelectAction:(OSInAppMessage *)message withAction:(OSInApp
     if (action.clickUrl)
         [self handleMessageActionWithURL:action];
     
-    [self handlePromptActions:action.promptActions];
+    if (action.promptActions.count > 0)
```
mikechoch, line 542: Should we nil check just in case?
`if (action.promptActions && action.promptActions.count > 0)`

PR #668, opened 2020-05-20
Fix Direct influence case

Jeasmine:
* Reset and cache indirect state to avoid direct being the last cached state

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/668)


iOS_SDK/OneSignalSDK/UnitTests/OutcomeV2Tests.m
```diff
@@ -197,8 +201,10 @@ - (void)testUnattributedSessionToIndirectSession {
     
     // 4. Rceive 3 notifications
```
mikechoch, line 202: `Rceive` should be `Receive`

Issue #179, opened 2017-02-08
SwiftPM support


TofPlay:
Hi,
[SwiftPM](https://goo.gl/E01ufX) is Swift's official package manager. In the near future it will replace Carthage and CocoaPods.

Here some articles about SwiftPM:
* [An Introduction to the Swift Package Manager](https://goo.gl/byiOc6)
* [Swift Dependency Management for iOS](https://goo.gl/7eaz1A)
* [Getting Started with Swift Package Manager (SwiftPM)](https://goo.gl/rCHeQU)
* [Swift.org: Package Manager](https://goo.gl/E01ufX)

When your component is compatible with SwiftPM add on your `README.md` [![SwiftPM Compatible](https://img.shields.io/badge/SwiftPM-Compatible-brightgreen.svg)](https://goo.gl/E01ufX)  badge on top and add a section for the installation with SwiftPM.

jkasten2:
@TofPlay Thanks for the suggestion. We will look into it in the future.

jkasten2:
@TofPlay I took a look but Apple doesn't support iOS with their SwiftPM. This is noted on their gihub page.
https://github.com/apple/swift-package-manager#swift-package-manager-project

Feel free to reopen this issue if Apple added support for iOS.

michael-mckenna:
@jkasten2 Now that it's the official dependency manager, will this be implemented soon?

jkasten2:
@michael-mckenna Thanks for the update, it looks like Xcode 11 adds support for the Swift package Manager for iOS.
https://github.com/apple/swift-package-manager#swift-package-manager-project

jeffc-dev:
Looking forward to seeing SwiftPM support added.

We're working to end the use of Cocoapods on our end and migrate all of both our own and 3rd party code to SPM packages whenever possible. The Xcode 11 integration is very slick.

Is this something OneSignal plans to add soon, @jkasten2?

makoni:
+1

meandkareem:
+1

MagzhanKZ:
+1

RyuX51:
+1

rgomezp:
Hi everyone,
We're working hard on many new features, products, etc... so thanks so much for your patience!

I've begun attempts to do this but I'm struggling to find good resources on how to add SPM support to an existing library, so my efforts are hitting a wall. Most content I've found is around creating a brand new package from scratch (which is quite straight forward).

[...]

PR #673, opened 2020-05-26
Feature/swiftPM

emawby:
Using XCode’s built in Swift PM support

1. Open XCode and navigate to your App’s project editor.
2. Select the Swift Packages tab and click the “+” to add a new package
3. Paste in the OneSignal iOS SDK Github URL. **To test pre-merge use https://github.com/emawby/OneSignal-iOS-SDK.git**
4. Choose to version the package off of the feature/SwiftPM branch
5. Select the static or dynamic library target
6. Navigate to the OneSignalNotificationExtension Target -> Build Phases Tab. Under the “Link Binary with Libaries section clickthe “+” and add the OneSignal library.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/673)

Closes https://github.com/OneSignal/OneSignal-iOS-SDK/issues/179


mikechoch:
After our talk today in sprint planning I think this is best left unmerged for now until we figure out the correct timeline for `Obj-C/Swift` usage and `SwiftPM` support

scottyelvington:
I'm just adding a note here that I've added to the SPM discussion. Companies are already starting the migration to SPM due to a few reasons:
1) it's been in development longer than some of the newer swift technologies
2) it's been used in production for longer than some of the newer swift technologies
3) it already supports fantastic Xcode integration
4) this is ultimately the future of development in Xcode

I don't think it can be ignored for very long without having unintended consequences

iOS_SDK/OneSignalSDK/Source/OneSignalNotificationData.h
```diff
@@ -1,38 +0,0 @@
...
-
-@interface OneSignalNotificationData : NSObject
-
-+ (void)saveLastNotificationFromBackground:(NSString * _Nullable)notificationId;
-+ (void)saveLastNotification:(NSString * _Nullable)notificationId;
-+ (void)saveLastNotificationWithBackground:(NSString *  _Nullable)notificationId wasOnBackground:(BOOL)wasOnBackground;
-+ (OSLastNotification * _Nonnull)getLastNotification;
-+ (NSString * _Nullable)getLastNotificationId;
-
-@end
```
mikechoch, line 38: Why are we removing these unnecessary `.h` and `.m` files in this PR?

mikechoch:
Went through setup only issue I had was that the `libOneSignal.a` prevented build because of duplicate error. We should use these instructions for our docs once we have this fully supported and we should add that if you were previous using our `.a` file it should be removed.

**Examples:**
Before removing the `libOneSignal.a`
Screen Shot 2020-07-08 at 12 07 17 PM
After removing the `libOneSignal.a`
Screen Shot 2020-07-08 at 12 07 28 PM

iOS_SDK/OneSignalSDK/Source/OneSignalNotificationData.h
emawby: They prevent the project from compiling when using SwiftPM so we need to remove them for this to work.

emawby:
> Went through setup only issue I had was that the `libOneSignal.a` prevented build because of `Multiple commands` error. We should use these instructions for our docs once we have this fully supported and we should add that if you were previous using our `.a` file it should be removed. Also it might be very helpful to add screenshots for each of your instruction steps.
> 
> **Examples:**
> Before removing the `libOneSignal.a`
> Screen Shot 2020-07-08 at 12 07 17 PM
> After removing the `libOneSignal.a`
> Screen Shot 2020-07-08 at 12 07 28 PM

Yep that makes sense. Most folks will probably need to remove the OneSignal Pod or OneSignal from their cart file. We can either have a separate SwiftPM setup page like we do for Carthage or move Cocoapods to its own page and make SwiftPM the setup that is in the base document.

rgomezp:
Snaps

dannyhertz:
Hey! Any ETA on this? Looking forward to integrating once SPM support is ready.

emawby:
This is using Swift tools version 5.2 and distributes a library not a framework (SwiftPM limitation).

In the future we will also distribute this via an XCFramework using Swift tools version 5.3

Package.swift
```diff
@@ -0,0 +1,27 @@
+// swift-tools-version:5.2
```
jkasten2, line 1: Would this support Xcode 11.0? If not then anything blocking us from lowering this to support it?
Xcode 10.x is not / almost not support any more by Apple so it is ok to omit that.

Package.swift
```diff
@@ -0,0 +1,27 @@
...
+            type: .dynamic,
+            targets: ["OneSignal"]),
+    ],
+    targets: [
+        .target(
+            name: "OneSignal",
+            dependencies: [],
+            path: "iOS_SDK/OneSignalSDK/",
+            sources: ["Source"],
+            publicHeadersPath:"Source"),
```
jkasten2, line 24: Using "Source" for `publicHeadersPath` results in all internal classes being accessible. Can we scope this down so only `OneSignal.h` is accessible?

Some ideas on how this could be done is pointing to one of the following paths if a specific file isn't possible:
* `iOS_SDK/OneSignalSDK/Framework/OneSignal.framework/Headers`
* `iOS_SDK/OneSignalSDK/Framework/OneSignal.framework/Versions/A/Headers`

Package.swift
```diff
@@ -0,0 +1,27 @@
...
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "OneSignal",
+    products: [
+        .library(
+            name: "OneSignal",
+            type: .static,
```
jkasten2, line 11: I noticed this in the swift docs:
> /// Create a library product that can be used by clients that depend on this package.
///
/// A library's product can either be statically or dynamically linked. It
/// is recommended to not declare the type of library explicitly to let the
/// Swift Package Manager choose between static or dynamic linking depending
/// on the consumer of the package.

Should we remove the `OneSignalDynamic` section and this ` type: .static,` line so Xcode can automatically pick?

jkasten2:
> This is using Swift tools version 5.2 and distributes a library not a framework (SwiftPM limitation).
> 
> In the future we will also distribute this via an XCFramework using Swift tools version 5.3

We might be able to use a pre-processor to support an XCFramework without breaking support for Xcode 11.0. Something we can do in a later PR however, so we don't have to delay this one.
```swift
#if swift(>=5.3)
// Add XCFramework here
#endif
```
Idea from: https://theswiftdev.com/the-swift-package-manifest-file/

Package.swift
emawby: fixed this!

Package.swift
emawby: fixed!

Package.swift
emawby: I made it 5.1 since that is what comes with 11.0

emawby:
Closing this PR since it was off the incorrect branches

Package.swift
jkasten2: Fixed in the new PR
https://github.com/OneSignal/OneSignal-iOS-SDK/pull/712/files#diff-37ca2dd15ca0f6b1b49e78db084ef5b9R6-R21

Package.swift
jkasten2: Fixed in the new PR
https://github.com/OneSignal/OneSignal-iOS-SDK/pull/712/files#diff-37ca2dd15ca0f6b1b49e78db084ef5b9R1

Package.swift
jkasten2: A new folder was made: `SwiftPM/Public/Headers`
https://github.com/OneSignal/OneSignal-iOS-SDK/pull/712/files#diff-37ca2dd15ca0f6b1b49e78db084ef5b9R19

jkasten2:
Replacement PR is https://github.com/OneSignal/OneSignal-iOS-SDK/pull/712

PR #674, opened 2020-05-26
Version instead of Build

emawby:
setting the game_version to use the App version instead of the App build number

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/674)


iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1606,8 +1604,9 @@ + (void)registerUserInternal {
     if (deviceModel)
         dataDic[@"device_model"] = deviceModel;
     
-    if (build)
-        dataDic[@"game_version"] = build;
+    NSString *version = infoDictionary[@"CFBundleShortVersionString"];
```
mikechoch, line 1607: Is there a `CFStringConstant` for `@"CFBundleShortVersionString"`?

mikechoch: Answered my own question, it seems like a no. This can be resolved

PR #681, opened 2020-06-02
Fix NSInvalidUnarchiveOperationException due to decodable class renamed

Jeasmine:
* Add migration class for future cases

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/681)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
@@ -26,21 +26,44 @@ THE SOFTWARE.
 */
 
 #import <Foundation/Foundation.h>
-#import "OneSignal.h"
 #import "OSMigrationController.h"
+#import "OSInfluenceDataRepository.h"
 #import "OSIndirectInfluence.h"
+#import "OneSignal.h"
+#import "OneSignalUserDefaults.h"
+#import "OneSignalCommonDefines.h"
+#import "OneSignalHelper.h"
+
+@interface OneSignal ()
++ (OSInfluenceDataRepository *)influenceDataRepository;
+@end
 
 @implementation OSMigrationController
 
 - (void)migrate {
     [self migrateToVersion_02_14_00_AndGreater];
+    [self saveCurrentSDKVersion];
 }
 
 /**
  * Support renaming of decodable classes for cached data
  */
 - (void)migrateToVersion_02_14_00_AndGreater {
-    [NSKeyedUnarchiver setClass:[OSIndirectInfluence class] forClassName:@"OSIndirectNotification"];
+    let influenceVersion = 21400;
+    let sdkVersion = [OneSignalUserDefaults.initShared getSavedIntegerForKey:OSUD_CACHED_SDK_VERSION defaultValue:0];
+    if (sdkVersion < influenceVersion) {
+        [OneSignal onesignal_Log:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Migrating from version : %d", sdkVersion]];
+
+        [NSKeyedUnarchiver setClass:[OSIndirectInfluence class] forClassName:@"OSIndirectNotification"];
+        NSArray<OSIndirectInfluence *> * indirectInfluenceData = [[OneSignal influenceDataRepository] lastNotificationsReceivedData];
+        if (indirectInfluenceData)
+            [[OneSignal influenceDataRepository] saveNotifications:indirectInfluenceData];
+    }
+}
+
+- (void)saveCurrentSDKVersion {
+    let currentVersion = [[OneSignal sdk_version_raw] intValue];
+    [OneSignalUserDefaults.initShared saveIntegerForKey:OSUD_CACHED_SDK_VERSION withValue:currentVersion];
 }
 
 @end
--- a/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
@@ -1,0 +1,46 @@
+/**
+Modified MIT License
+
+Copyright 2020 OneSignal
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+1. The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+2. All copies of substantial portions of the Software may only be used in connection
+with services provided by OneSignal.
+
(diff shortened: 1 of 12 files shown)


iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -0,0 +1,46 @@
...
+@implementation OSMigrationController
+
+- (void)migrate {
+    [self migrateFromVersion_2_14_0_to_greater];
+}
+
+/**
+ * Support decodable of renamed classes
+ */
+- (void)migrateFromVersion_2_14_0_to_greater {
```
mikechoch, line 42: Let's rename to this?
`migrateToVersion_02_14_00_AndGreater`

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -0,0 +1,46 @@
...
+#import "OSIndirectInfluence.h"
+
+@implementation OSMigrationController
+
+- (void)migrate {
+    [self migrateFromVersion_2_14_0_to_greater];
+}
+
+/**
+ * Support decodable of renamed classes
```
mikechoch, line 40: Just wording the comment a bit
`Support renaming of decodable classes for cached data`

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -0,0 +1,46 @@
...
+
+#import <Foundation/Foundation.h>
+#import "OneSignal.h"
+#import "OSMigrationController.h"
+#import "OSIndirectInfluence.h"
+
+@implementation OSMigrationController
+
+- (void)migrate {
+    [self migrateFromVersion_2_14_0_to_greater];
```
mikechoch, line 36: What if someone downgrades, shouldn't we revert the name back as well? Or is this too much of a one off case/unexpected behavior?

`[self migrateToVersion_02_13_01_AndLesser];`
`[NSKeyedUnarchiver setClass:[OSIndirectNotification class] forClassName:@"OSIndirectInfluence"];`

Jeasmine: For this, we will need to save the last version know, in that way we can choose for which migration to use.
@jkasten2 what do you think? should we add version logic and contemplate this case?

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -0,0 +1,46 @@
...
+
+- (void)migrate {
+    [self migrateToVersion_02_14_00_AndGreater];
+}
+
+/**
+ * Support renaming of decodable classes for cached data
+ */
+- (void)migrateToVersion_02_14_00_AndGreater {
+    [NSKeyedUnarchiver setClass:[OSIndirectInfluence class] forClassName:@"OSIndirectNotification"];
```
jkasten2, line 43: Does this actually migrate the data or just use the old name for reading and writing?

If this is just to keep the old name can this just live in the `OSIndirectInfluence` class itself?

jkasten2: I see, since the key being saved and loaded isn't changing the data itself isn't being migrated. I think we should either migrate the data here or move this name mapping to `OneSignalUserDefaults.initStandard` since this is where the mapping is being applied.

Do we also need to call `NSKeyedArchiver` so the old name is also used when saving?

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -0,0 +1,69 @@
...
+
+@interface OneSignal ()
++ (OSInfluenceDataRepository *)influenceDataRepository;
+@end
+
+@implementation OSMigrationController
+
+- (void)migrate {
+    [self migrateToVersion_02_14_00_AndGreater];
+    [self saveLastKnowVersion];
```
mikechoch, line 45: Rename to `saveCurrentSDKVersion`

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * Support renaming of decodable classes for cached data
+ */
+- (void)migrateToVersion_02_14_00_AndGreater {
+    let influenceVersion = 21400;
+    let sdkVersion = [OneSignalUserDefaults.initShared getSavedIntegerForKey:OSUD_CACHED_SDK_VERSION defaultValue:0];
```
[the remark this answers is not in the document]
Jeasmine: yeah but we don't have a version saved until now, so for this first migration we are migrating all versions to 21401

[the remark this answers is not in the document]
mikechoch: Makes sense, so other migration methods down the road might return early in the case. But not this one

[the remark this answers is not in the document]
jkasten2: A default of `0` makes sense to me

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
jkasten2: If someone downgrades to 2.13.1 then there won't be any code there to downgrade. So writing a `migrateToVersion_02_13_01_AndLesser` here for the 2.14.1 release would be impossible to hit its run condision.

There are three options to come to mind to handle downgrading an SDK version:
1. We could make a 2.13.2 release to support downgrading, but we would always need to make sure we make all release with a migration a minor version bump & someone would know this is the only safe way to migrate.
2. Make a migration rollback schema. It will contain things like codeable class renames, field renames, add / removing fields, etc. We will need to support all possible change operations in the first release so when the code is rolled back the generic code can read all the operations it needs to rollback.
3. Clear OneSignal data. This is what we do for the Android SDK today with SQL database, we just clear all records for the database to be safe when the SDK is rolled back.

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -0,0 +1,69 @@
...
+    if (sdkVersion < influenceVersion) {
+        [OneSignal onesignal_Log:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Migrating from version : %d", sdkVersion]];
+
+        [NSKeyedUnarchiver setClass:[OSIndirectInfluence class] forClassName:@"OSIndirectNotification"];
+        NSArray<OSIndirectInfluence *> * indirectInfluenceData = [[OneSignal influenceDataRepository] lastNotificationsReceivedData];
```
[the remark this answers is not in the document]
Jeasmine: Yes this will migrate too, I added an unittest for this case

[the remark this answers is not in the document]
jkasten2: 👍 I see the newly added `testIndirectInfluenceToIndirectInfluenceMigration` which covers this!

PR #688, opened 2020-06-09
Fix DIRECT IAM influence being cache

Jeasmine:
* IAM Direct influence shoudn't be cache because DIRECT Influence will be only while the IAM is being display
   * Instead of caching DIRECT cache INDIRECT

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/688)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageTracker.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageTracker.m
@@ -86,7 +86,10 @@ THE SOFTWARE.
 }
 
 - (void)cacheState {
-    [self.dataRepository cacheIAMInfluenceType:self.influenceType];
+    // We only need to cache INDIRECT and UNATTRIBUTED influence types
+    // DIRECT is downgrade to INDIRECT to avoid inconsistency state
+    // where the app might be close before dismissing current displayed IAM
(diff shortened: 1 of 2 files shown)


iOS_SDK/OneSignalSDK/Source/OSInAppMessageTracker.m
```diff
@@ -86,7 +86,7 @@ - (void)initInfluencedTypeFromCache {
 }
 
 - (void)cacheState {
-    [self.dataRepository cacheIAMInfluenceType:self.influenceType];
+    [self.dataRepository cacheIAMInfluenceType:self.influenceType == DIRECT ? INDIRECT : self.influenceType];
 }
```
mikechoch, line 90: Does this do the same thing as Android `cacheState` method? Doesn't look like you have the `nil` check

Jeasmine: because the inits between iOS and Android are slightly differents so iOS won't reach the point of having influence type nil, but I will add it to stay consistent between platforms

Jeasmine: I checked again, and the reason behind this is because on iOS enums cant be null, so it will always default. That why there is no nil check

PR #695, opened 2020-06-17
Always call registerForRemoteNotifications from the main thread

emawby:
This is a fix for an issue showing up in our react-native SDK where the `registerForRemoteNotifications` is happening on a background thread. This causes a UI freeze.
https://github.com/OneSignal/react-native-onesignal/issues/749

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/695)


emawby:
Luckily we already are doing this for both location prompting and IAM prompt actions

jkasten2:
@emawby Feel free to merge into master for this one! Since is it a fix it can safely go into the next release whether it is major, minor , or patch.

PR #687, opened 2020-06-09
Implement notification foreground handler with display option

emawby:
- The handler can be set via the public setter method and takes an `OSNotificationGenerationJob`.

- The `OSNotificationGenerationJob` can be used to change the display type of the notification before it is shown e.g. making a push notification silent if the app is already open.

- At the end of the `willShowInForegroundHandler` block the `OSNotificationGenerationJob's` `complete` method should be called to indicate that the app should proceed with handling the notification. If the `complete` method is not called within 25 seconds of receiving the `willShowInForegroundHandler` callback then it will be called automatically.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/687)

PR #703, opened 2020-06-29
Feature/implement notification opened handler

emawby:
1. This PR fixes the implementation fo the Notification Opened Handler when the display type of a notification is changed in the `notificationWillShowInForegroundHandler`
2. It also renames our swizzled version of `receiveRemoteNotification` in our UIApplicationDelegate category from `oneSignalRemoteSilentNotification` to `oneSignalReceiveRemoteNotification` since the notification is not necessarily silent. 
3. In `UNUserNotificationCenter+OneSignal.m` when calling `notificationReceived` we always set `wasOpened` to no. This is called when the app is in the foreground before the notification has been presented to the user. There was previously a case where we set this opened flag to YES for a silent notification that had a non nil content body. This causes issues when overriding the display type of a notification to silent.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/703)

### DIFF
--- a/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
+++ b/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
@@ -31,7 +31,19 @@
 #import "AppDelegate.h"
 #import "ViewController.h"
 
+@interface OneSignalNotificationCenterDelegate: NSObject<UNUserNotificationCenterDelegate>
+@end
+@implementation OneSignalNotificationCenterDelegate
+
+- (void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNotification:(UNNotification *)notification withCompletionHandler:(void (^)(UNNotificationPresentationOptions))completionHandler {
+    NSLog(@"Appdelegatewillpresentdelegate");
+}
+
+@end
+
 @implementation AppDelegate
+
+OneSignalNotificationCenterDelegate *_notificationDelegate;
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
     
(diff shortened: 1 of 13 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
```diff
@@ -504,7 +504,7 @@ + (void)handleNotificationAction:(OSNotificationActionType)actionType actionID:(
     
     [OneSignalTrackFirebaseAnalytics trackOpenEvent:result];
     
-    if (!notificationOpenedHandler)
+    if (!notificationOpenedHandler || displayType != OSNotificationDisplayTypeNotification)
```
mikechoch, line 507: Instead of checking `displayType != OSNotificationDisplayTypeNotification`

Lets do `displayType == OSNotificationDisplayTypeSilent` so its a little more readable in that fact that `Silent` should not fire opened handler

iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
```diff
@@ -107,10 +106,10 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     return YES;
 }
 
-#define ONESIGNAL_APP_ID_KEY_FOR_TESTING @"77e32082-ea27-42e3-a898-c72e141824ef"
+#define ONESIGNAL_APP_ID_KEY_FOR_TESTING @"ONESIGNAL_APP_ID_KEY_FOR_TESTING"
```
mikechoch, line 109: We should put the default `appId` (`0ba9731b-33bd-43f4-8b59-61172e27447d`) here

iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
```diff
@@ -107,10 +106,10 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
 
 + (NSString*)getOneSignalAppId {
-    NSString* newAppId = @"77e32082-ea27-42e3-a898-c72e141824ef";
+    NSString* newAppId = @"0ba9731b-33bd-43f4-8b59-61172e27447d";
```
mikechoch, line 112: This can stay as the default (`0ba9731b-33bd-43f4-8b59-61172e27447d`) as well, but the point is that we can easily chan get `appId` and rerun without having to worry about the app using the cached one but showing in the UI the new one.

emawby:
> LGTM, can you squash those commits, to avoid those merge and revert merge on the history
Squashed!

jkasten2:
@Jeasmine Ah good catch
@emawby I think a rebase with the following command will fix this:
```bash
git rebase --onto feature/implement_notification_foreground_handler_with_display_option feature/implement_notification_opened_handler~1 feature/implement_notification_opened_handler
```

`feature/implement_notification_opened_handler~1` as it seems this PR should only include one commit, the most recent one. The other 2 commits are a repeated from the base branch `feature/implement_notification_foreground_handler_with_display_option`

mikechoch:
@emawby and I cherry picked this and we will be closing it in the future. The cherry picked commits went into [PR 687](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/687)

emawby:
The changes in this PR have been merged to #687 
d7e595079598c9f3f29dc3cd25e86e06b6290e98 5dea464d0b531764c6d2010cdc9db5d60dd1b65a

PR #708, opened 2020-07-09
Adding XCFramework version of the SDK

emawby:
This PR adds an XCFramework version of the OneSignal SDK. Using the XCFramework version of the SDK allows Mac Catalyst support and simulator support using the same architecture (x86_64). This means OneSignal now supports Mac Catalyst for Swift apps.

The XCFramework is currently only being distributed by Cocoapods. Carthage support for XCFrameworks is not yet available and SwiftPM support is only available in the Xcode 12 beta. 

The PR includes

1.  A new build script and target to build the XCFramework for iOS devices, simulators, and Mac Catalyst
2. The podspec for the XCFramework version of the OneSignal SDK
3. A compiled version of the XCFramework based on version 2.14.3

**Known Issue**
The released version of Cocoapods does not properly setup the XCFramework for the Notification Service Extension. The unreleased master version of Cocoapods fixes this issue so it should be fixed in the next release. You can clone and use the local master version of Cocoapods in the interim.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/708)


emawby:
> LGTM, seems some test are failing, does them fail locally, or are them some flaky tests?

They aren't failing locally so I think they might be flaky

OneSignalXCFramework.podspec
```diff
@@ -0,0 +1,17 @@
...
+    s.summary          = "OneSignal push notification library for mobile apps."
+    s.homepage         = "https://onesignal.com"
+    s.license          = { :type => 'MIT', :file => 'LICENSE' }
+    s.author           = { "Joseph Kalash" => "joseph@onesignal.com", "Josh Kasten" => "josh@onesignal.com" , "Brad Hesse" => "brad@onesignal.com"}
+    
+    s.source           = { :git => "https://github.com/OneSignal/OneSignal-iOS-SDK.git", :tag => s.version.to_s }
+    s.platform     = :ios
+    s.requires_arc = true
+    
+    s.ios.vendored_frameworks = 'iOS_SDK/OneSignalSDK/OneSignal_XCFramework/OneSignal.xcframework'
```
jkasten2, line 13: It seems `vendored_frameworks` supports an array:
* https://www.rubydoc.info/github/CocoaPods/Core/Pod%2FSpecification%2FDSL:vendored_frameworks=
* https://guides.cocoapods.org/syntax/podspec.html#group_multi_platform_support

If this works it will allow us to only have to keep one cocoapod updated.

jkasten2: Tested this and it results in both being in the pod project.
This somehow didn't create a duplicate error when building or running.  It seems cocoapods might be doing something to dedup by just picking one as it does result if both are imported manually outside of cocoapods, but we shouldn't be relying on it.
image

We will switch to only use xcframework in the future anyway. Once cocoapods 1.10 is released with it's [Revamped XCFramework Integration](http://blog.cocoapods.org/CocoaPods-1.10.0-beta/)

Issue #648, opened 2020-03-29
UIScene / SwiftUI Compatibility - WindowScene (effects In-App Message Displaying)


jkasten2:
### iOS API behavior change
In order to show your own `UIWindow` you need now set `windowScene` which is a `UIWindowScene` type if your app is using a `UIScene`, otherwise it will not show.

## Resulting OneSignal SDK Issues
This issue results in In-App Messages not showing for app that have switched to using Scenes in their app.

### Overview of SDK changes
In the OneSignal 2.13.0 SDK we add a `UIWindow` to display an IAM in [`webViewContentFinishedLoading`](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/2.13.0/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m#L640) where we need to add a line to at `self.window.windowScene`.

### Which UIWindowScene to use?
The most straightforward `UIWindowScene` to use is `UIApplication.sharedApplication.keyWindow.windowScene`. It is safe to use for apps that do not have multiple scenes.
For multiple scene apps Apple recommends to ["track windows manually"](https://developer.apple.com/videos/play/wwdc2019/212/?time=2425) instead of using `keyWindow`.

For an SDK this means picking a good default behavior, and providing an API for developers to pick the scene, if necessary. Since most apps will be single Scene apps using `keyWindow.windowScene` will be the best default. Even if the app does have multiple scenes this value will still be available and will always be a visible window. If `keyWindow` isn't picking a good window for the OneSignal In-App Message for your app you can set the key window with `makeKeyWindow`. *Please open a new issue if this behavior does not work after this ticket is closed.*

## References
* [Consumer Level - How to multitask with Split View on your iPad](https://www.youtube.com/watch?v=nSBZKr5kXYM)
* [UIApplication Deprecations](https://developer.apple.com/videos/play/wwdc2019/212/?time=2374)
* https://stackoverflow.com/questions/57060606/uiwindow-not-showing-over-content-in-ios-13
* https://developer.apple.com/documentation/uikit/app_and_environment/scenes/supporting_multiple_windows_on_ipad

PR #714, opened 2020-08-04
Fixing lifecycle methods for Scene based Applications

emawby:
### Previously 
The SDK was swizzling App Delegate lifecycle methods to add behavior for features like

- Session counts
- Session time
- IAM triggers
- Updating push permission status

This behavior was broken for Scene based applications since the following App Delegate methods are not called:

- `applicationDidBecomeActive`
- `applicationWillResignActive`
- `applicationDidEnterBackground`
- `applicationWillEnterForeground`

### Now
The SDK does  not swizzle the App Delegate lifecycle methods and instead listens for equivalent the NSNotificationCenter lifecycle notifications. If it is a scene based App we listen for:
- `UISceneDidEnterBackgroundNotification`
- `UISceneDidActivateNotification`
- `UISceneWillDeactivateNotification`

If it is not a scene based App we listen for:
- `UIApplicationDidEnterBackgroundNotification`
- `UIApplicationDidBecomeActiveNotification`
- `UIApplicationWillResignActiveNotification`

### Dev Project and Unit Test Changes
When doing development I made the development App Scene based, but I did not commit the change since we would not be able to run it on versions less than iOS 13. To make the App scene based:

1. Add the SceneDelegate.h and SceneDelegate.m files
image

2. Update the App's info.plist to include the following entries 
image

By listening for Notifications instead of Swizzling the App Delegate I needed to change our UnitTests in the following ways

- Change `foregroundApp` and `backgroundApp` in `UnitTestCommonMethods` to post a notification instead of calling the lifecycle method on the appDelegate
- Add a new common method `useSceneLifecycle:(BOOL)sceneLifecycle` that adds the "UIApplicationSceneManifest" key to the mock info.plist dictionary, meaning that we will listen for scene notifications instead of app notifications
- Add an overrider for `OSInAppMessagingViewController` to fix an internalInconsistency crash when running unit tests since the class was listening for lifecycle notifications and updating UI
- Add a new unit test that validates the scene lifecycle observers are working properly

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/714)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -81,6 +81,8 @@
 #import "OSInAppMessageAction.h"
 #import "OSInAppMessage.h"
 
+#import "OneSignalLifecycleObserver.h"
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wundeclared-selector"
 
@@ -622,6 +624,8 @@ static OneSignalOutcomeEventsController *_outcomeEventsController;
     
     if ([OneSignalTrackFirebaseAnalytics libraryExists])
         [OneSignalTrackFirebaseAnalytics init];
+    
+    [OneSignalLifecycleObserver registerLifecycleObserver];
     
     return self;
 }
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalHelper.h
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalHelper.h
@@ -27,6 +27,7 @@
 
 #import "OneSignal.h"
 #import "OneSignalWebView.h"
+#import "UIApplication+OneSignal.h"
 
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
@@ -32,6 +32,7 @@ THE SOFTWARE.
 #import "OneSignalTracker.h"
 #import "OneSignalLocation.h"
 #import "OSMessagingController.h"
+#import "UIApplication+OneSignal.h"
 
 @implementation OneSignalLifecycleObserver
 
@@ -47,16 +48,9 @@ static OneSignalLifecycleObserver* _instance = nil;
     return _instance;
 }
 
-+ (BOOL)isAppUsingUIScene {
-    if (@available(iOS 13.0, *)) {
-        return [[NSBundle mainBundle] objectForInfoDictionaryKey:@"UIApplicationSceneManifest"] != nil;
-    }
-    return NO;
-}
-
 + (void)registerLifecycleObserver {
     // Replacing swizzled lifecycle selectors with notification center observers for scene based Apps
-    if ([self isAppUsingUIScene]) {
+    if ([UIApplication isAppUsingUIScene]) {
         [self registerLifecycleObserverAsUIScene];
     } else {
         [self registerLifecycleObserverAsUIApplication];
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
@@ -1,10 +1,29 @@
-//
-//  OneSignalLifecycleObserver.m
-//  OneSignal
-//
-//  Created by Elliot Mawby on 7/30/20.
-//  Copyright © 2020 Hiptic. All rights reserved.
-//
+/**
+Modified MIT License
+
+Copyright 2020 OneSignal
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+1. The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+2. All copies of substantial portions of the Software may only be used in connection
+with services provided by OneSignal.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
 
 #import <Foundation/Foundation.h>
 #import "OneSignalLifecycleObserver.h"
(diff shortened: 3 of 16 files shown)


iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
```diff
@@ -215,14 +215,37 @@ + (void)initOneSignalAndThreadWait {
 
 + (void)foregroundApp {
     UIApplicationOverrider.currentUIApplicationState = UIApplicationStateActive;
-    UIApplication *sharedApp = [UIApplication sharedApplication];
-    [sharedApp.delegate applicationDidBecomeActive:sharedApp];
+    
+    if (@available(iOS 13.0, *)) {
```
jkasten2, line 219: I see we have this iOS 13 and sceneManifest check in at least three spots. Could you make this a single method we could call? Such as `isAppUsingUIScene`?

iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
```diff
@@ -0,0 +1,115 @@
...
+    [[NSNotificationCenter defaultCenter] addObserver:[OneSignalLifecycleObserver sharedInstance] selector:@selector(didBecomeActive) name:UIApplicationDidBecomeActiveNotification object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:[OneSignalLifecycleObserver sharedInstance] selector:@selector(willResignActive) name:UIApplicationWillResignActiveNotification object:nil];
+}
+
++ (void)removeObserver {
+    [[NSNotificationCenter defaultCenter] removeObserver:[OneSignalLifecycleObserver sharedInstance]];
+}
+     
+- (void)didBecomeActive {
+    [OneSignal onesignal_Log:ONE_S_LL_INFO message:@"application/scene didBecomeActive"];
```
jkasten2, line 87: Sorry poor wording on last my comment, this should be verbose or debug. The fact on travis is going over it's max log means we should change what we are setting as for the setLogLevel in the tests

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -623,6 +625,8 @@ + (id)initWithLaunchOptions:(NSDictionary*)launchOptions
     if ([OneSignalTrackFirebaseAnalytics libraryExists])
         [OneSignalTrackFirebaseAnalytics init];
     
+    [OneSignalLifecycleObserver registerLifecycleObserver];
```
jkasten2, line 628: I think we are seeing so many logs of "application/scene didBecomeActive" in the tests since this init gets called for most tests and is duplicating calls.

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
```diff
@@ -215,14 +216,35 @@ + (void)initOneSignalAndThreadWait {
 
 + (void)foregroundApp {
     UIApplicationOverrider.currentUIApplicationState = UIApplicationStateActive;
-    UIApplication *sharedApp = [UIApplication sharedApplication];
-    [sharedApp.delegate applicationDidBecomeActive:sharedApp];
+    
+    if ([OneSignalLifecycleObserver isAppUsingUIScene]) {
+        if (@available(iOS 13.0, *)) {
```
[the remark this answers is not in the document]
emawby: It is needed to get rid of an Xcode warning about only be available in iOS 13. I could suppress the warning with a pragma if that is preferred.

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
```diff
@@ -215,14 +216,35 @@ + (void)initOneSignalAndThreadWait {
...
+        [[NSNotificationCenter defaultCenter] postNotificationName:UIApplicationDidBecomeActiveNotification object:nil];
+    }
 }
 
 + (void)backgroundApp {
     UIApplicationOverrider.currentUIApplicationState = UIApplicationStateBackground;
-    UIApplication *sharedApp = [UIApplication sharedApplication];
-    [sharedApp.delegate applicationWillResignActive:sharedApp];
+    if ([OneSignalLifecycleObserver isAppUsingUIScene]) {
+        if (@available(iOS 13.0, *)) {
```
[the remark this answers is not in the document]
emawby: Same as above

iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
emawby: This was due to setting in the log level in the NotificationExtensionHandler

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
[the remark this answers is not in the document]
jkasten2: Ya don't think Xcode is smart enough to detect `isAppUsingUIScene` is already checking for iOS 13. I am ok with leaving this if check here then

iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
```diff
@@ -0,0 +1,115 @@
...
+            _instance = [[OneSignalLifecycleObserver alloc] init];
+        }
+    }
+    
+    return _instance;
+}
+
++ (BOOL)isAppUsingUIScene {
+    if (@available(iOS 13.0, *)) {
+        return [[NSBundle mainBundle] objectForInfoDictionaryKey:@"UIApplicationSceneManifest"] != nil;
```
jkasten2, line 52: Confirmed with Apple's docs this is the source of truth if the app is setup as a scene.

> The presence of this key indicates that the app supports scenes and does not use an app delegate object to manage transitions to and from the foreground or background.
https://developer.apple.com/documentation/bundleresources/information_property_list/uiapplicationscenemanifest

iOS_SDK/OneSignalSDK/Source/OneSignal.m
jkasten2: This isn't showing up now in the logs due to the default level is now changed. However if logging is turned up this prints many times per event as it subscribes more than once.

Since this is below the [`initDone` check above](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/714/files#diff-bcb6ac33175109d2d958d0e4391f7aa2R595-R597) this won't be an issue on in a real app.
However thinking we clear the observers in between each test so this is close to a device.

iOS_SDK/OneSignalSDK/Source/UIApplication+OneSignal.m
```diff
@@ -0,0 +1,55 @@
...
+#import "UIApplication+OneSignal.h"
+#import "OneSignalCommonDefines.h"
+
+@implementation UIApplication (OneSignal)
+
++ (BOOL)applicationIsActive {
+    if ([self isAppUsingUIScene]) {
+        if (@available(iOS 13.0, *)) {
+            
+            for (UIWindow *window in [UIApplication sharedApplication].windows) {
```
jkasten2, line 37: We do this in a different way by using `UIApplication.sharedApplication.keyWindow.windowScene` instead.
https://github.com/OneSignal/OneSignal-iOS-SDK/blob/a0aa0e6fb482d649fdb781975c86cb24591d3239/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m#L689-L698

See these notes as well
https://github.com/OneSignal/OneSignal-iOS-SDK/issues/648

emawby: Fixed this!

jkasten2: 👍

iOS_SDK/OneSignalSDK/Source/OneSignal.m
jkasten2: This is only an issue with the tests.
It doesn't seem to be breaking any of the test from what I see but it creates a lot of duplicated events since the observer gets registered over and over which creates a lot of log entries if you turn up the logging to VERBOSE

iOS_SDK/OneSignalSDK/Source/OneSignalHelper.h
```diff
@@ -27,6 +27,7 @@
 
 #import "OneSignal.h"
 #import "OneSignalWebView.h"
+#import "UIApplication+OneSignal.h"
```
Jeasmine, line 30: is this import needed?

iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
```diff
@@ -0,0 +1,109 @@
...
+        [OneSignalLocation onFocus:YES];
+        [[OSMessagingController sharedInstance] onApplicationDidBecomeActive];
+    }
+}
+
+- (void)willResignActive {
+    [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:@"application/scene willResignActive"];
+    
+    if ([OneSignal app_id])
+            [OneSignalTracker onFocus:YES];
```
Jeasmine, line 94: can you reformat this code?

iOS_SDK/OneSignalSDK/Source/OneSignalNotificationServiceExtensionHandler.m
```diff
@@ -41,7 +41,7 @@ @implementation OneSignalNotificationServiceExtensionHandler
 + (UNMutableNotificationContent*)didReceiveNotificationExtensionRequest:(UNNotificationRequest*)request
                                          withMutableNotificationContent:(UNMutableNotificationContent*)replacementContent {
     // Set default log level of NSE to VERBOSE so we get all logs from NSE logic
-    [OneSignal setLogLevel:ONE_S_LL_VERBOSE visualLevel:ONE_S_LL_NONE];
+    //[OneSignal setLogLevel:ONE_S_LL_VERBOSE visualLevel:ONE_S_LL_NONE];
```
Jeasmine, line 44: is this ok?

iOS_SDK/OneSignalSDK/Source/UIApplication+OneSignal.m
```diff
@@ -0,0 +1,53 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#import "UIApplication+OneSignal.h"
+#import "OneSignalCommonDefines.h"
+
+@implementation UIApplication (OneSignal)
+
++ (BOOL)applicationIsActive {
```
Jeasmine, line 33: should we start using this method instead of == UIApplicationStateActive? there are different places where we use that check

iOS_SDK/OneSignalSDK/Source/OneSignalHelper.h
emawby: I added it so that we can easily use the category throughout the SDK. We shouldn't rely on UIApplication's state throughout the SDK we should use this to account for scene based apps

iOS_SDK/OneSignalSDK/Source/OneSignalNotificationServiceExtensionHandler.m
emawby: We need to do this or the unit tests print too much and break CircleCI

iOS_SDK/OneSignalSDK/Source/UIApplication+OneSignal.m
emawby: We should but I don't want to increase the scope of this fix too much if the change isn't going to solve any issues. I added the category to OneSignalHelper to make it easier in the future

iOS_SDK/OneSignalSDK/Source/OneSignalLifecycleObserver.m
emawby: fixed!

iOS_SDK/OneSignalSDK/Source/OneSignalNotificationServiceExtensionHandler.m
Jeasmine: gotcha, but I think we want to keep this setLogLevel, Travis will print verbose not debug, so if there are logs that were added maybe change to debug?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
jkasten2: 👍 I see we have `[OneSignalLifecycleObserver removeObserver];` in the tests now which clean up observers carrying over from test to test.

iOS_SDK/OneSignalSDK/Source/OneSignalNotificationServiceExtensionHandler.m
jkasten2: This is the line where we are setting the log level for the tests.
https://github.com/OneSignal/OneSignal-iOS-SDK/blob/2e4a36cf7dc2edd3ffcf525d9f28edaf0011ad72/iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m#L156

The line noted in this specific comment is in the NSE, which depending on the run order of the tests would turn up the log level.

At first I thought this was our dev app but this is our SDK source. Let's remove this line as the comment above it. I think log level setting is best to live in the NSE code in our setup guide.

emawby: Removed this line and updated the verbose message below it. Also updated the iOS setup

PR #718, opened 2020-08-05
Fixing webviews not being cleared properly

emawby:
Webviews were not being cleared properly if multiple were queued in a row. #606 

In `OneSignalWebView's` `showInApp` method we were attempting to reuse `navController` and `viewControllerForPresentation` when opening webviews in the App. However the `presentationController delegate` lifecycle methods were not being properly called for the second webivew after removing the `viewControllerForPresentation.view` from its superview. 

This was because upon clearing the `navController` its presentationController delegate would get reset, and we only set the delegate if the `navController` was nil. The fix is to always set the `navController.presentationController.delegate = self`

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/718)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
@@ -105,8 +105,9 @@ UIViewController *viewControllerForPresentation;
     if (!navController) {
         navController = [[UINavigationController alloc] initWithRootViewController:self];
         navController.modalTransitionStyle = UIModalTransitionStyleCoverVertical;
-        navController.presentationController.delegate = self;
     }
+    navController.presentationController.delegate = self;
+
     if (!viewControllerForPresentation) {
         viewControllerForPresentation = [[UIViewController alloc] init];
         [[viewControllerForPresentation view] setBackgroundColor:[UIColor clearColor]];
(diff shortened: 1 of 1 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
```diff
@@ -105,8 +105,9 @@ - (void)showInApp {
     if (!navController) {
         navController = [[UINavigationController alloc] initWithRootViewController:self];
         navController.modalTransitionStyle = UIModalTransitionStyleCoverVertical;
-        navController.presentationController.delegate = self;
     }
+    navController.presentationController.delegate = self;
```
Jeasmine, line 109: can you add a comment to explain this? for the future 😬

iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
```diff
@@ -122,8 +123,9 @@ - (void)showInApp {
     
     UIWindow* mainWindow = [[UIApplication sharedApplication] keyWindow];
     
-    if (!viewControllerForPresentation.view.superview)
-        [mainWindow addSubview:[viewControllerForPresentation view]];
+    if (!viewControllerForPresentation.view.superview) {
+         [mainWindow addSubview:[viewControllerForPresentation view]];
```
Jeasmine, line 127: is this well formatted?

iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
jkasten2: This looks like a standard delegate setup, I am ok without having a comment in code
https://developer.apple.com/documentation/uikit/uipresentationcontroller/1618329-delegate

iOS_SDK/OneSignalSDK/Source/OneSignalWebView.m
jkasten2: ah yes, there is an extra space here

emawby: Fixed this one!

PR #719, opened 2020-08-05
Fixing notification received when App is Inactive

emawby:
When a notification is received when the App is:
1. Open in the foreground
2. Inactive (the notification center or control center is pulled up over the App)

OneSignal was treating the App as being opened from the Background.

This PR fixes the Inactive state by not treating the notification as being opened. The notification received handler still fires.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/719)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -2243,15 +2243,14 @@ static NSString *_lastnonActiveMessageId;
         }
     }
     // Method was called due to a tap on a notification - Fire open notification
-    else if (application.applicationState != UIApplicationStateBackground) {
+    else if (application.applicationState == UIApplicationStateActive) {
         [OneSignalHelper lastMessageReceived:userInfo];
         
-        if (application.applicationState == UIApplicationStateActive)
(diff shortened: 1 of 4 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -2243,15 +2243,14 @@ + (BOOL)remoteSilentNotification:(UIApplication*)application UserInfo:(NSDiction
         }
     }
     // Method was called due to a tap on a notification - Fire open notification
```
jkasten2, line 2245: Changes themselves look good here to fix the inactive state.

Could you test opening a notification in the following states to ensure it still works.
* Tapping on a notification that cold starts the app. Swipe away the app, then tapping on a notification.
  - Possible this could be consider `UIApplicationStateInactive`
* Tapping on a notification while the app is in the foreground
* Tapping on a notification while the app is in the background

jkasten2: Tapping on a notification that cold starts the app ends up being `UIApplicationStateActive` when this event fires

PR #722, opened 2020-08-11
Add migration when application updated but not launched yet

Jeasmine:
* If the application was upgraded but not launched and then a notification arrives,
    OneSignalNotificationServiceExtensionHandler, onNotificationReceived will fail due to accessing saved
    notifications ids without migration

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/722)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
@@ -54,23 +54,27 @@ THE SOFTWARE.
 - (void)migrateToVersion_02_14_00_AndGreater {
     let influenceVersion = 21400;
     let uniqueCacheOutcomeVersion = 21403;
-    let sdkVersion = [OneSignalUserDefaults.initShared getSavedIntegerForKey:OSUD_CACHED_SDK_VERSION defaultValue:0];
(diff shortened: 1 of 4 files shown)


iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -56,21 +56,25 @@ - (void)migrateToVersion_02_14_00_AndGreater {
     let uniqueCacheOutcomeVersion = 21403;
     let sdkVersion = [OneSignalUserDefaults.initShared getSavedIntegerForKey:OSUD_CACHED_SDK_VERSION defaultValue:0];
     if (sdkVersion < influenceVersion) {
-        [OneSignal onesignal_Log:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Migrating OSIndirectNotification from version: %d", sdkVersion]];
+        [OneSignal onesignal_Log:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Migrating OSIndirectNotification from version: %ld", sdkVersion]];
```
emawby, line 59: Changing it to %ld is correct but it is getting a warning because it thinks sdkVersion should be an int. Instead of `let sdkVersion` lets do `long sdkVersion` to get rid of the warnings.

PR #728, opened 2020-08-25
Fix register user race condition

jkasten2:
* Fixes scenario where user registration could be attempted before SDK finished initializing.
* Also added more logging to debug any related future issues better

Recommend reviewing commit by commit

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/728)


jkasten2:
Cleaned up history with `git rebase --interactive --autosquash master`

PR #733, opened 2020-08-28
Change set launch options to init

emawby:
This PR changes `setLaunchOptions` to `init:launchOptions` making it more clear how to properly initialize the OneSignal SDK. Additional parameters should not be added to this init unless they are essential for OneSignal functionality and are required on every launch of the App

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/733)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.h
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.h
@@ -469,7 +469,7 @@ extern NSString* const ONESIGNAL_VERSION;
 
 #pragma mark Initialization
 + (void)setAppId:(NSString* _Nonnull)newAppId;
-+ (void)setLaunchOptions:(NSDictionary* _Nullable)launchOptions;
++ (void)initWithLaunchOptions:(NSDictionary* _Nullable)launchOptions;
 // TODO: Remove before releasing major release 3.0.0
 + (void)setAppSettings:(NSDictionary* _Nonnull)settings;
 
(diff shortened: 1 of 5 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -167,9 +167,9 @@ @implementation OneSignal
 // Make sure launchOptions have been set
 // We need this BOOL because launchOptions can be null so simply null checking
 //  won't validate whether or not launchOptions have been set
-static BOOL hasSetLaunchOptions = false;
+static BOOL hasInitWithLaunchOptions = false;
```
jkasten2, line 170: The name `hasInitWithLaunchOptions` implies to me that launchOptions was `non-nil`.
Could be called `hasCalledInitWithLaunchOptions` but if all we are looking for is `init` was call then thinking we can just keep it simple and call it `hasCalledInit`.

iOS_SDK/OneSignalSDK/Source/OneSignal.h
```diff
@@ -469,7 +469,7 @@ extern NSString* const ONESIGNAL_VERSION;
 
 #pragma mark Initialization
 + (void)setAppId:(NSString* _Nonnull)newAppId;
-+ (void)setLaunchOptions:(NSDictionary* _Nullable)launchOptions;
++ (void)init:(NSDictionary* _Nullable)launchOptions;
```
jkasten2, line 472: Should this be `initWithLaunchOptions:` to be more consistent with Objective-C naming? But we should also consider what it will look like for Swift too. If we go with `initWithLaunchOptions:` would it result in `OneSignal.init(launchOptions)` in Swift?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
emawby: I want to distinguish between initDone and this boolean so I will go with `hasCalledInitWithLaunchOptions `

PR #734, opened 2020-08-28
Making OSDevice set its state on init

emawby:
This PR makes `OSDevice` retain its own state that is set on `init` and never changed, instead of always asking OneSignal for the most up to date user states. 

`getUserDevice` will always return a new instance of `OSDevice`

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/734)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSDevice.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSDevice.m
@@ -30,36 +30,26 @@
 
 @implementation OSDevice
 
-- (BOOL)isNotificationEnabled {
-    return [[[OneSignal getPermissionSubscriptionState] permissionStatus] reachable];
-}
-
-- (BOOL)isUserSubscribed {
-    return [[[OneSignal getPermissionSubscriptionState] subscriptionStatus] userSubscriptionSetting];
-}
-
-- (BOOL)isSubscribed {
-    return [[[OneSignal getPermissionSubscriptionState] subscriptionStatus] subscribed];
-}
-
-- (OSNotificationPermission)getNotificationPermissionStatus {
-    return [[[OneSignal getPermissionSubscriptionState] permissionStatus] status];
-}
-
-- (NSString *)getUserId {
-    return [[[OneSignal getPermissionSubscriptionState] subscriptionStatus] userId];
-}
-
-- (NSString *)getPushToken {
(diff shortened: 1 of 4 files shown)


jkasten2:
> I think we want to have [OneSignal getPermissionSubscriptionState] as a constructor parameter, correct me if I'm wrong @jkasten2 @rgomezp

Ya, that would make it more clear that it is just an object to provide this state and hides less inside of `OSDevice`. Does follow the rules of separation of responsibilities closer too.

iOS_SDK/OneSignalSDK/Source/OSDevice.m
```diff
@@ -30,36 +30,30 @@
...
-- (NSString *)getEmailAddress {
-    return [[[OneSignal getPermissionSubscriptionState] emailSubscriptionStatus] emailAddress];
+- (id)init {
+    return [[OSDevice alloc] initWithSubscriptionState:[OneSignal getPermissionSubscriptionState]];
+}
+
+- (instancetype)initWithSubscriptionState:(OSPermissionSubscriptionState *)state {
+    self = [super init];
+    if (self) {
+        _notificationEnabled = [[state permissionStatus] reachable];
```
Jeasmine, line 40: maybe an Objetive C question, this will assign the value of reachable to the notificationEnabled property or will this act as a pointer and whenever notificationEnabled is called it  will call reachable getter?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -320,11 +320,8 @@ + (void)setMSubscriptionStatus:(NSNumber*)status {
     mSubscriptionStatus = [status intValue];
 }
 
-static OSDevice* _userDevice;
 + (OSDevice *)getUserDevice {
-    if (!_userDevice)
-        _userDevice = [OSDevice new];
-    return _userDevice;
+    return [[OSDevice alloc] initWithSubscriptionState:[OneSignal getPermissionSubscriptionState]];
```
Jeasmine, line 324: I imagined we will still have this inited once and have getter internally on OSDevice accessing permissionStatus and  subscriptionStatus, what do you think? May I be missing something?

emawby: If we only init once then there is no change from the previous behavior. The OSDevice will still be the same object that will have the state changing underneath it

Jeasmine (approved):

PR #737, opened 2020-09-04
Foreground handler with completion block

emawby:
This PR removes the `complete` method from the `OSNotificationGenerationJob` and instead sends a completion block that takes a `OSNotificationDisplayType` parameter. 

**Questions:** 
1. Should we rename `OSNotificationGenerationJob`?
2. Should the `displayType` property on `OSNotificationGenerationJob` remain mutable?
  Currently a developer can change the `displayType` to silent, but have the block timeout and we will respect the `displayType` change
  We could make `displayType` readonly and not allow developers to change the displayType unless they call the completion block.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/737)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.h
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.h
@@ -196,8 +196,6 @@ typedef NS_ENUM(NSUInteger, OSNotificationDisplayType) {
 /* The message body */
 @property(readonly, nullable)NSString *body;
 
-// Method controlling completion from the notificationWillShowInForegroundHandler
-// If 'complete' is not called within 25 seconds of receiving the OSNotificationGenerationJob in notificationWillShowInForegroundHandler then 'complete' will be automatically fired.
 @end
 
 @interface OSNotificationOpenedResult : NSObject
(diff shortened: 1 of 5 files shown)


Jeasmine:
Thinking about the future, is there any other data that make sense to return on the completion? Make sense to have the user an object with the new configuration, and that one has the display type?

From my point of view, OSNotificationGenerationJob at least on Android can be confused as a background thread. So I'm a supporter of renaming it.

Given that the display type is returned on the completion I think make sense to avoid mutability, in that way is clear that the completion handler is for modifications.

emawby:
> Thinking about the future, is there any other data that make sense to return on the completion? Make sense to have the user an object with the new configuration, and that one has the display type?
> 
> From my point of view, OSNotificationGenerationJob at least on Android can be confused as a background thread. So I'm a supporter of renaming it.
> 
> Given that the display type is returned on the completion I think make sense to avoid mutability, in that way is clear that the completion handler is for modifications.

I don't think there is any other data we need from the user if the only thing that they can change is the display type.

iOS_SDK/OneSignalSDK/Source/OneSignal.h
```diff
@@ -178,11 +178,11 @@ typedef NS_ENUM(NSUInteger, OSNotificationDisplayType) {
...
 @end
 
-/* OneSignal OSNotificationGenerationJob used in notificationWillShowInForegroundHandler. The display type for the notification can be changed before it is presented.*/
-@interface OSNotificationGenerationJob : NSObject
+/* OneSignal OSPredisplayNotification used in notificationWillShowInForegroundHandler. The display type for the notification can be changed before it is presented.*/
+@interface OSPredisplayNotification : NSObject
 
 /* Display method of the notification */
-@property(nonatomic)OSNotificationDisplayType displayType;
+@property(readonly, nonatomic)OSNotificationDisplayType displayType;
```
jkasten2, line 185: We don't need the `OSNotificationDisplayType` on this `OSPredisplayNotification`since it something you pass to the completion block.

emawby: Removed!

jkasten2 (approved):

Jeasmine (approved):

PR #742, opened 2020-09-16
Custom base url

luis901101:
This is a feature request to allow developers to opt in to use a proxy server as OneSignal api base url. It happen that OneSignal domains are blocked in some countries and by some ISPs… so I think this could be a good way to allow devs to use OneSignal.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/742)


luis901101 (commented):
I just reverted my changes locally to check if all the unit-tests passes on clear master branch and it failed the exact same one as the one indicated by the travis-ci build. So I think is not something related to my changes.
What can I do next?

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/742#-:-MHmcauE5p8YRsJ_rQju:bgezw33)* status: 0 of 6 files reviewed, all discussions resolved

PR #745, opened 2020-09-19
Fixing redisplaying for dynamic trigger IAMs

emawby:
This PR allows IAMs with only dynamic triggers to be redisplayed by setting `message.isTriggerChanged = YES;` when the trigger fires. Additionally `evaluateMessages` now evaluates message triggers before setting redisplay data

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/745)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
@@ -697,8 +697,18 @@ static BOOL _isInAppMessagingPaused = false;
     }
 }
 
+- (void)makeRedisplayMessagesAvailableWithTriggers:(NSArray<NSString *> *)triggerIds {
+    for (OSInAppMessage *message in self.messages) {
+        if ([self.redisplayedInAppMessages objectForKey:message.messageId]
+            && [_triggerController hasSharedTriggers:message newTriggersKeys:triggerIds]) {
+            message.isTriggerChanged = YES;
+        }
+    }
+}
+
 #pragma mark OSTriggerControllerDelegate Methods
-- (void)triggerConditionChanged {
+- (void)triggerConditionChanged:(NSString *)triggerId {
+    [self makeRedisplayMessagesAvailableWithTriggers:@[triggerId]];
     // We should re-evaluate all in-app messages
     [self evaluateMessages];
 }
--- a/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
@@ -493,8 +493,11 @@ static BOOL _isInAppMessagingPaused = false;
 
 - (void)persistInAppMessageForRedisplay:(OSInAppMessage *)message {
     // If the IAM doesn't have the re display prop or is a preview IAM there is no need to save it
-    if (![message.displayStats isRedisplayEnabled] || message.isPreview)
-      return;
+    if (![message.displayStats isRedisplayEnabled] || message.isPreview) {
+        [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:[NSString stringWithFormat:@"not persisting %@",message.displayStats]];
+        return;
+    }
+
 
     let displayTimeSeconds = self.dateGenerator();
     message.displayStats.lastDisplayTime = displayTimeSeconds;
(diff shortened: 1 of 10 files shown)


jkasten2 (approved):

iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
```diff
@@ -697,7 +699,27 @@ - (void)addKeySceneToWindow:(UIWindow*)window {
     }
 }
 
+- (void)dynamicTriggerCompleted:(NSString *)triggerId {
+    [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:[NSString stringWithFormat:@"messageDynamicTriggerCompleted called with triggerId: %@", triggerId]];
+    [self makeRedisplayMessagesAvailableWithTriggers:@[triggerId]];
+}
+
+- (void)makeRedisplayMessagesAvailableWithTriggers:(NSArray<NSString *> *)triggerIds {
```
Jeasmine, line 707: this logic is deferring a bit from Android, do we want to come with a common logic?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1712,6 +1712,7 @@ + (void)registerUserInternal {
     }
     
     [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:@"Calling OneSignal create/on_session"];
+    sessionLaunchTime = [NSDate date];
```
Jeasmine, line 1715: good catch! I need to add this on Android

iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
```diff
@@ -697,7 +699,27 @@ - (void)addKeySceneToWindow:(UIWindow*)window {
...
+            && [_triggerController hasSharedTriggers:message newTriggersKeys:triggerIds]) {
+            message.isTriggerChanged = YES;
+        }
+    }
+}
+
 #pragma mark OSTriggerControllerDelegate Methods
+- (void)triggerConditionChanged:(NSString *)triggerId {
+    [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:[NSString stringWithFormat:@"trigger condition changed for triggerId: %@", triggerId]];
+    [self makeRedisplayMessagesAvailableWithTriggers:@[triggerId]];
```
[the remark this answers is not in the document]
emawby: It isn't working properly if I remove this line, but I am struggling to figure out why. Still testing

iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
emawby: After working on it I think that not duplicating the state in the redisplayedMessages might be better what do you think?

Jeasmine: I know that can be confusing having 2 different lists, although I see that having the redisplay messages separated can optimize the iteration. Because on worst case redisplay list will be the same length as the iam list, but in the best case redisplay iams will be empty. But is also true that there are not huge lists, or they shouldn't be, so there is not that much impact on the performance. So if you see that is more readable to only iterate iams list then I'm ok with removing the redisplay iteration for trigger change check

jkasten2:
There is a bug on master this PR also happens to address
1.  IAM with one custom trigger
2. "Every time trigger conditions are satisfied" select
3. IAM shown the first time with addTrigger
4. And again if addTrigger is called again with the same key and value as expected
5. Background the app then foreground the app
6. Call addTrigger again but issues is it won't display again.

iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
emawby: Right I think having it more readable and not trying to synchronize the state is the better approach here. Sorry for making you implement it in the first place!

jkasten2 (approved):

iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
Jeasmine: ok agree then, will make the change under major release branch

Jeasmine (approved):

PR #747, opened 2020-09-21
Add json representation

emawby:
This PR adds a `jsonRepresentation` public method to OSDeviceState and OSNotificationOpenedResult to make them more easily used from wrappers

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/747)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSDeviceState.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSDeviceState.m
@@ -27,6 +27,8 @@
 
 #import <Foundation/Foundation.h>
 #import "OneSignal.h"
+#import "OneSignalHelper.h"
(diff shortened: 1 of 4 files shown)


jkasten2 (commented):
JSON property style should match between iOS and Android.

We went with snake case instead of camel casing for JSON generated from Android:
https://github.com/OneSignal/OneSignal-Android-SDK/pull/1147/files#diff-fef47f52cdc16c2b17920592c37c8630R119-R135

JSON doesn't a standard here so we just need to pick one.

Thinking snake case would be easier to search for if we need to string search on where the rubber hits the road between wrapper SDK.

jkasten2 (approved):
Stick with camel case

PR #751, opened 2020-09-24
Default auto prompt false

jkasten2:
## Dropped kOSSettingsKeyAutoPrompt
This change defaults the logic of `kOSSettingsKeyAutoPrompt` to `false` and drops the public API.
Removed `kOSSettingsKeyAutoPrompt` internally but kept `usesAutoPrompt` in this PR to minimize logic changes.
It can be cleaned up in a future PR where more testing can be done

## Reviewer
Recommend reviewing commit by commit

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/751)

### DIFF
--- a/Examples/ObjectiveCExample/OneSignalDemo/AppDelegate.m
+++ b/Examples/ObjectiveCExample/OneSignalDemo/AppDelegate.m
@@ -90,7 +90,7 @@
     };
     
     // (Optional) - Configuration options for OneSignal settings.
-    id oneSignalSettings = @{kOSSettingsKeyInAppLaunchURL : @NO, kOSSettingsKeyAutoPrompt : @YES};
+    id oneSignalSettings = @{kOSSettingsKeyInAppLaunchURL : @NO};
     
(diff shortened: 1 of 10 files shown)


emawby (approved):
lgtm

Jeasmine (commented):
If I understand this right, we will never auto prompt. The user can't configure auto prompting, or will the user be able to configure it still?

jkasten2:
@Jeasmine Auto prompting for permission app start is a bad practice so we are removing this. It also simpiles things as way have an "auto flag" for something when you can just call the method `OneSignal.promptForPushNotifications(userResponse: {})` instead.

Jeasmine:
> @Jeasmine Auto prompting for permission on app start is a bad practice so we are removing this. It also simpiles things as way have an "auto flag" for something when you can just call the method `OneSignal.promptForPushNotifications(userResponse: {})` instead.

good, just checking! Thxs! LGTM

Jeasmine (approved):

PR #755, opened 2020-09-24
Updating example project for setSubscription to disablePush change

emawby:
Using disablePush instead of setSubscription in example projects

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/755)


jkasten2 (changes requested):
I thinking about it would be clearer if the UI said "OneSignal Push Enabled" / "OneSignal Push Disabled" instead?

Tested the swift project still seeing "Set Subscription ON" / OFF

Looks like we didn't update 
`onOSSubscriptionChanged` in the `ViewController.swift` in the example project.

emawby:
> I thinking about it would be clearer if the UI said "OneSignal Push Enabled" / "OneSignal Push Disabled" instead?
> 
> Tested the swift project still seeing "Set Subscription ON" / OFF
> 
> Looks like we didn't update
> `onOSSubscriptionChanged` in the `ViewController.swift` in the example project.

fixed!

jkasten2 (approved):

PR #761, opened 2020-10-05
Feature/support app clips

emawby:
This PR adds supports to app clips, as well as updating the dev app to include an app clip and be compatible with app store uploads.

For App Clip support we now support the `ephemeral` permission state which is granted automatically to app clips to allow push for 8 hours. There is no prompt needed for this state, but app clips can still prompt for permanent push permissions. 
`OSPermission` and `OneSignalNotificationSettingsiOS10` where updated to reflect this.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/761)


iOS_SDK/OneSignalSDK/Source/OneSignalNotificationSettingsIOS10.m
```diff
@@ -74,6 +74,12 @@ - (void)getNotificationPermissionState:(void (^)(OSPermissionState *subscription
             status.answeredPrompt = settings.authorizationStatus != UNAuthorizationStatusNotDetermined && settings.authorizationStatus != provisionalStatus;
             status.provisional = (settings.authorizationStatus == 3);
             status.accepted = settings.authorizationStatus == UNAuthorizationStatusAuthorized && !status.provisional;
+            if (@available(iOS 14.0, *)) {
+                status.ephemeral = (settings.authorizationStatus == UNAuthorizationStatusEphemeral);
```
jkasten2, line 78: `UNAuthorizationStatusEphemeral` is only available in Xcode 12. Since we need to build with older versions of Xcode to support bitcode from them we have to use the raw value by defining it as our on const or hard coded value here.

jkasten2 (changes requested):

iOS_SDK/OneSignalSDK/Source/OneSignalNotificationSettingsIOS10.m
emawby: Good catch! fixed this

iOS_SDK/OneSignalSDK/Source/OneSignalCommonDefines.h
```diff
@@ -199,6 +199,8 @@ typedef enum {GET, POST, HEAD, PUT, DELETE, OPTIONS, CONNECT, TRACE} HTTPMethod;
 #define ERROR_PUSH_NEVER_PROMPTED          -18
 #define ERROR_PUSH_PROMPT_NEVER_ANSWERED   -19
 
+#define AUTH_STATUS_EPHEMERAL 4
```
jkasten2, line 202: nit, might be easier to search for and update this later if we add a comment that this is `UNAuthorizationStatusEphemeral`

jkasten2 (approved):

PR #784, opened 2020-10-28
remove IDFA from onesignal (Major Release)

emawby:
OneSignal was using IDFA to avoid duplicate users for re-installs, but this is being removed to avoid App store rejections.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/784)


jkasten2 (approved):
Changes look good, merge after tests pass

PR #795, opened 2020-11-13
Fix IAM push prompt dismissing open settings alert

emawby:
This PR fixes an IAM push prompt when the user has previously denied push. In this case the "Open Settings" alert would be immediately dismissed. This was caused by an erroneous `dismissViewController` AND the prompt action immediately calling the completion block before the alert had actually been addressed.

To fix this I removed the extra `dismissViewController` and moved the completion block call when showing the open settings alert into the completion block for the alert

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/795)


Jeasmine (approved):

jkasten2 (approved):
The code change itself looks good.
The 3 scenarios that come to mind we should manually test are:
1. Prompting for notification permissions for the very first time from an IAM
2. Prompting for notification permissions after declined from an IAM
3. Pressing an IAM action button that closes the IAM

PR #823, opened 2020-12-03
Fix/direct outcomes

emawby:
The primary goal of this PR is to fix Direct Sessions in the major release.

The problem was that we were relying on the `foreground` variable to decide if we should upgrade the session,

```
if (!foreground) {
        OneSignal.appEntryState = NOTIFICATION_CLICK;
        [OneSignal.sessionManager onDirectInfluenceFromNotificationOpen:_appEntryState withNotificationId:messageId];
}
```

and in the major release I had changed the setting of the foreground variable to:
`let isAppForeground = [[UIApplication sharedApplication] applicationState] != UIApplicationStateBackground;`
From:
`let isAppForeground = [[UIApplication sharedApplication] applicationState] == UIApplicationStateActive;`

This was to account for the inactive state which is technically in the foreground.

The first commit on this branch just flips it back to `[[UIApplication sharedApplication] applicationState] == UIApplicationStateActive;` to fix direct outcomes.

However, I don't think we need foreground or isActive to be passed around at all anymore. Since we are using the native iOS call for willShowInForeground we don't need to track the states on notification received. We were only logging those variables and (improperly) using them for direct outcomes. So I removed the foreground and isActive variables and am checking the state directly when deciding if we need to upgrade the session for direct outcomes. 

I also added unit tests for outcomes while in the inactive state

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/823)

### DIFF
--- a/iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
+++ b/iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
@@ -61,7 +61,7 @@ NSString * serverUrlWithPath(NSString *path) {
 
 @interface OneSignal ()
 
-+ (void)notificationReceived:(NSDictionary*)messageDict foreground:(BOOL)foreground isActive:(BOOL)isActive wasOpened:(BOOL)opened;
++ (void)notificationReceived:(NSDictionary*)messageDict wasOpened:(BOOL)opened;
 
 @end
 
@@ -367,10 +367,7 @@ static XCTestCase* _currentXCTestCase;
 }
 
 + (void)handleNotificationReceived:(NSDictionary*)messageDict wasOpened:(BOOL)opened {
-    BOOL foreground = UIApplication.sharedApplication.applicationState != UIApplicationStateBackground;
-    BOOL isActive = UIApplication.sharedApplication.applicationState == UIApplicationStateActive;
-    
-    [OneSignal notificationReceived:messageDict foreground:foreground isActive:isActive wasOpened:opened];
+    [OneSignal notificationReceived:messageDict wasOpened:opened];
 }
 
 + (NSDictionary*)createNotificationUserInfo:(NSString *)notificationId {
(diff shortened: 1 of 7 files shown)


Jeasmine (approved):
Awesome! This is way more clear

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
```diff
@@ -272,6 +272,17 @@ + (void)backgroundApp {
     }
 }
 
++ (void)setAppInactive {
+    UIApplicationOverrider.currentUIApplicationState = UIApplicationStateInactive;
+}
+
++ (void)pullDownNotificationCenter {
```
jkasten2, line 279: Weird, so simply just pulling down the notification center actually backgrounds then foregrounds then *backgrounds again*?

jkasten2: I like the fact this was added as a helper method to keep the tests clean

jkasten2 (approved):
Test failing, this is ok but @emawby please confirm tests pass locally.
We can fix the log limit in another PR as it is unrelated to this one.

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
emawby: Thats what I found in my testing. I put break points in the lifecycle methods when pulling it down to see what happens it resigns active becomes active and then resigns active

emawby:
The unit tests pass locally

jkasten2 (approved):
@emawby 👍 Merge it in!

iOS_SDK/OneSignalSDK/UnitTests/UnitTestCommonMethods.m
jkasten2: Very weird, what version of iOS? It is good to have the exact behavior in test so we should keep it, however just makes me wonder if this is some recent bug Apple introduced to iOS.

PR #826, opened 2020-12-07
Late setNotificationOpenedHandler call processing

jkasten2:
### Changes
* If `OneSignal.setNotificationOpenedHandler` is called after
didFinishLaunchingWithOptions then we still want their handler to fire.
   - So basically the OneSignal SDK will hold onto the event for them
   and fire it as soon as handler is setup.
* This means more flexibility on where / when the developer wants to call
`setNotificationOpenedHandler`.
* This is the exact behavior provided in the Android SDK in this commit.
   - https://github.com/OneSignal/OneSignal-Android-SDK/commit/ce5ad7d07e9d34df207e7514dd3fff98fb26f184#diff-2708f86f77e539cab647ecede9d5fcbead3bf0f954d7f9e536eafbc78ba760d6R340-R344
   - ^ To view this click "Load diff" on `OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java`

### Manual Testing
Tested on iOS 14.2 with the following steps.
1. Send notification to device
2. Swipe away app.
   - To get app into a killed state.
3. Tap on notification.
   - This is a "cold start" thanks to step 2
4. Code setup to call `OneSignal.setNotificationOpenedHandler` with a delay of 3 seconds
```Objective-C
dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 3 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
    id openNotificationHandler = ^(OSNotificationOpenedResult *result) {
        UIAlertView *alert = [[UIAlertView alloc] initWithTitle:@"Notification Opened Event!" message:@"Notification Opened Event!" delegate:self cancelButtonTitle:@"Close" otherButtonTitles:@"Ok", nil];
        [alert show];
    };
    [OneSignal setNotificationOpenedHandler:openNotificationHandler];
});
```
5. Wait 3 seconds
6. Confirmed alert dialog showed.

### When to merge
* This can be done either before or after the final 3.0.0 release.
* However should be done before any wrapper final releases that depend on 3.0.0.
   - Otherwise wrapper will continue to need to handle this edge case, which is not ideal. Ideal too as we will be retesting this flow as well.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/826)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
@@ -214,6 +214,8 @@ static NSString *_lastMessageIdFromAction;
 NSDictionary* lastMessageReceived;
 UIBackgroundTaskIdentifier mediaBackgroundTask;
 
+static NSMutableArray<OSNotificationOpenedResult*> *unprocessedOpenedNotifis;
+
 + (void)resetLocals {
     [OneSignalHelper lastMessageReceived:nil];
     _lastMessageIdFromAction = nil;
@@ -221,6 +223,8 @@ UIBackgroundTaskIdentifier mediaBackgroundTask;
 
     notificationWillShowInForegroundHandler = nil;
     notificationOpenedHandler = nil;
+    
+    unprocessedOpenedNotifis = nil;
 }
 
 OSNotificationWillShowInForegroundBlock notificationWillShowInForegroundHandler;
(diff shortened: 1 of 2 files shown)


jkasten2:
Test failure is due to log size in TravisCI. I confirmed all tests passed locally.

iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
```diff
@@ -231,6 +235,23 @@ + (void)setNotificationWillShowInForegroundBlock:(OSNotificationWillShowInForegr
...
+    if (!unprocessedOpenedNotifis)
+        unprocessedOpenedNotifis = [NSMutableArray new];
+    return unprocessedOpenedNotifis;
+}
+
++ (void)fireNotificationOpenedHandlerForUnprocessedEvents {
+    if (!notificationOpenedHandler)
+        return;
+    
+    for(OSNotificationOpenedResult* notification in [self getUnprocessedOpenedNotifis]) {
```
Jeasmine, line 251: space after the for 🙏

iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
```diff
@@ -358,8 +379,10 @@ + (void)handleNotificationAction:(OSNotificationActionType)actionType actionID:(
     
     [OneSignalTrackFirebaseAnalytics trackOpenEvent:result];
     
-    if (!notificationOpenedHandler)
+    if (!notificationOpenedHandler) {
+        [[self getUnprocessedOpenedNotifis] addObject:result];
```
Jeasmine, line 383: maybe have this addObject in a method, to have the UnprocessedOpenedNotifis mutations centralized?

iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
jkasten2: Good catch! I also do that 🤣

jkasten2: Created fixup commit 4e0fc0c2b46a1660a4d3b706c161354e85f682f2 to address this.

iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
jkasten2: Created fixup commit 4e0fc0c2b46a1660a4d3b706c161354e85f682f2 to address this.

Jeasmine (approved):

emawby (approved):

jkasten2:
Ran `git rebase -i --autosquash major_release_3.0.0` then `git push -f` to clean up fixup commit.

PR #837, opened 2020-12-17
Use Binary Target For Swift PM

emawby:
Using a [Binary Target](https://developer.apple.com/documentation/swift_packages/distributing_binary_frameworks_as_swift_packages) for Swift PM instead of the standard SwiftPM Target primarily gets us 2 things

1. Significantly faster (minutes) Swift PM Install speed
1. A much easier path to adding Swift files within the SDK

However, Binary Targets are only available with Xcode 12 so developers on Xcode 11 would lose SwiftPM support. They could still use Cocoapods with both the Framework and XCFramework.

To use a binary target we need to:

*  update the Swift Package to use Swift tools 5.3 (this is what breaks Xcode 11)
*  use a `.binaryTarget` which requires a URL to a zipped XCFramework, and a checksum to validate the archive.
  * The checksum is computed using `swift package compute-checksum FILENAME`
* Host the archived Framework in a CDN or as an Asset in the GitHub release

This PR

- Updates the Swift.package to use the binary target
- Updates the `build_xcframework.sh` script to 
    - zip the framework once it has been built 
    - compute the checksum for the archive 
    - automatically replace the checksum in the `Swift.package` file
    - Ask for the new release version number
    - automatically replace the version in the URL in the `Swift.package` file

With the script updated in this way the only change to our release process will be to attach the zipped archive to our GitHub releases as an asset.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/837)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.1
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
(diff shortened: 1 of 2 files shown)


Jeasmine (approved):

emawby:
This PR is no longer needed. No changes need to be made to the OneSignal-iOS-SDK

PR #841, opened 2020-12-22
Fix showing IAMS past stop time

emawby:
This change involves reading in end_time from JSON when creating an IAM
so that we can decide to not show the IAM (Cached case) if the IAM
should be finished. The date comes in iso8601 format so I added a
NSDateFormatter category with that format to easily retrieve it.
Note we cannot use the Apple provided iso8601 date formatter class until
we drop iOS 9 support.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/841)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessage.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessage.m
@@ -101,6 +101,13 @@
         message.displayStats = [OSInAppMessageDisplayStats instanceWithJson:json[@"redisplay"]];
     else
         message.displayStats = [[OSInAppMessageDisplayStats alloc] init];
+    
+    if (json[@"end_time"] && [json[@"end_time"] isKindOfClass:[NSString class]]) {
+        NSString *stringEndTime = json[@"end_time"];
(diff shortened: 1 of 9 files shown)


iOS_SDK/OneSignalSDK/Source/OSInAppMessage.m
```diff
@@ -101,6 +101,13 @@ + (instancetype)instanceWithJson:(NSDictionary * _Nonnull)json {
         message.displayStats = [OSInAppMessageDisplayStats instanceWithJson:json[@"redisplay"]];
     else
         message.displayStats = [[OSInAppMessageDisplayStats alloc] init];
+    
+    if (json[@"end_time"] && [json[@"end_time"] isKindOfClass:[NSString class]]) {
+        NSString *stringEndTime = json[@"end_time"];
+        NSDateFormatter *dateFormatter = [NSDateFormatter iso8601DateFormatter];
```
Jeasmine, line 107: can we have this object on Android too?

Jeasmine (approved):

PR #842, opened 2020-12-23
Fix content available notif firing opened handler when received in foreground

emawby:
https://app.asana.com/0/1198177413178126/1199361722609493

When a content available notification **with a badge count or sound** and **without a title**  is received when the app is in the foreground it will not be displayed, but the notification opened handler will fire. This PR makes it so that the opened handler will no longer fire. 

The reason it was firing is that we were not treating it as a remote silent notification because it contains a badge count or sound.

We now check `isDisplayableNotification` before firing notification received with opened as true
```
+ (BOOL)isDisplayableNotification:(NSDictionary*)msg {
    if ([self isRemoteSilentNotification:msg]) {
        return false;
    }
    return (msg[@"title"] && [msg[@"title"] length] > 0) || (msg[@"body"] && [msg[@"body"] length] > 0);
}
```
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/842)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -2216,7 +2216,7 @@ static NSString *_lastnonActiveMessageId;
     else if (application.applicationState == UIApplicationStateActive) {
         [OneSignalHelper lastMessageReceived:userInfo];
 
-        if (![OneSignalHelper isRemoteSilentNotification:userInfo] && (userInfo[@"title"] && [userInfo[@"title"] length] > 0)) {
+        if ([OneSignalHelper isDisplayableNotification:userInfo]) {
              [OneSignal notificationReceived:userInfo wasOpened:YES];
         }
         return startedBackgroundJob;
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -2216,7 +2216,7 @@ static NSString *_lastnonActiveMessageId;
     else if (application.applicationState == UIApplicationStateActive) {
         [OneSignalHelper lastMessageReceived:userInfo];
 
-        if (![OneSignalHelper isRemoteSilentNotification:userInfo]) {
+        if (![OneSignalHelper isRemoteSilentNotification:userInfo] && (userInfo[@"title"] && [userInfo[@"title"] length] > 0)) {
              [OneSignal notificationReceived:userInfo wasOpened:YES];
         }
         return startedBackgroundJob;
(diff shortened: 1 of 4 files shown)


Jeasmine (approved):

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -2216,7 +2216,7 @@ + (BOOL)receiveRemoteNotification:(UIApplication*)application UserInfo:(NSDictio
     else if (application.applicationState == UIApplicationStateActive) {
         [OneSignalHelper lastMessageReceived:userInfo];
 
-        if (![OneSignalHelper isRemoteSilentNotification:userInfo]) {
+        if (![OneSignalHelper isRemoteSilentNotification:userInfo] && (userInfo[@"title"] && [userInfo[@"title"] length] > 0)) {
```
jkasten2, line 2219: Instead of adding an extra "title" check to this if statement thinking we should add a helper that better fits what we need to check here.

Suggested new method called `isDisplayableNotification, that will check if the notification would be displayed in the Notification Center on the device. The only requirement for this is that the notification have a message body I believe. Not sure if a notification would be displayed with only a title.

Also check may not be needed at all if we can confirm `UNNotificationCenterDelegate`'s  `userNotificationCenter:didReceiveNotificationResponse:withCompletionHandler:` does not fire for silent notifications. This code path might just be called from the AppDelegate which we may not need to process for iOS 10+.

jkasten2 (changes requested):

iOS_SDK/OneSignalSDK/Source/OneSignal.m
emawby: The notification is displayed with only a title and it is also displayed with only a message so I think we need to check both cases. I think the isDisplayableNotification is the way to go

iOS_SDK/OneSignalSDK/Source/OneSignalHelper.m
```diff
@@ -282,6 +282,13 @@ + (BOOL)isRemoteSilentNotification:(NSDictionary*)msg {
     return true;
 }
 
++ (BOOL)isDisplayableNotification:(NSDictionary*)msg {
+    if ([self isRemoteSilentNotification:msg]) {
+        return false;
+    }
+    return (msg[@"title"] && [msg[@"title"] length] > 0) || (msg[@"body"] && [msg[@"body"] length] > 0);
```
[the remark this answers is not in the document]
emawby: Ya that is what I am seeing in my testing as well. Given thats the case I believe we only need to check for anything in the alert key. If there is no title, body, or subtitle the key isn't there and the notification doesn't show. If any of those things are there the alert key is there and the notification will show. I tested with just an image attachment and the notification doesn't show.

jkasten2 (approved):

PR #735, opened 2020-09-02
Iam carousel page impressions

emawby:
This PR adds SDK support for individual IAM Carousel page impressions.

It adds a new model `OSInAppMessagePage` and bridge event type `OSInAppMessageBridgeEventTypePageChange` that reads the `pageId` from the new `page_change` event that is received from the carousel's javascript.

To send the page impression we use the new pageImpression endpoint using the new request `OSRequestInAppMessagePageViewed`
Page impressions are only sent once per Impression of a Carousel. The ids are tracked in a new set `viewedPageIds` on the `OSInAppMessage` object

Additionally we receive the pageId for click events, but do not currently use them.
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/735)

PR #843, opened 2020-12-24
Release 3.1.0

emawby:
## Added Support for IAM Carousel Per Page Impressions
* The SDK will now send `page impressions` for individual IAM Carousel pages. #735
## Fixes a bug where IAMs would be shown past their stop time. 
* Cached IAMs could be shown past their stop time. The SDK now checks if an IAM is past its stop time before it is displayed. #841

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/843)


Jeasmine (approved):
is ok to jump to 3.1.0 instead of 3.0.2?

jkasten2 (approved):

jkasten2:
@Jeasmine Good question, since this is a new feature I would say increasing the MINOR version number instead of PATCH is correct.

PR #849, opened 2021-01-05
Remove device identifier from register user dictionary

emawby:
The device identifier was being used to not create a new player record if a user had multiple apps installed from the same developer. This was different from android confusing to customers so the functionality is being removed

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/849)


Jeasmine (commented):
I'm a bit confused. How will this behave now?

emawby:
> I'm a bit confused. How will this behave now?

we will always create a new player when an app is uninstalled/reinstalled

Jeasmine (approved):

PR #850, opened 2021-01-07
Fix testSetExternalUserId_WithAuthToken_forPush_withCompletion_beforR…

Jeasmine:
Fix testSetExternalUserId_WithAuthToken_forPush_withCompletion_beforRegister

* External id can be delayed due to remote params missing or user Id not set
* If remote params are retrieved but user id is not set, we should wait until register user is called for setting the external id

emawby let me know if this interferes with https://github.com/OneSignal/OneSignal-iOS-SDK/pull/835

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/850)


emawby (approved):
Looks good! This doesn't reintroduce the bug with force quitting and coming back

PR #853, opened 2021-01-11
Fix subscription observers firing on launch

emawby:
The OSSubscriptionObserver was firing 3 times on launch for a fresh install.

1. For prompting permission
1. getting the APNs token
1. For registering the user

Now it will only fire after register user has been called and either succeeded or failed. For subsequent launches we rely on the `isOnSessionSuccessfulForCurrentState` boolean.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/853)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -834,7 +834,6 @@ static OneSignalOutcomeEventsController *_outcomeEventsController;
 + (void)downloadIOSParamsWithAppId:(NSString *)appId {
     [self onesignal_Log:ONE_S_LL_DEBUG message:@"Downloading iOS parameters for this application"];
     _didCallDownloadParameters = true;
-    
     [OneSignalClient.sharedClient executeRequest:[OSRequestGetIosParams withUserId:self.currentSubscriptionState.userId appId:appId] onSuccess:^(NSDictionary *result) {
         if (result[IOS_REQUIRES_EMAIL_AUTHENTICATION]) {
             self.currentEmailSubscriptionState.requiresEmailAuth = [result[IOS_REQUIRES_EMAIL_AUTHENTICATION] boolValue];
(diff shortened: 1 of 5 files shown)


emawby:
The branch does not include the set external user id unit test fix

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1536,6 +1535,17 @@ + (void)setIsOnSessionSuccessfulForCurrentState:(BOOL)value {
     isOnSessionSuccessfulForCurrentState = value;
 }
 
+static BOOL _registerUserFinished = false;
++ (BOOL)isRegisterUserFinished {
+    if (_registerUserFinished || isOnSessionSuccessfulForCurrentState) {
+        NSLog(@"ECM register finished");
```
jkasten2, line 1541: Should use `onesignalLog` instead of `NSLog`

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1536,6 +1535,17 @@ + (void)setIsOnSessionSuccessfulForCurrentState:(BOOL)value {
...
+    } else {
+        NSLog(@"ECM register NOT finished");
```
jkasten2, line 1543: Should use `onesignalLog` instead of `NSLog`

jkasten2 (changes requested):
* `NSLog` should not be used
* Missing test to cover new logic

iOS_SDK/OneSignalSDK/Source/OneSignal.m
emawby: my fixup didn't get force pushed I removed the logs

iOS_SDK/OneSignalSDK/Source/OneSignal.m
emawby: my fixup didn't get force pushed I removed the logs

jkasten2 (approved):

PR #858, opened 2021-01-14
Generalize Channel data access

Jeasmine:
Make the server requests more generic. This will smooth SMS integration by making every channel Synchronizer responsible for retrieving the data specific for its channel.

 OSStateSynchronizer will only handle an array of OSUserStateSynchronizer, and all requests are made with the same data.

It is recommended to review commit by commit.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/858)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
@@ -212,7 +212,7 @@ THE SOFTWARE.
     [OneSignalClient.sharedClient executeSimultaneousRequests:requests withSuccess:^(NSDictionary<NSString *, NSDictionary *> *results) {
         // The tags for email & push are identical so it doesn't matter what we return in the success block
         if (nowProcessingCallbacks) {
-            NSDictionary *resultTags = results[OS_PUSH] ?: results[OS_EMAIL];
+            NSDictionary *resultTags = [self getFirstResultByChannelPriority:results];
             
             for (OSPendingCallbacks *callbackSet in nowProcessingCallbacks)
                 if (callbackSet.successBlock)
@@ -220,9 +220,10 @@ THE SOFTWARE.
         }
     } onFailure:^(NSDictionary<NSString *, NSError *> *errors) {
         if (nowProcessingCallbacks) {
+            NSError *error = (NSError *)[self getFirstResultByChannelPriority:errors];
             for (OSPendingCallbacks *callbackSet in nowProcessingCallbacks) {
                 if (callbackSet.failureBlock) {
-                    callbackSet.failureBlock((NSError *)(errors[OS_PUSH] ?: errors[OS_EMAIL]));
+                    callbackSet.failureBlock(error);
                 }
             }
         }
@@ -290,4 +291,14 @@ THE SOFTWARE.
     }];
 }
 
+-(id)getFirstResultByChannelPriority:(NSDictionary<NSString *, id> *)results {
+    id result;
+    for (NSString* channelId in OS_CHANNELS) {
+        result = results[channelId];
+        if (result)
+            return result;
+    }
+    
+    return nil;
+}
 @end
--- a/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
@@ -83,6 +83,19 @@ THE SOFTWARE.
         return nil;
 }
 
+- (NSArray<OSUserStateSynchronizer *> * _Nonnull)getStateSynchronizers {
+    NSMutableArray *stateSynchronizers = [NSMutableArray new];
+    
+    let pushStateSyncronizer = [self getPushStateSynchronizer];
+    let emailStateSyncronizer = [self getEmailStateSynchronizer];
+    [stateSynchronizers addObject:pushStateSyncronizer];
+    
+    if (emailStateSyncronizer)
+        [stateSynchronizers addObject:emailStateSyncronizer];
+    
+    return stateSynchronizers;
+}
+
 - (void)registerUserWithState:(OSUserState *)registrationState withSuccess:(OSMultipleSuccessBlock)successBlock onFailure:(OSMultipleFailureBlock)failureBlock {
     let pushStateSyncronizer = [self getPushStateSynchronizer];
     let emailStateSyncronizer = [self getEmailStateSynchronizer];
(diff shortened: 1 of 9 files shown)


iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -204,15 +198,13 @@ - (void)sendTagsWithAppId:(NSString *)appId
               sendingTags:(NSDictionary *)tags
               networkType:(NSNumber *)networkType
       processingCallbacks:(NSArray *)nowProcessingCallbacks {
-    let pushStateSyncronizer = [self getPushStateSynchronizer];
-    let emailStateSyncronizer = [self getEmailStateSynchronizer];
-    
+    let stateSyncronizer = [self getStateSynchronizers];
+    // Begin constructing the request for the external id update
```
jkasten2, line 202: This comment isn't correct, this is a tags update. Recommend removing instead of updating comment as the method name already describes the request about to be made.

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -204,15 +198,13 @@ - (void)sendTagsWithAppId:(NSString *)appId
...
-    requests[OS_PUSH] = [pushStateSyncronizer sendTagsWithUserId:_currentSubscriptionState.userId appId:appId sendingTags:tags networkType:networkType emailAuthHashToken:nil externalIdAuthHashToken:_currentSubscriptionState.externalIdAuthCode];
-    
-    if (emailStateSyncronizer)
-        requests[OS_EMAIL] = [emailStateSyncronizer sendTagsWithUserId:_currentEmailSubscriptionState.emailUserId appId:appId sendingTags:tags networkType:networkType emailAuthHashToken:_currentEmailSubscriptionState.emailAuthCode externalIdAuthHashToken:nil];
-    
+    for (OSUserStateSynchronizer* userStateSynchronizer in stateSyncronizer) {
+        requests[userStateSynchronizer.getChannelId] = [userStateSynchronizer sendTagsWithAppId:appId sendingTags:tags networkType:networkType];
+    }
+
     [OneSignalClient.sharedClient executeSimultaneousRequests:requests withSuccess:^(NSDictionary<NSString *, NSDictionary *> *results) {
```
jkasten2, line 208: Recommend we address the duplication of these callbacks. As well it's extensibility of more channels. Example `NSDictionary *resultTags = results[OS_PUSH] ?: results[OS_EMAIL];`

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -237,27 +229,23 @@ - (void)sendPurchases:(NSArray *)purchases appId:(NSString *)appId {
     if (!_currentSubscriptionState.userId)
         return;
     
-    let pushStateSyncronizer = [self getPushStateSynchronizer];
-    let emailStateSyncronizer = [self getEmailStateSynchronizer];
-    
+    let stateSyncronizer = [self getStateSynchronizers];
+    // Begin constructing the request for the external id update
```
jkasten2, line 233: This comment isn't correct, this is a sendPurchases update. Recommend removing instead of updating comment as the method name already describes the request about to be made.

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -237,27 +229,23 @@ - (void)sendPurchases:(NSArray *)purchases appId:(NSString *)appId {
...
 
     [OneSignalClient.sharedClient executeSimultaneousRequests:requests withSuccess:nil onFailure:nil];
 }
 
 - (void)sendBadgeCount:(NSNumber *)badgeCount appId:(NSString *)appId {
-    let pushStateSyncronizer = [self getPushStateSynchronizer];
-    let emailStateSyncronizer = [self getEmailStateSynchronizer];
-    
+    let stateSyncronizer = [self getStateSynchronizers];
+    // Begin constructing the request for the external id update
```
jkasten2, line 244: This comment isn't correct, this is a bage update. Recommend removing instead of updating comment as the method name already describes the request about to be made.

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -266,14 +254,12 @@ - (void)sendLocation:(os_last_location *)lastLocation
                appId:(NSString *)appId
          networkType:(NSNumber *)networkType
      backgroundState:(BOOL)background {
-    let pushStateSyncronizer = [self getPushStateSynchronizer];
-    let emailStateSyncronizer = [self getEmailStateSynchronizer];
-    
+    let stateSyncronizer = [self getStateSynchronizers];
+    // Begin constructing the request for the external id update
```
jkasten2, line 258: This comment isn't correct, this is a location update. Recommend removing instead of updating comment as the method name already describes the request about to be made.

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -282,16 +268,18 @@ - (void)sendOnFocusTime:(NSNumber*)totalTimeActive
                  params:(OSFocusCallParams *)params
             withSuccess:(OSMultipleSuccessBlock)successBlock
               onFailure:(OSMultipleFailureBlock)failureBlock {
-    let pushStateSyncronizer = [self getPushStateSynchronizer];
-    let emailStateSyncronizer = [self getEmailStateSynchronizer];
-    
+    let stateSyncronizer = [self getStateSynchronizers];
+    // Begin constructing the request for the external id update
```
jkasten2, line 272: This comment isn't correct, this is a focus update. Recommend removing instead of updating comment as the method name already describes the request about to be made.

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -81,27 +81,29 @@ - (OSUserStateSynchronizer *)getEmailStateSynchronizer {
...
-    pushDataDic[@"identifier"] = _currentSubscriptionState.pushToken;
+    if (emailStateSyncronizer)
+        [stateSynchronizers addObject:emailStateSyncronizer];
     
+    return stateSynchronizers;
+}
+
+- (void)registerUserWithState:(OSUserState *)registrationState withSuccess:(OSMultipleSuccessBlock)successBlock onFailure:(OSMultipleFailureBlock)failureBlock {
+    let stateSyncronizer = [self getStateSynchronizers];
+    // Begin constructing the request for the external id update
```
jkasten2, line 99: This comment isn't correct, this is a register request. Recommend removing instead of updating comment as the method name already describes the request about to be made.

jkasten2 (changes requested):

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
Jeasmine: good catch!

Jeasmine: I made a method for parsing the results, I will love to have a more generic way of parsing each data from each channel, but that will take more work. I think like this can proceed, and in other phase improve more!

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -290,4 +291,22 @@ - (void)sendOnFocusTime:(NSNumber*)totalTimeActive
     }];
 }
 
+-(id)getResultByChannel:(NSDictionary<NSString *, id> *)results {
```
jkasten2, line 294: Why have two lists (`OS_SECONDARY_CHANNELS` & `OS_SECONDARY_CHANNELS `) instead of one `OS_CHANNELS` as ` @[OS_PUSH, OS_EMAIL]`?

Then this code could be:
```objective-c
   for (NSString* channelId in OS_CHANNELS) {
        result = results[channelId];
        if (result)
            return result;
    }
   return nil;
```

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
jkasten2: I agree, we can clean this up more later

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
jkasten2: Also this get the first result and the name doesn't reflect that. A better name might be `getFirstAvailableResult` or `getFirstResultByChannelPriority`

jkasten2 (changes requested):

iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
Jeasmine: > Why have two lists (`OS_SECONDARY_CHANNELS` & `OS_SECONDARY_CHANNELS `) instead of one `OS_CHANNELS` as ` @[OS_PUSH, OS_EMAIL]`?
> 
> Then this code could be:
> 
> ```objc
>    for (NSString* channelId in OS_CHANNELS) {
>         result = results[channelId];
>         if (result)
>             return result;
>     }
>    return nil;
> ```

You are right, I was thinking of having different types of priority channels, but this can be simplified. At this level, there is no need for this abstraction

emawby (approved):

jkasten2 (approved):

PR #859, opened 2021-01-15
Background location updates

emawby:
This PR enables background location updates using Apple's significant changes API. When the app is backgrounded and always allow location is enabled we kick off a background task that uses `startMonitoringSignificantLocationChanges` to send location updates. [This API](https://developer.apple.com/documentation/corelocation/getting_the_user_s_location/using_the_significant-change_location_service?language=objc) will wake the app to send location updates even if the user has swiped the app away. 

## Location Distance and Time
It seems "significant location" isn't configurable and could be 500 meters or more and at most every 5 minutes per Apple's note, although it was sending more often than every 5 minutes in my testing.
> Apps can expect a notification as soon as the device moves 500 meters or more from its previous notification. It should not expect notifications more frequently than once every five minutes. If the device is able to retrieve data from the network, the location manager is much more likely to deliver notifications in a timely manner.

https://developer.apple.com/documentation/corelocation/cllocationmanager/1423531-startmonitoringsignificantlocati?language=objc

It seems that it is not possible to do a time based poll of the users location in the background.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/859)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalLocation.m
@@ -158,7 +158,7 @@ static OneSignalLocation* singleInstance = nil;
         if ((int)[NSClassFromString(@"CLLocationManager") performSelector:@selector(authorizationStatus)] == kCLAuthorizationStatusAuthorizedAlways) {
             [OneSignalLocation beginTask];
             [sendLocationTimer invalidate];
-            sendLocationTimer = [NSTimer scheduledTimerWithTimeInterval:adjustedTime target:self selector:@selector(sendLocation) userInfo:nil repeats:NO];
+            sendLocationTimer = [NSTimer scheduledTimerWithTimeInterval:adjustedTime target:self selector:@selector(requestLocation) userInfo:nil repeats:NO];
             [[NSRunLoop mainRunLoop] addTimer:sendLocationTimer forMode:NSRunLoopCommonModes];
         } else {
             sendLocationTimer = NULL;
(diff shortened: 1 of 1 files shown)


jkasten2 (changes requested):
## significantLocationChangeMonitoringAvailable
In[ Apple's docs they recommend calling `CLLocationManager.significantLocationChangeMonitoringAvailable()` ](https://developer.apple.com/documentation/corelocation/getting_the_user_s_location/using_the_significant-change_location_service?language=objc)before `startMonitoringSignificantLocationChanges`. Not sure if this required to prevent any errors or just a recommendation on Apples part. Add a note to this PR with why it isn't needed if so.

## Location Distance and Time
It seems "significant location" isn't configurable and could be 500 meters or more and at most every 5 minutes per Apple's note.
> Apps can expect a notification as soon as the device moves 500 meters or more from its previous notification. It should not expect notifications more frequently than once every five minutes. If the device is able to retrieve data from the network, the location manager is much more likely to deliver notifications in a timely manner.

https://developer.apple.com/documentation/corelocation/cllocationmanager/1423531-startmonitoringsignificantlocati?language=objc

recommend we also note this in the PR.

emawby:
> ## significantLocationChangeMonitoringAvailable
> In[ Apple's docs they recommend calling `CLLocationManager.significantLocationChangeMonitoringAvailable()` ](https://developer.apple.com/documentation/corelocation/getting_the_user_s_location/using_the_significant-change_location_service?language=objc)before `startMonitoringSignificantLocationChanges`. Not sure if this required to prevent any errors or just a recommendation on Apples part. Add a note to this PR with why it isn't needed if so.
> 
> ## Location Distance and Time
> It seems "significant location" isn't configurable and could be 500 meters or more and at most every 5 minutes per Apple's note.
> 
> > Apps can expect a notification as soon as the device moves 500 meters or more from its previous notification. It should not expect notifications more frequently than once every five minutes. If the device is able to retrieve data from the network, the location manager is much more likely to deliver notifications in a timely manner.
> 
> https://developer.apple.com/documentation/corelocation/cllocationmanager/1423531-startmonitoringsignificantlocati?language=objc
> 
> recommend we also note this in the PR.

I added the location distance and time note to the description. As for checking for authorization I don't think it is needed right now because we are only calling it when always in use authorization is enabled which includes significantLocationChangeMonitoring. That said it is probably good practice to add it anyways in case the requirements change in the future.

jkasten2 (approved):

PR #862, opened 2021-01-22
Fix failing tests

Jeasmine:
Following tests were failing on master

* Fix testPushNotificationToken, observer will be always null if permission consent is required
* Fix testCallingMethodsWorks_beforeInit, OneSignalLocationOverrider data needs to be reset
* Fix testSubscriptionChangeObserverBasic, testSubscriptionChangeObserverBasic, _registerUserFinished need to be reset

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/862)


emawby (approved):
lgtm!

PR #873, opened 2021-02-10
Using generic/platform to include all architectures

emawby:
The previous xcodebuild archive commands which listed out architectures was not working for M1 macs as of the latest Xcode version. Using generic/platform is the recommended way of building xcframeworks for all architectures.

Tested on an M1 Dev Kit and Intel mac. Tested in Xcode 12.3 and 11.0

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/873)


jkasten2 (approved):

jkasten2 (approved):
Changes look good, would also but good to add to the PR which versions of Xcode were tested.
Also link to any docs Apple has on this generic build would be good reference if we need it later too.

PR #877, opened 2021-02-12
Forwarding notification from willPresentInForeground for non OneSignal notifications

emawby:
This is the 3.X.X version of the fix

This issue manifested primarily when customers tried to integrate with Firebase and OneSignal at the same time. Notifications sent from Firebase would not be forwarded by OneSignal correctly but notifications sent from OneSignal were forwarded.
This only happened when the application did not set the `UNUserNotificationCenter` delegate.

This has been tested in iOS native.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/877)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/UNUserNotificationCenter+OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/UNUserNotificationCenter+OneSignal.m
@@ -157,6 +157,24 @@ static UNNotificationSettings* cachedUNNotificationSettings;
     [self setOneSignalUNDelegate:delegate];
 }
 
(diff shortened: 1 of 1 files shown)


Jeasmine (approved):
LGTM seems like the same change as https://github.com/OneSignal/OneSignal-iOS-SDK/pull/876. Shouldn't we merge this one and then cherry-pick to the backported branch?

jkasten2 (approved):

emawby:
> LGTM seems like the same change as #876. Shouldn't we merge this one and then cherry-pick to the backported branch?

It is slightly different in the major release since the major release is split into 2 methods

PR #891, opened 2021-03-02
WIP - Register with OneSignal without delay; If we have a token or won't be getting one w/o permission

jkasten2:
* We want to register as soon as we get a push token without delaying.
* The network optimization noted in the deleted comment is a very small
and has negative impacts such as delaying when IAMs are shown

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/891)


emawby:
@jkasten2 Looks like this change is already in main. Should we close this PR?

Issue #848, opened 2021-01-05
[Crash] Several crashes in OneSignalUserDefaults.m and once.h


ToMark1881:
**Description:**
I have a lot of crashes in firebase Crashlytics from OneSignal SDK.
Updating to OneSignal SDK 3.0.0 didn't help. (2.15.3 -> 2.15.4 -> 3.0.0)
The first cases were December 14th (**Google down**?)

**Environment**

1. OneSignal SDK 3.0.0
1. CocoaPods
1. Xcode 12

**Steps to Reproduce Issue:**

1. Open App
2. Crash

Crashes from Firebase Crashlytics

**Stack traces:**
OneSignalUserDefaults.m and once.h

**Stack trace 1:**
```
Crashed: com.apple.main-thread
0  libdispatch.dylib              0x188678550 _dispatch_once_wait.cold.1 + 28
1  libdispatch.dylib              0x188645708 _dispatch_gate_broadcast_slow + 194
*2  SuperDealProd                  0x104ddb8c8 +[OSMessagingController sharedInstance] + 84 (once.h:84) <-- here
3  SuperDealProd                  0x104dd5a20 +[OneSignal receivedInAppMessageJson:] + 1853 (OneSignal.m:1853)
4  SuperDealProd                  0x104dd54bc __33+[OneSignal registerUserInternal]_block_invoke + 1820 (OneSignal.m:1820)
5  libdispatch.dylib              0x18864324c _dispatch_call_block_and_release + 32
6  libdispatch.dylib              0x188644db0 _dispatch_client_callout + 20
7  libdispatch.dylib              0x1886527ac _dispatch_main_queue_callback_4CF + 836
8  CoreFoundation                 0x1889cc11c **CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE**
9  CoreFoundation                 0x1889c6120 __CFRunLoopRun + 2508
10 CoreFoundation                 0x1889c521c CFRunLoopRunSpecific + 600
11 SuperDealProd                  0x104be0eac GAIUncaughtExceptionHandler + 69 (GAIUncaughtExceptionHandler.m:69)
12 CoreFoundation                 0x188a50b80 __handleUncaughtException + 632
13 libobjc.A.dylib                0x19cfa6f58 _objc_terminate() + 132
14 libc++abi.dylib                0x19d0ac0e0 std::__terminate(void (*)()) + 20
[...]

PR #894, opened 2021-03-04
Fix IAM redisplay cache userdefaults crashes

emawby:
Fix for issue #848

`OS_IAM_REDISPLAY_DICTIONARY` contains `OSInAppMessage` objects so we can't save it as a dictionary to nsuserdefaults or it will crash. We were do this when deleting IAMs that were over 6 months old.

We are now saving them properly as codeable data, but previously saved dictionary data will still cause a crash when trying to retrieve the cache as codeabledata so a migration was created to convert the dictionary data to codeable data.

To properly test this and catch crashes like this in the future I have removed the NSUserDefaultsOverrider. Since we are now running tests in a containing app, we can use the standard UserDefaults implementation that catches types issues like this one.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/894)


jkasten2 (changes requested):
> Reviewed 1 of 1 files at r1, 12 of 12 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/894)* status: all files reviewed, 1 unresolved discussion (waiting on @emawby and @Jeasmine)

___
*[iOS_SDK/OneSignalSDK/Source/OSMigrationController.m, line 93 at r3](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/894#-MUywcfYEBaqfkW3lZ9s:-MUywcfYEBaqfkW3lZ9t:b7ocrtc) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/0cc604b2884ede6ad663775f5fe6bd2b055760ef/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m#L93)):*
> ```objc
>     
>     @try {
>         __unused NSMutableDictionary *_ =[[NSMutableDictionary alloc] initWithDictionary:[OneSignalUserDefaults.initStandard
> ```

Why use `__unused`? Do we get a warning if we ignore the input and put comment instead? If that is the case can we still give this a variable name instead of using `_` as the name?

Jeasmine (approved):
Nice testing and fix!

emawby (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/894)* status: 13 of 15 files reviewed, 1 unresolved discussion (waiting on @jkasten2)

___
*[iOS_SDK/OneSignalSDK/Source/OSMigrationController.m, line 93 at r3](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/894#-MUywcfYEBaqfkW3lZ9s:-MUz6qWGA814c6H6_yN1:b-6aksgs) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/0cc604b2884ede6ad663775f5fe6bd2b055760ef/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m#L93)):*

Why use `__unused`? Do we get a warning if we ignore the input and put comment instead? If that is the case can we still give this a variable name instead of using `_` as the name?
</blockquote></details>

unused gets rid of the Xcdoe warning. I named it redisplayDict

jkasten2 (approved):
> Reviewed 2 of 2 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/894)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @emawby)

PR #901, opened 2021-03-15
Remove alert for launch URLs

emawby:
This alert was causing problems for customers where it could be dismissed when the app was opened. It does not seem to be a desirable function so we will remove it. This change in behavior will need to be communicated to customers, but the change will not cause any build issues.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/901)


jkasten2 (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/901)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine)

PR #903, opened 2021-03-16
External user id not being sent fix

emawby:
This PR fixes two scenarios where we were not sending the external user id when we should be.

1. When we have already set the external push user id, and are now trying to set the external email user id to the same value. For this we had a comment "If we are making a request to email user, we need validate both external user ids", but I am not sure why that is the case.
1. When you have a cached external user id, but the player has been deleted from the dashboard and you are trying to set the external user id to the cached value

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/903)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
@@ -148,6 +148,13 @@ THE SOFTWARE.
         //update push player id
         if (results.count > 0 && results[OS_PUSH][@"id"]) {
             self.currentSubscriptionState.userId = results[OS_PUSH][@"id"];
+            
+            // Player record was deleted so we should reset external user id
+            let cachedPlayerId = [OneSignalUserDefaults.initStandard getSavedStringForKey:OSUD_PLAYER_ID_TO defaultValue:nil];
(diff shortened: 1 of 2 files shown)


iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
```diff
@@ -148,6 +148,13 @@ - (void)registerUserWithState:(OSUserState *)registrationState withSuccess:(OSMu
         //update push player id
         if (results.count > 0 && results[OS_PUSH][@"id"]) {
             self.currentSubscriptionState.userId = results[OS_PUSH][@"id"];
+            
+            // Player record was deleted so we should reset external user id
+            let cachedPlayerId = [OneSignalUserDefaults.initStandard getSavedStringForKey:OSUD_PLAYER_ID_TO defaultValue:nil];
```
Jeasmine, line 153: When the player is deleted from the server on Android, we clean userId and all cached data for that player. Finally, re-trigger the player's creation

jkasten2 (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/903)* status: all files reviewed, 1 unresolved discussion (waiting on @Jeasmine)

PR #907, opened 2021-03-18
Fix Dialog Issue

iAugux:
There are two issues for the dialog.

1. The `OneSignalDialogControllerOverrider` was included to the SDK target mistakenly once again. Same as PR https://github.com/OneSignal/OneSignal-iOS-SDK/pull/420
1. Currently we are presenting the dialog(`UIViewController`) from `keyWindow.rootViewController`, but if the rootViewController already presented another vc, it's not allowed to present the alert controller anymore. So we need to present the alert from the visible(top most) vc.

> _[Presentation] Attempt to present <UIAlertController: 0x15b818400> on <Test.RootVC: 0x157e078e0> (from <Test.RootVC: 0x157e078e0>) which is already presenting <_TtC4TestP33_8314E842ED9EA302AFA529AE16E3E3E21VC: 0x158e04be0>__

_Tested on iOS 14.4 Xcode 12.4_

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/907)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalDialogController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalDialogController.m
@@ -90,8 +90,8 @@
 }
 
 - (void)displayDialog:(OSDialogRequest * _Nonnull)request {
-    let rootViewController = [[[UIApplication sharedApplication] keyWindow] rootViewController];
-    
+    let visibleViewController = [OneSignalDialogController visibleViewController];
+
     let controller = [UIAlertController alertControllerWithTitle:request.title message:request.message preferredStyle:UIAlertControllerStyleAlert];
     
     [controller addAction:[UIAlertAction actionWithTitle:request.cancelTitle style:UIAlertActionStyleCancel handler:^(UIAlertAction * _Nonnull action) {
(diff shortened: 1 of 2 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignalDialogController.m
```diff
@@ -139,4 +139,23 @@ - (void)clearQueue {
     self.queue = [NSMutableArray new];
 }
 
++ (UIViewController *)visibleViewController {
+    let rootViewController = [[[UIApplication sharedApplication] keyWindow] rootViewController];
+    return [OneSignalDialogController getVisibleViewControllerFrom:rootViewController];
+}
+
++ (UIViewController *) getVisibleViewControllerFrom:(UIViewController *) vc {
```
rogerluan, line 147:
```suggestion
+ (UIViewController *)getVisibleViewControllerFrom:(UIViewController *) vc {
```

rogerluan (approved):
I can verify that this has fixed both of those issues 👍 Nice catch.

@ OneSignal team, maybe there's a more permanent way to fix the `OneSignalDialogControllerOverrider.h` issue that has already regressed more than once? Not sure why this was implemented the way it is currently, but seems to be easy enough to cause problems again in the future. 🙏

jkasten2 (approved):
Thanks for the PR! Changes look good to me.

I don't think `OneSignalDialogControllerOverrider.h` could have caused an issue since the `.m` isn't included. Indeed it should be removed from the target though.
Unsure how this got into the target the first time, or this time. A lint rule in the future is one idea I have to prevent it though.

PR #917, opened 2021-04-23
Removing duplicate checks for set_external_user_id

emawby:
We have seen some issues where set_external_user_id is not being set on channel records with an error that it has already been set. This PR simplifies the set_external_user_id code to always set the external user id, even if it matches the cached value for that channel. This will make setExternalUserId more reliable and simplifies the code path.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/917)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
@@ -200,16 +200,6 @@ THE SOFTWARE.
                                                                                       withAppId:appId];
     }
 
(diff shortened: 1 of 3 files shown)


jkasten2 (approved):
:lgtm:

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/917#-:-MZ-MIDIBYZJz1Ls8Jr2:bnfp4nl)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @Jeasmine)

PR #924, opened 2021-05-12
Fixed Pre-existing Notification Delegate

tanaynigam:
This happens when another library's delegate is loaded before OneSignal's delegate code.
This commit solves it by reassigning the pre loaded delegate into swizzle in OneSignal's  methods.
This issue was discovered through a Unity issue where Unity Mobile Notification Services would load their delegate before OneSignal's delegate.
https://github.com/OneSignal/OneSignal-Unity-SDK/issues/245

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/924)

PR #926, opened 2021-05-13
Fix pre-existing notification delegate

jkasten2:
## Summary
This fixes an issue where OneSignal's notification open event would not fire if a UNUserNotificationCenterDelegate was assigned before the OneSignal was loaded into memory.

## Details
This commit solves it by reassigning the pre-loaded delegate into swizzle in OneSignal's methods.
This issue was discovered through a Unity issue where Unity Mobile Notification Services would load their delegate before OneSignal's delegate.
* OneSignal/OneSignal-Unity-SDK#245

## Tests
* Added a new test to cover this specific pre-existing case.

## Related PR
This applies the same fix from PR #924 but includes some refactoring to make it testable.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/926)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -2945,10 +2945,7 @@ static NSString *_lastnonActiveMessageId;
     if (!NSClassFromString(@"UNUserNotificationCenter"))
         return;
 
-    [OneSignalUNUserNotificationCenter swizzleSelectors];
-
-    // Set our own delegate if one hasn't been set already from something else.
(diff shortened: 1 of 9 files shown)


emawby (approved):
> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 3 of 3 files at r3, 2 of 2 files at r4, 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/926)* status: 5 of 9 files reviewed, all discussions resolved (waiting on @emawby and @tanaynigam)

PR #930, opened 2021-05-18
Fix IAM Tag Substitution and IAMs not updating after being edited.

emawby:
This PR fixes two issues with IAMs

1. We were caching the results of loading HTML content for IAMs. This meant that if an IAM which was scheduled to display every time a trigger fired gets updated on the dashboard, we may not display the updates to the user. I have turned off local caching for the load HTML call to fix this.
2. If tags weren't cached on the device they were not being used in tag substitution because the result of getPlayerTags was not being used to populate the playerTags object. This is fixed and the playerTags class has been slightly refactored to make it easier more clear to work with.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/930)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m
@@ -57,7 +57,7 @@
 - (NSString *)getTagsString {
     NSError *error;
     OSPlayerTags *tags = [OneSignal getPlayerTags];
-    if (!tags.allTags) {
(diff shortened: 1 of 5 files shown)


jkasten2 (approved):
> Reviewed 1 of 1 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/930)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine)

PR #939, opened 2021-06-03
Fix clearing badge count on reinstall

emawby:
Badges were not being cleared on app reinstall until the user was prompted for push permissions, because we were checking for badge permissions before clearing badges. We don't need permission to set the badge count to 0 and in fact if they don't have permission then we definitely should be setting it to 0. 

This change affected a unit test that was trying to reproduce an iOS 10.2.1 issue where permissions would be set to denied when backgrounding on the push prompt. I am not sure if this change fixed the iOS 10.2.1 issue, but I modified the test so that it passes.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/939)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -2181,8 +2181,7 @@ static NSString *_lastnonActiveMessageId;
     else
         disableBadgeClearing = NO;
     
-    if (disableBadgeClearing ||
-        ([self.osNotificationSettings getNotificationPermissionState].notificationTypes & NOTIFICATION_TYPE_BADGE) == 0)
+    if (disableBadgeClearing)
         return false;
     
     bool wasBadgeSet = [UIApplication sharedApplication].applicationIconBadgeNumber > 0;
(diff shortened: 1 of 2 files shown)


jkasten2 (commented):
> Badges were not being cleared on app reinstall until the user was prompted for push permissions, because we were checking for badge permissions before clearing badges.

I was pretty sure badges were cleared when you uninstall the app then reinstall it on iOS. Also I think if you tried to set `applicationIconBadgeNumber` without permission it shows a warning in the Xcode log, maybe only if it is a non-zero value though.

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/939#-:-MbIAiGK9Ey9MgbRa2Nz:b-96ve24)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @Jeasmine)

jkasten2 (approved):
Regardless if this fixes the actual issue in question I approve this PR as this is good clean up if the extra check is jsut unnecessary.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/939#-:-MbIcNHD3RN_Y0z_kX9P:b-i1u6q1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine)

PR #949, opened 2021-06-15
Fixing IAM displaying twice on cold start

emawby:
If we display IAMs prior to receiving the response to on session, we could be displaying cached IAMs that have been paused, or edited. Additionally this is causing an issue where we display IAMs twice because of the `self.messages = newMessages` line in `updateInAppMessagesFromOnSession`. We overwrite the `seenInSession` variable of the old object with the new one. By waiting to display IAMs until onSession has returned we fix this issue. To avoid a case where cached IAMs aren't shown when onSession fails I am now calling `updateInAppMessagesFromCache` in the failure block of `registerUserInternal`.

When trying to display a message before onSession has completed we will add it to the queue. We will then evaluate the message queue again once onSession finishes. If onSession fails we will still show cached IAMs because `updateInAppMessagesFromCache` calls `evaluateMessages`

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/949)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
@@ -160,7 +160,6 @@ static BOOL _isInAppMessagingPaused = false;
 
 - (void)updateInAppMessagesFromCache {
     self.messages = [OneSignalUserDefaults.initStandard getSavedCodeableDataForKey:OS_IAM_MESSAGES_ARRAY defaultValue:[NSArray new]];
-
     [self evaluateMessages];
 }
 
@@ -241,8 +240,9 @@ static BOOL _isInAppMessagingPaused = false;
             [self.messageDisplayQueue addObject:message];
         
         // Return early if an IAM is already showing
-        if (self.isInAppMessageShowing)
+        if (self.isInAppMessageShowing || !OneSignal.isRegisterUserFinished) {
             return;
+        }
         // Return early if the app is not active
         if (![UIApplication applicationIsActive]) {
             [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:@"Pause IAMs display due to app inactivity"];
@@ -286,7 +286,6 @@ static BOOL _isInAppMessagingPaused = false;
         [OneSignal onesignal_Log:ONE_S_LL_VERBOSE message:@"In app messages will not show while paused"];
         return;
     }
-    
     self.isInAppMessageShowing = true;
     [self showMessage:message];
 }
--- a/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
@@ -240,7 +240,7 @@ static BOOL _isInAppMessagingPaused = false;
             [self.messageDisplayQueue addObject:message];
         
         // Return early if an IAM is already showing
-        if (self.isInAppMessageShowing || !OneSignal.isRegisterUserFinished) {
+        if (self.isInAppMessageShowing) {
             return;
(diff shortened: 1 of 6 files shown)


jkasten2 (approved):
Changes themselves look good to me. Would be good to have test for this however.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/949#-:-McIMji96SOuX1WQGYw4:bcrdkeu)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @Jeasmine)

emawby (commented):
Added a unit test validating that we don't try to display an IAM before init is called. Some customers have reported crashes in IAMs and it could be because we were previously trying to send impressions for cached IAMs before init is called (no appid or playerid) if they weren't always calling initWithLaunchOptions. The stack trace I got from this unit test without my changes looks similar to the customer stack traces I have seen 

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/949#-:-McQ0maYAgey8LMd1KrS:b-9s0r03)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @Jeasmine)

iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
```diff
@@ -241,8 +240,9 @@ - (void)presentInAppMessage:(OSInAppMessage *)message {
             [self.messageDisplayQueue addObject:message];
         
         // Return early if an IAM is already showing
-        if (self.isInAppMessageShowing)
+        if (self.isInAppMessageShowing || !OneSignal.isRegisterUserFinished) {
```
Jeasmine, line 243: Maybe we can add a comment regarding this conditional addition. I imagine you are not putting it on  `[self evaluateMessages];` because even if the user is not registered, some triggers might have happened that make an IAM appear, but we need to wait for the user registration. Right? 

Can it happen that after the session call, the IAM added to the queue is deleted from the Server? Might messageDisplayQueue end in a weird consistency situation? because messageDisplayQueue is not used until evaluateMessageDisplayQueue is called

emawby: Ya thats a good point I think I should check earlier so that the message doesn't enter the queue. I don't want the message in the queue and not in the messages array

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -1928,6 +1928,8 @@ + (void)registerUserInternal {
                     callbackSet.failureBlock(error);
             }
         }
+        
+        [OSMessagingController.sharedInstance updateInAppMessagesFromCache];
```
Jeasmine, line 1932: I'm confused with one thing since you are checking for user registration to show IAM. At this point, OneSignal.isRegisterUserFinished; will return true? because it will immediateOnSessionRetry, then both _registerUserFinished || isOnSessionSuccessfulForCurrentState will be false right? Am I missing something? Are the cached IAM being displayed? Thanks in advance for clarifying this

emawby: Ya you are right this isn't working we would need a registerUserFailed boolean or settings registerUserFinished to true if it fails (it gets reset when it is called again).

emawby: I added a registerUserSuccessful boolean to be used in place of registerUserFinished in some places. Added comments to make it clear that registerUserFinished is also set to true when the call fails

Jeasmine (approved):

PR #961, opened 2021-07-21
Fix/calling nil failure block and OneSignal.h set to public

emawby:
This PR fixes 2 issues.
1. We were not checking that the failureBlock of sendTags was non-nil before trying to call it.
2. The OneSignal.h header file was set to project instead of Public which prevents building.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/961)


jkasten2 (approved):
:lgtm:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/961#-:-MfACqraDFBUvi9AxhVM:bnfp4nl)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @Jeasmine and @nan-li)

PR #984, opened 2021-08-24
Enable no margin IAMs

emawby:
Current IAMs always include a non-zero margin meaning that IAMs cannot fill the entire screen. To enable full screen IAMs we will now optionally disable the margin based on new data being sent as part of the loadHTMLContent response for an IAM. This data will come from a new "styles" dictionary in the following keys:
top_margin: 0
bottom_margin: 0
left_margin: 0 
right_margin: 0

If "styles" is absent, or it does not contain a margin key that margin will keep the default margin value. 

For now the SDK will only respect a value of 0. Any other value will result in the default margins. In the future we can use values from these keys to set customized margin values.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/984)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m
@@ -244,14 +244,7 @@
             [[OneSignal sessionManager] onInAppMessageReceived:self.message.messageId];
 
         let baseUrl = [NSURL URLWithString:OS_IAM_WEBVIEW_BASE_URL];
-        NSMutableDictionary *fakeData = [[NSMutableDictionary alloc] initWithDictionary:data];
-        fakeData[@"styles"] = @{
-            @"top_margin" : @"0",
-            @"bottom_margin" : @"0",
-            @"left_margin" : @"0",
-            @"right_margin" : @"0",
-        };
-        [self parseContentData:fakeData];
+        [self parseContentData:data];
         if (self.waitForTags) {
             return;
         }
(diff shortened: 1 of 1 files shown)


emawby (commented):
Do you think we should change the JSON or keep the JSON the same and change the client? With this approach I was hoping we wouldn't need to change the JSON in the future.

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/984#-:-MiMu0YU0E3XBsc4K7ep:b-thfnzf)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @Jeasmine, @jkasten2, and @nan-li)

___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m, line 248 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/984#-Mi83gE68kt4AShM91SR:-MiMtNiZ9fR1NbObF-zR:bea0jo6) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/89ec570d50a61aaa7bb446de44d6402a94dcd270/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m#L248)):*

The wording `fake` is something you would use in tests but this is part of the source. I would suggest something like `default`.
</blockquote></details>

This was used during my testing but shouldn't be in my final commit. I am assuming you are looking at the changes commit by commit so this should be removed already.
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m, line 269 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/984#-Mi7qdYT2wNuTVRjF84O:-MiMv9vVBl3h-wpMjpTP:b-896fix) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/89ec570d50a61aaa7bb446de44d6402a94dcd270/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m#L269)):*

Would be cleaner if we extracted `data[@"styles"]` into a variable we so don't have to repeat it here.
</blockquote></details>

Done.

emawby (commented):
updated to use height and width margin boolean

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/984#-:-MiNtMIxD6bhDUd6GJju:b-h8w5uq)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @Jeasmine, @jkasten2, and @nan-li)

PR #992, opened 2021-09-03
Test to ensure we register after apns response

jkasten2:
This new test is similar to the existing `testNotificationPermissionsAcceptedBeforeAddingOneSiganl_waitsForAPNSTokenBeforePlayerCreate` test above it however we are testing to ensure we register even when notification permissions are not enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/992)


emawby (approved):
looks good! We probably don't want to wait for APNs to register the user but that is a separate issue

PR #998, opened 2021-09-20
Fix badge incrementing when cancelling display in foreground

emawby:
Fixes https://github.com/OneSignal/react-native-onesignal/issues/1256

When choosing to not display a notification with our foreground handler iOS does not increment the badge count or send the notification to the notification center, but we cache the increased value in OneSignal's user defaults. This cause a subsequent notification to include the cancelled notifications badge increment value.

In order to prevent this we need to reset the cached badge count if the notification is not displayed. In order to fix this for non-incremental badge changes we need to also store the previous badge count value to NSUserDefaults.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/998)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSNotification.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSNotification.m
@@ -27,11 +27,15 @@
 
 #import <Foundation/Foundation.h>
 
+#import <UIKit/UIKit.h>
+
 #import "OSNotification+Internal.h"
 
 #import "OneSignal.h"
 
 #import "OneSignalCommonDefines.h"
+
+#import "OneSignalUserDefaults.h"
 
 @implementation OSNotification
 
@@ -298,6 +302,14 @@
 
  - (OSNotificationDisplayResponse)getCompletionBlock {
      OSNotificationDisplayResponse block = ^(OSNotification *notification){
+         /*
+          If notification is null here then display was cancelled and we need to
+          reset the badge count to the value prior to receipt of this notif
+          */
+         if (!notification) {
+             NSInteger previousBadgeCount = [UIApplication sharedApplication].applicationIconBadgeNumber;
+             [OneSignalUserDefaults.initShared saveIntegerForKey:ONESIGNAL_BADGE_KEY withValue:previousBadgeCount];
+         }
          [self complete:notification];
      };
      return block;


jkasten2 (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/998)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @emawby, @nan-li, and @rgomezp)

___
*[iOS_SDK/OneSignalSDK/Source/OSNotification.m, line 312 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/998#-Mk5uemMA_LlgkA4qFUY:-Mk5uemMA_LlgkA4qFUZ:bnmyv28) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/c01359a03796732a47eb87962144b58d49c619cd/iOS_SDK/OneSignalSDK/Source/OSNotification.m#L312)):*
> ```objc
>           */
>          if (!notification) {
>              NSInteger previousBadgeCount = [OneSignalUserDefaults.initShared getSavedIntegerForKey:ONESIGNAL_PREVIOUS_BADGE_KEY defaultValue:0];
> ```

Instead of using us having to track a previous value we could instead read the iOS value `applicationIconBadgeNumber` again here. This means one less state we have to keep track of and means we also don't need a migration to correctly set the previous badge value.

emawby (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/998)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jkasten2, @nan-li, and @rgomezp)

___
*[iOS_SDK/OneSignalSDK/Source/OSNotification.m, line 312 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/998#-Mk5uemMA_LlgkA4qFUY:-MkEMS_j37x3nHqrRGi_:b-b5kji1) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/c01359a03796732a47eb87962144b58d49c619cd/iOS_SDK/OneSignalSDK/Source/OSNotification.m#L312)):*

Instead of using us having to track a previous value we could instead read the iOS value `applicationIconBadgeNumber` again here. This means one less state we have to keep track of and means we also don't need a migration to correctly set the previous badge value.
</blockquote></details>

Ya that will work and I will make the change. The problem with this is that we then use a `UIApplication sharedApplication` method in `OSNotification` which is used in code required by our extension handlers. This means we won't be able to get rid of all sharedApplication references in an NSE only framework. 

However when that transition does happen we could move this logic to an extension that only exists in the full framework.

jkasten2 (approved):
> Reviewed 4 of 4 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/998)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nan-li and @rgomezp)

___
*[iOS_SDK/OneSignalSDK/Source/OSNotification.m, line 312 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/998#-Mk5uemMA_LlgkA4qFUY:-MkJHnuQ1Bmm_lUt4tOf:bat239) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/c01359a03796732a47eb87962144b58d49c619cd/iOS_SDK/OneSignalSDK/Source/OSNotification.m#L312)):*

Ya that will work and I will make the change. The problem with this is that we then use a `UIApplication sharedApplication` method in `OSNotification` which is used in code required by our extension handlers. This means we won't be able to get rid of all sharedApplication references in an NSE only framework. 

However when that transition does happen we could move this logic to an extension that only exists in the full framework.
</blockquote></details>

We can move this logic outside of `OSNotification` in the future to account for the NSE decoupling. This doesn't seems something the OSNotification class should be responsible for anyway.

PR #1000, opened 2021-09-23
2.x.x unity IAM orientation fix

emawby:
Cherry pick of #999 

Unity overrides orientation behavior and enables all orientations in supportedInterfaceOrientations, regardless of the values set in the info.plist. It then uses its own internal logic for restricting the Application's views to the selected orientations.

Our InAppMessageViewController inherits the behavior of all orientations being allowed, so we need to manually set the supported orientations based on the values in the plist.

To do this we can read the values from the supported orientations array in the mainBundle and then return the appropriate UIInterfaceOrientationMask.

If new orientations are released in future iOS versions we will need to update this method

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1000)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m
@@ -656,6 +656,40 @@
 }
 
 /*
+ Unity overrides orientation behavior and enables all orientations in supportedInterfaceOrientations, regardless of
+ the values set in the info.plist. It then uses its own internal logic for restricting the Application's views to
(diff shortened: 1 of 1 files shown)


jkasten2 (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1000#-:-MkhVb5lAvoE3mgPCREp:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamschlesinger and @nan-li)

adamschlesinger (approved):

PR #1009, opened 2021-10-01
WiP: call `failureBlock` during `setExternalUserId` if one or more channels return success of false

nan-li:
- we were always calling the `successBlock` after making the request to set external user ID
- which means we call the `successBlock` even when the backend doesn't set it and return the error of invalid auth hath
- now we check if the result has any channels that returned "success" of false and call the `failureBlock`
- (to remove later) I made some comments in the code about what to do for certain things

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1009)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m
@@ -202,16 +202,39 @@ THE SOFTWARE.
     }
 
     [OneSignalClient.sharedClient executeSimultaneousRequests:requests withCompletion:^(NSDictionary<NSString *,NSDictionary *> *results) {
-        if (results[OS_PUSH] && results[OS_PUSH][OS_SUCCESS] && [results[OS_PUSH][OS_SUCCESS] boolValue]) {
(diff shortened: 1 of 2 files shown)


emawby (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1009)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @emawby and @nan-li)

___
*[iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m, line 205 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1009#-MlBxRDv4zVQt_bkRuM0:-MlBxRDv4zVQt_bkRuM1:b-vl3rdp) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/d6c27474c7b0bae1230c3e29b40016401ff74749/iOS_SDK/OneSignalSDK/Source/OSStateSynchronizer.m#L205)):*
> ```objc
> 
>     [OneSignalClient.sharedClient executeSimultaneousRequests:requests withCompletion:^(NSDictionary<NSString *,NSDictionary *> *results) {
>         var requestDidFail = false;
> ```

I would like us to start typing variables in objective-c and avoiding using var!

PR #1018, opened 2021-11-04
Fullbleed In App Messages

emawby:
# Description
## One Line Summary
Added support for fullbleed in app messages to the iOS SDK. 

## Details
Fullscreen and Carousel In App Messages that do not have margins will now bleed under the status bar to be completely full screen.
The dashboard has added the ability for fullscreen and Carousel IAMs to be created without margins. When doing so it sets `remove_height_margin` and `remove_width_margin` to true in `styles` (this was already being read by the SDK). Currently we don't support only width or height removal so I am consolidating these flags into an isFullscreen boolean that controls the fullbleed behavior. 

In order to handle notched devices we need to tell the Javascript about the device insets so that IAM content doesn't overlap with notches and the status bar. To do this we are modifying the IAMs HTML and calling setSafeAreaInsets with a javascript object of the form
```
{
top: ,
bottom: ,
right: ,
left: ,
}
```
#### Orientation Change

We also need to call the setSafeAreaInsets() method again if the device is rotated since the side of the notches have changed. To do so we are now responding to the resize javascript event (previously unused).

#### Misc Changes
Since we need to access UIApplications's keyWindow safeAreaInsets we need to be on the main thread once we receive the message's HTML. I have changed where we go to the main thread from the message view to the response of the loadHTML call.

Additionally the webView needs to know if it is fullScreen right away so that carousels are provided with the correct width on initial load and orientation change. To do this I added a setIsFullscreen method to `OSInAppMessageView` that enables/disables margins.

### Motivation
Customer request for full screen IAMs.

### Background Context
When we initially discussed removing margins from IAMs we discussed the future ability to remove just width or height margins so we split them into their own booleans. We eventually decided against that approach and to just focus on fullbleed IAMs.

### Scope
This change involves coordinating new behavior between the SDK and IAM Javascript. It makes no new API calls and provides no new public APIs in the SDK. This change should not affect non fullscreen/carousel IAMs. 

# Testing
## Unit testing
✅  All tests pass after the below modifications
We don't infrastructure for testing properties of the OSInAppMessageViewController currently or good ways to test a messages HTML content. 

## Manual testing
✅  Tested on various iOS simulators for carousels and fullscreen IAMs. Tested both landscape and portrait

### Example of the Result

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [X] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing
   - [X] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [X] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [X] Code is as readable as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018)

### DIFF
--- a/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
+++ b/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
@@ -85,7 +85,7 @@ OneSignalNotificationCenterDelegate *_notificationDelegate;
     [OneSignal addSubscriptionObserver:self];
     [OneSignal addEmailSubscriptionObserver:self];
     [OneSignal setInAppMessageLifecycleHandler:self];
-    [OneSignal pauseInAppMessages:false];
+    [OneSignal pauseInAppMessages:true];
 
     [OneSignal setNotificationWillShowInForegroundHandler:notificationReceiverBlock];
     [OneSignal setNotificationOpenedHandler:openNotificationHandler];
@@ -95,7 +95,7 @@ OneSignalNotificationCenterDelegate *_notificationDelegate;
     return YES;
 }
 
-#define ONESIGNAL_APP_ID_DEFAULT @"9ddf83fd-6cf8-403d-98a1-bd5285b23d0e"
+#define ONESIGNAL_APP_ID_DEFAULT @"0ba9731b-33bd-43f4-8b59-61172e27447d"
 #define ONESIGNAL_APP_ID_KEY_FOR_TESTING @"ONESIGNAL_APP_ID_KEY_FOR_TESTING"
 
 + (NSString*)getOneSignalAppId {
--- a/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
+++ b/iOS_SDK/OneSignalDevApp/OneSignalDevApp/AppDelegate.m
@@ -85,7 +85,7 @@ OneSignalNotificationCenterDelegate *_notificationDelegate;
     [OneSignal addSubscriptionObserver:self];
     [OneSignal addEmailSubscriptionObserver:self];
     [OneSignal setInAppMessageLifecycleHandler:self];
-    [OneSignal pauseInAppMessages:true];
+    [OneSignal pauseInAppMessages:false];
 
     [OneSignal setNotificationWillShowInForegroundHandler:notificationReceiverBlock];
     [OneSignal setNotificationOpenedHandler:openNotificationHandler];
@@ -95,7 +95,7 @@ OneSignalNotificationCenterDelegate *_notificationDelegate;
     return YES;
 }
 
-#define ONESIGNAL_APP_ID_DEFAULT @"0ba9731b-33bd-43f4-8b59-61172e27447d"
+#define ONESIGNAL_APP_ID_DEFAULT @"9ddf83fd-6cf8-403d-98a1-bd5285b23d0e"
 #define ONESIGNAL_APP_ID_KEY_FOR_TESTING @"ONESIGNAL_APP_ID_KEY_FOR_TESTING"
 
 + (NSString*)getOneSignalAppId {
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.h
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.h
@@ -44,6 +44,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype _Nonnull)initWithMessage:(OSInAppMessageInternal *)inAppMessage withScriptMessageHandler:(id<WKScriptMessageHandler>)messageHandler;
 - (void)resetWebViewToMaxBoundsAndResizeHeight:(void (^) (NSNumber *newHeight)) completion;
+- (void)updateSafeAreaInsets;
 - (void)setupWebViewConstraints;
 - (void)loadReplacementURL:(NSURL *)url;
 - (void)loadedHtmlContent:(NSString *)html withBaseURL:(NSURL *)url;
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.h
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.h
@@ -49,7 +49,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)loadReplacementURL:(NSURL *)url;
 - (void)loadedHtmlContent:(NSString *)html withBaseURL:(NSURL *)url;
 - (void)removeScriptMessageHandler;
-
+- (void)setIsFullscreen:(BOOL)isFullscreen;
 @end
 
 NS_ASSUME_NONNULL_END
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m
@@ -90,12 +90,12 @@
 - (void)loadedHtmlContent:(NSString *)html withBaseURL:(NSURL *)url {
     // UI Update must be done on the main thread
     NSLog(@"11111 [self.webView loadHTMLString:html baseURL:url];");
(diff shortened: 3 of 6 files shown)


jkasten2 (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018)* status: all files reviewed, 6 unresolved discussions (waiting on @emawby, @Jeasmine, and @nan-li)

___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m, line 94 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-MnghsnY00CXwCHnPaOo:-MnghsnY00CXwCHnPaOp:bfon94k) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m#L94)):*
> ```objc
>     NSLog(@"11111 [self.webView loadHTMLString:html baseURL:url];");
>     
>     NSLog(@"222222 [self.webView loadHTMLString:html baseURL:url];");
> ```

Remove `NSLog`, or switch to OneSignal logger.
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m, line 140 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-MngjDAxB3FKHA-J8rs0:-MngjDAxB3FKHA-J8rs1:bv2zucd) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m#L140)):*
> ```objc
>     
>    
>     CGRect mainBounds = UIScreen.mainScreen.bounds;
> ```

These 5 new lines are duplicated in `setIsFullscreen:`. We should deduplicate this code.
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m, line 168 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-MngkC3a1ufjjoeBO34v:-MngkC3a1ufjjoeBO34w:b-a6g458) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m#L168)):*
> ```objc
> 
> - (void)updateSafeAreaInsets {
>     if (@available(iOS 11, *)) {
> ```

Prefer early return over long if block.
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m, line 275 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-MngmglL-zTx_uj2DZfP:-MngmglL-zTx_uj2DZfQ:b-9zaga) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m#L275)):*
> ```objc
> }
> 
> - (NSString *)setContentInsetsInHTML:(NSString *)html {
> ```

This code looks very similar to `updateSafeAreaInsets` in `OSInAppMessageView`. Is there a reason for the difference or could they be made the same?
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m, line 280 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-Mngn13mAu1K0emOFpV2:-Mngn13mAu1K0emOFpV3:b3t99dd) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m#L280)):*
> ```objc
>         UIWindow *keyWindow = UIApplication.sharedApplication.keyWindow;
>         if (!keyWindow) {
>             return newHTML;
> ```

We can't support fullscreen if there isn't a `keyWindow`? Is there another way to support apps that doesn't have this?
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m, line 736 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-Mngjjq0ACk5ykF9_ye2:-Mngjjq0ACk5ykF9_ye3:b-6g5040) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m#L736)):*
> ```objc
>             }
>             case OSInAppMessageBridgeEventTypePageResize: {
>                 // Unused resize event for IAM during actions like orientation changes and displaying an IAM
> ```

Update comment, this resize is used now.

jkasten2 (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-:-Mo0GQfB2s_vZbqDbYcy:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine and @nan-li)

___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m, line 168 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-MngkC3a1ufjjoeBO34v:-Mo0FR4tC2foRaFQBjNu:b4c9lyr) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m#L168)):*

Xcode warnings are not smart enough for that unfortunately. It will complain that you aren't guarding availability. I would rather not use compiler flags to remove the warning
</blockquote></details>

:thumbsup: 
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m, line 275 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-MngmglL-zTx_uj2DZfP:-Mo0G-th8XIFpTTULDcK:b-8fx2u7) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m#L275)):*

updateSafeAreaInsets is using the evaluateJavaScript method on a webView whereas setContentInsetsInHTML is modifying the HTML string. 
</blockquote></details>

:+1: 
___
*[iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m, line 280 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1018#-Mngn13mAu1K0emOFpV2:-Mo0GFWs8hdfSqO5xivp:b4c9lyr) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/6c82f9e6a1cfba44d4455725f6f2abc2ddedbdf0/iOS_SDK/OneSignalSDK/Source/OSInAppMessageViewController.m#L280)):*

So the only way there wouldn't be a keyWindow is if something is going wrong and the IAM is being displayed before any UI has been rendered (this still works for SwiftUI) . The deprecation of keywindow is because it is possible for iOS apps to have multiple windowScenes and this will return a window across all connected scenes, but I think that is actually what we want since we only care about safe area insets. This check is only to avoid the potential rare crash. 

If we cannot get a window then we cannot find out the safeAreaInsets so we cannot adjust the content.
</blockquote></details>

:thumbsup:

Issue #1045, opened 2022-02-08
OneSignal seems to swizzle app delegate methods and `didReceiveRemoteNotification` is not called


hyouuu:
**Description:**
Just updated OneSignal from 2.x to the latest 3.10.0 with SPM. In 2.x there was a `OSHandleNotificationReceivedBlock` that we can set to deal with received remote notification, but 3.x removed it and from the doc we should handle it in appDelegate's 
```
func application(
        _ application: UIApplication,
        didReceiveRemoteNotification userInfo: [AnyHashable : Any],
        fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void)
```

However, it is never called nor `didRegisterForRemoteNotificationsWithDeviceToken` or `didFailToRegisterForRemoteNotificationsWithError`. I turned OneSignal Verbose logging on, and when a silent notification is sent I got these stacktrace:

VERBOSE: onesignalUserNotificationCenter:willPresentNotification:withCompletionHandler: Fired! some message
VERBOSE: finishProcessingNotification: Fired!
VERBOSE: Notification display type: 7
VERBOSE: notificationReceived called! opened: NO
VERBOSE: callLegacyAppDeletegateSelector:withCompletionHandler: Fired!
VERBOSE: finishProcessingNotification: call completionHandler with options: 7
VERBOSE: oneSignalReceiveRemoteNotification:UserInfo:fetchCompletionHandler:
VERBOSE: notificationReceived called! opened: NO

So apparently OneSignal got the remote notification received trigger, but masked the original appDelegate. If in `didFinishLaunchingWithOptions` I explicitly set `application.delegate = self` other weird behaviors would happen.

Could you restore `OSHandleNotificationReceivedBlock` and make `oneSignalReceiveRemoteNotification` call it? Should be pretty straightforward and it would keep the 2.x behavior.

**Environment**
3.10.0 with SPM using the XCFramework version

**Steps to Reproduce Issue:**
  1. Install iOS SDK version 2.8.2 with Cocoapods into a project
  2. Initialize the SDK in didFinishLaunchingWithOptions:
  3. Attempt to receive a push notification

jkasten2:
@hyouuu Thanks for reporting, providing logging, and providing detailed behavior details.

From the log it looks like the OneSignal SDK is trying to call your `didReceiveRemoteNotification: fetchCompletionHandler` method based on this entry:
```
[...]

PR #1048, opened 2022-02-11
Option to provide AppDelegate for SwiftUI lifecycle etc

hyouuu:
Should fix https://github.com/OneSignal/OneSignal-iOS-SDK/issues/1045#issuecomment-1032300940 but I've no idea how to build OneSignal XCFramework to test and haven't touched ObjC for ages so might have issues

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1048)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.h
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.h
@@ -42,6 +42,7 @@
 */
 
 #import <Foundation/Foundation.h>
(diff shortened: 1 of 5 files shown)


jkasten2:
@hyouuu Thanks for the PR! From a quick look it seems this could cover the `UIApplicationDelegateAdaptor` case in issue #1045. However we first would like an investigation if it is possible to auto detect this and correctly handle it without needing a new API and requiring extra setup steps from app developers.

PR #1068, opened 2022-04-04
Log entire `NSError` instead of just `localizedDescription`

nan-li:
# Description
## One Line Summary
Give developers more error details by logging the entire `NSError` instead of just the `localizedDescription`, which many not exist for the error.

### Motivation
These changes were brought up when a customer was getting failure for sending receive receipts but the error logged was `(Error 0 in OneSignal Error.)`, not so informative. This is because the error's `localizedDescription` was being logged when it didn't have that information set. When missing, a default string is constructed from the domain and code.

### Context
To properly provide `localizedDescription`, we would have to provide a key `NSLocalizedDescriptionKey` in the `userInfo` dictionary. Most of the NSErrors created in the SDK use a key of `"error"` instead, and in a few other cases, a key of `"returned"`. Only 1 error at the time of this PR uses `NSLocalizedDescriptionKey`.

The SDK creates and handles errors in a variety of ways, and the scope of this PR is limited to logging.

I also looked at other error logging that used `error.description` and this does give the full error details as well.

### Example Log
Here is an example of what is logged for `error` or `error.description`:
```
Error Domain=OneSignal Error Code=0 "(null)" UserInfo={error=Attempted to perform an HTTP request (OSRequestReceiveReceipts) before the user provided privacy consent.}
```

The "(null)" is the localized description.

### Scope
Just affects logging only.

## Manual testing
Tested what is logged on different errors using the error object itself, `error. localizedDescription`, and `error.description` (which is used in some other instances).

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1068)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSInAppMessageView.m
@@ -67,7 +67,7 @@
     NSString *jsonString;
     if (!jsonData) {
         [OneSignal onesignalLog:ONE_S_LL_ERROR message:
-         [NSString stringWithFormat:@"Error parsing tag dictionary to json :%@",error.localizedDescription]];
+         [NSString stringWithFormat:@"Error parsing tag dictionary to json: %@", error]];
     } else {
          jsonString = [[NSString alloc] initWithData:jsonData encoding:NSUTF8StringEncoding];
     }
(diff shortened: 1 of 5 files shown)


emawby (approved):
This looks great. Thank you for addressing this across the entire SDK!. Can you add to the PR an example of how the log appears when logging the entire error?

PR #1072, opened 2022-04-12
Update OneSignal.podspec to use xcframeworks

emawby:
# Description
## One Line Summary
Updating the `OneSignal.podspec` to distribute the modularized xcframeworks identically to `OneSignalXCFramework.podspec`

## Details
We are unable to distribute a fat framework from the modularized branch using Cocoapods, and we should be using XCFrameworks going forward. This PR makes the two podspecs identical. When we do a major release we can sunset one of the two.

### Motivation
required to distribute the modularized branch for customers still using OneSignal.podspec

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1072)


nan-li (approved):
LTGM

PR #1112, opened 2022-07-28
Add push subscription basic functionality

nan-li:
# Description
## One Line Summary
Add the basic functionality of push subscriptions.

### Motivation
The `OSUser` has a `pushSubscription` property called by `OneSignal.user.pushSubscription`. We want to stub out this class.

### Scope
- The `OneSignalUserManager` may or may not have an `OSUser`.
- An `OSUser` will always have an `OSPushSubscription`.
- An `OSPushSubscription` will always have a `subscriptionId` and `enabled` (which is `false` by default if not provided since on iOS we must always prompt first). It may or may not have a `token` (the push token).
- Add `OSPushSubscriptionObserver` protocol to observe subscription state changes, and rename the existing `OSSubscriptionObserver` to this to be clear it is for the **push** subscription, but it hasn't yet been hooked up properly to trigger.
- The push subscription observer has been rescoped away from the user as it is tied to the device instead. 

# Testing
## Unit testing
Unit testing done is just for checking method and property access, and relying on printed statements to see what is happening.
* In order to start testing, added a public `testCreatePushSubscription()` method in `OSUser` so we can test a user's `pushSubscription`. Shouldn't actually be public or maybe even exist.
* Added a test called `testTheseShouldNotWork` to dump things that should't be accessible or settable as they come up, and comment out as they are addressed/changed.

## Manual testing
Not implemented enough to run properly.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing
   - [ ] Any Public API changes are explained in the PR details and **conform to existing APIs**

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1112)


emawby (approved):
This will need to change post modeling merge

PR #1162, opened 2022-11-23
Find app_id in path if path contains "apps/"

fhboswell:
# Description
## One Line Summary
`app_id` must be included with each request. In order to reduce load on the server we fail requests early on the client that we know will fail on the server. Currently the check in OneSignalRequest looks for `app_id` in the body as a param. However there is another valid way to send the `app_id` that is after "/apps" in the path". This was overlooked initially because it functions correctly in the testing harness `OneSignalClientOverrider` by checking the path and the request body. the PR makes the logic the same in OneSignalRequest and OneSignalClientOverrider.

- Find `app_id` in path if path contains "apps/" 
- Update tests

### Motivation
Getting the error when trying to call the new Live Activity endpoint:

`2022-11-22 16:52:01.080841-0800 OneSignalExample[58470:3775197] Error code: -1 and message: Error Domain=OneSignalError Code=-1 "(null)" UserInfo={error=HTTP Request (OSRequestLiveActivityEnter) must contain app_id parameter}`

`app_id` is in the path rather than a param in the body, this fails the request

# Testing
## Unit testing
Modified OneSignalClientOverrider with the same behavior so that in the future people can count on the testing harness to have the same functionality as the real world case.

## Manual testing
Tested manually with many different cases

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1162)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m
+++ b/iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m
@@ -89,7 +89,14 @@
 }
 
 -(BOOL)missingAppId {
-    return self.parameters[@"app_id"] == nil || [self.parameters[@"app_id"] length] == 0;
+
+  if ([self.path containsString:@"apps/"]) {
+    NSArray *pathComponents = [self.path componentsSeparatedByString:@"/"];
+    NSUInteger x = [pathComponents indexOfObject:@"apps"] + 1; // Find the index that follows "apps" in the path
+    return ([pathComponents count] < x || [[pathComponents objectAtIndex:x] length] == 0);
+  }
+   
+  return (self.parameters[@"app_id"] == nil || [self.parameters[@"app_id"] length] == 0);
 }
 
 -(void)attachBodyToRequest:(NSMutableURLRequest *)request withParameters:(NSDictionary *)parameters {
--- a/iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m
+++ b/iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m
@@ -89,11 +89,10 @@
 }
 
 -(BOOL)missingAppId {
-
   if ([self.path containsString:@"apps/"]) {
     NSArray *pathComponents = [self.path componentsSeparatedByString:@"/"];
     NSUInteger x = [pathComponents indexOfObject:@"apps"] + 1; // Find the index that follows "apps" in the path
-    return ([pathComponents count] < x || [[pathComponents objectAtIndex:x] length] == 0);
+    return ([pathComponents count] < x || [[pathComponents objectAtIndex:x] length] == 0 || [[pathComponents objectAtIndex:x] isEqual: @"(null)"]);
   }
    
   return (self.parameters[@"app_id"] == nil || [self.parameters[@"app_id"] length] == 0);
--- a/iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OSRequests.m
+++ b/iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OSRequests.m
@@ -72,10 +72,6 @@
     
     return request;
 }
-
--(BOOL)missingAppId {
-    return false; //this request doesn't have an app ID parameter
-}
 @end
 
 @implementation OSRequestSendTagsToServer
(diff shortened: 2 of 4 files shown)


fhboswell:
An alternative is that I add

```
-(BOOL)missingAppId {
     return false; //this request doesn't have an app ID parameter
 }
```

to the Live Activity API requests and modify the OneSignalClientOverrider accordingly

iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m
```diff
@@ -89,7 +89,14 @@ -(NSMutableURLRequest *)urlRequest {
 }
 
 -(BOOL)missingAppId {
-    return self.parameters[@"app_id"] == nil || [self.parameters[@"app_id"] length] == 0;
+
+  if ([self.path containsString:@"apps/"]) {
+    NSArray *pathComponents = [self.path componentsSeparatedByString:@"/"];
+    NSUInteger x = [pathComponents indexOfObject:@"apps"] + 1; // Find the index that follows "apps" in the path
+    return ([pathComponents count] < x || [[pathComponents objectAtIndex:x] length] == 0);
```
nan-li, line 96: What situation does `[pathComponents count] < x` account for? I think this would always be `false`.

Thinking about an example `@"apps/%@/"` ...
When the request is created, most of the time the `app_id` is injected by string interpolation, and you account for when it's the empty string `@"apps//"`. I think it can be likely that the app_id is actually `nil` in which case the path is `@"apps/(null)/"`

[[pathComponents objectAtIndex:x]  isEqual: @"(null)"]

nan-li: I agree with changing the method itself than overriding `missingAppId` in the live activity requests, since the new user model requests will also have app_id in the path, not params.

jkasten2 (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1162)* status: all files reviewed, 1 unresolved discussion (waiting on @brismithers, @emawby, and @nan-li)

___
*[`iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m` line 96 at r1](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1162#-NHX3j7Ev_v0qeGYWFk8-r1-96:-NHYCubk6_j-F9yitVcX:bumvh1d) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/95bbebadd51da9f8fc524093a22d57f1e6adc03e/iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m#L96)):*
<details><summary><i>Previously, nan-li (Nan) wrote…</i></summary><blockquote>

I agree with changing the method itself than overriding `missingAppId` in the live activity requests, since the new user model requests will also have app_id in the path, not params.
</blockquote></details>

The `[pathComponents count] < x` ensures it doesn't throw `NSRangeException`, if the  path ends with "apps/" this would be the case.

I believe the `[[pathComponents objectAtIndex:x] length] == 0` check would catch "apps//".

However as Nan noted I it's possible it could be "apps/(null)/" and we need a check for that too. I think we don't have that issue for a dictionary, since they key would just not be there.

iOS_SDK/OneSignalSDK/OneSignalCore/Source/API/OneSignalRequest.m
fhboswell: Good catch nan, thanks `[[pathComponents objectAtIndex:x] isEqual: @"(null)"]`

nan-li:
> The [pathComponents count] < x ensures it doesn't throw NSRangeException, if the path ends with "apps/" this would be the case.

I'm still struggling to understand.
`"a/b/apps/"` -> `["a", "b", "apps", ""]` has count = 4 and x = 3
`"/a/b/apps/"` -> `["", "a", "b", "apps", ""]` has count = 5 and x = 4

I think because we are looking for `"apps/"` and then splitting on `"/"`, it is guaranteed that there is a string at index `x`, even if it's the empty string.

Not a blocker, I just want to understand this.

nan-li (approved):
LGTM

jkasten2 (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1162)* status: all files reviewed, 1 unresolved discussion (waiting on @brismithers, @emawby, and @nan-li)

PR #1163, opened 2022-11-23
activity_id index out of bounds check

fhboswell:
# Description
## One Line Summary
After further testing with Nan, we decided that the index out of bounds check should read <= rather than <.
It is unreachable because the code will only be running if there is a slash after apps `apps/` and therefore the path components array will contain something after apps even if it is an empty string. However I believe this change will be helpful to avoid future mistakes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1163)


nan-li (approved):
LGTM

jkasten2 (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1163#-:-NHa0n_u5e5lmCTwdPrH:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fhboswell)

PR #1172, opened 2022-12-06
[user model] `InAppMessages` namespace

nan-li:
# Description
## One Line Summary
Make the `InAppMessages` namespace and move triggers and IAM methods over.

### Motivation
`InAppMessages` namespace
```objc
+ (void)addTrigger:(NSString * _Nonnull)key withValue:(id _Nonnull)value;
+ (void)addTriggers:(NSDictionary<NSString *, id> * _Nonnull)triggers;
+ (void)removeTriggerForKey:(NSString * _Nonnull)key;
+ (void)removeTriggersForKeys:(NSArray<NSString *> * _Nonnull)keys;
+ (void)clearTriggers;
+ (BOOL)paused NS_REFINED_FOR_SWIFT;
+ (void)paused:(BOOL)pause NS_REFINED_FOR_SWIFT;
+ (void)setInAppMessageClickHandler:(OSInAppMessageClickBlock _Nullable)block;
+ (void)setInAppMessageLifecycleHandler:(NSObject<OSInAppMessageLifecycleHandler> *_Nullable)delegate;
```

### Scope
This also re-enables tag substitution for IAMs. In order to do this we exposed getTags internally from the OneSignalUserManagerImpl. For now we will just use the tags that are already on the user and not try to pull from the server. If it turns out that we aren't hydrating tags prior to displaying IAMs this will need to change.

Also includes a commit addressing some misc `NS_REFINED_FOR_SWIFT` stuff that is unrelated to `InAppMessages.

# Testing
## Unit testing
None, update method calls in some basic test files to see what it looks like.

## Manual testing
None

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1172)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
@@ -396,6 +396,7 @@ static BOOL _isInAppMessagingPaused = false;
 - (void)loadTags {
     self.calledLoadTags = YES;
     // TODO: getTags
+    self.viewController.waitForTags = NO;
 //    [OneSignal getTags:^(NSDictionary *result) {
 //        if (self.viewController) {
 //            self.viewController.waitForTags = NO;
@@ -569,10 +570,10 @@ static BOOL _isInAppMessagingPaused = false;
  Checks if the IAM matches any triggers or if it exists in cached seenInAppMessages set
  */
 - (BOOL)shouldShowInAppMessage:(OSInAppMessageInternal *)message {
-//    return ![self.seenInAppMessages containsObject:message.messageId] &&
-//           [self.triggerController messageMatchesTriggers:message] &&
-//           ![message isFinished] &&
-//           OneSignal.isRegisterUserFinished; // TODO: this crashes app
+    return ![self.seenInAppMessages containsObject:message.messageId] &&
+           [self.triggerController messageMatchesTriggers:message] &&
+           ![message isFinished] &&
+           OneSignalUserManagerImpl.sharedInstance.pushSubscription.subscriptionId != nil;
     return true;
 }
 
--- a/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
@@ -172,8 +172,45 @@ static BOOL _isInAppMessagingPaused = false;
     [self evaluateMessages];
 }
 
-- (void)updateInAppMessagesFromOnSession:(NSArray<OSInAppMessageInternal *> *)newMessages {
-    [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"updateInAppMessagesFromOnSession"];
+- (void)getInAppMessagesFromServer:(NSString *)subscriptionId {
+    [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"getInAppMessagesFromServer"];
+
+    if (!subscriptionId) {
+        [self updateInAppMessagesFromCache];
+        return;
+    }
+    
+    OSRequestGetInAppMessages *request = [OSRequestGetInAppMessages withSubscriptionId:subscriptionId];
+    [OneSignalClient.sharedClient executeRequest:request onSuccess:^(NSDictionary *result) {
+        [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"getInAppMessagesFromServer success"];
+        if (result[@"in_app_messages"]) { // when there are no IAMs, will this still be there?
+            let messages = [NSMutableArray new];
+            
+            for (NSDictionary *messageJson in result[@"in_app_messages"]) {
+                let message = [OSInAppMessageInternal instanceWithJson:messageJson];
+                if (message) {
+                    [messages addObject:message];
+                }
+            }
+            
+            [self updateInAppMessagesFromServer:messages];
+            return;
+        }
+        
+        // TODO: Check this request and response. If no IAMs returned, should we really get from cache?
+        // This is the existing implementation but it could mean this user has no IAMs?
+        
+        // Default is using cached IAMs in the messaging controller
+        [self updateInAppMessagesFromCache];
+        
+    } onFailure:^(NSError *error) {
+        [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"getInAppMessagesFromServer failure"];
+        [self updateInAppMessagesFromCache];
+    }];
+}
+
+- (void)updateInAppMessagesFromServer:(NSArray<OSInAppMessageInternal *> *)newMessages {
+    [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"updateInAppMessagesFromServer"];
     self.messages = newMessages;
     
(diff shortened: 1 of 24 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignal.h
```diff
@@ -141,6 +145,22 @@ typedef void (^OSWebOpenURLResultBlock)(BOOL shouldOpen);
...
++ (void)addTriggers:(NSDictionary<NSString *, id> * _Nonnull)triggers;
++ (void)removeTriggerForKey:(NSString * _Nonnull)key;
++ (void)removeTriggersForKeys:(NSArray<NSString *> * _Nonnull)keys;
++ (void)clearTriggers;
+// Allows Swift users to: OneSignal.InAppMessages.Paused = true
++ (BOOL)paused NS_REFINED_FOR_SWIFT;
++ (void)paused:(BOOL)pause NS_REFINED_FOR_SWIFT;
+
+typedef void (^OSInAppMessageClickBlock)(OSInAppMessageAction * _Nonnull action);
++ (void)setInAppMessageClickHandler:(OSInAppMessageClickBlock _Nullable)block;
```
jkasten2, line 160: Now that this is under the `InAppMessage` namespace this should just be `setClickHandler`.

iOS_SDK/OneSignalSDK/Source/OneSignal.h
```diff
@@ -141,6 +145,22 @@ typedef void (^OSWebOpenURLResultBlock)(BOOL shouldOpen);
...
++ (void)setInAppMessageLifecycleHandler:(NSObject<OSInAppMessageLifecycleHandler> *_Nullable)delegate;
```
jkasten2, line 161: Now that this is under the `InAppMessage` namespace this should just be `setLifecycleHandler`.

iOS_SDK/OneSignalSDK/Source/OSMessagingController.m
```diff
@@ -526,10 +569,11 @@ - (BOOL)hasMessageTriggerChanged:(OSInAppMessageInternal *)message {
  Checks if the IAM matches any triggers or if it exists in cached seenInAppMessages set
  */
 - (BOOL)shouldShowInAppMessage:(OSInAppMessageInternal *)message {
-    return ![self.seenInAppMessages containsObject:message.messageId] &&
-           [self.triggerController messageMatchesTriggers:message] &&
-           ![message isFinished] &&
-           OneSignal.isRegisterUserFinished;
+//    return ![self.seenInAppMessages containsObject:message.messageId] &&
```
jkasten2, line 572: Why was this commented out? Don't we need to keep this logic so triggers work?

iOS_SDK/OneSignalSDK/Source/OneSignalInAppMessaging.m
```diff
@@ -0,0 +1,110 @@
...
++ (Class<OSInAppMessages>)InAppMessages {
+    return self;
+}
+
++ (void)start {
+    // Initialize the shared instance and start observing the push subscription
+    [OSMessagingController start];
+}
+
++ (void)setInAppMessageClickHandler:(OSInAppMessageClickBlock)block {
```
jkasten2, line 42: Same as comments above, can be drop `InAppMessage` from the name, since it is under this new class.

iOS_SDK/OneSignalSDK/Source/OneSignalInAppMessaging.m
```diff
@@ -0,0 +1,110 @@
...
+    [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"In app message click handler set successfully"];
+    [OSMessagingController.sharedInstance setInAppMessageClickHandler:block];
+}
+
++ (void)setInAppMessageLifecycleHandler:(NSObject<OSInAppMessageLifecycleHandler> *_Nullable)delegate; {
```
jkasten2, line 47: Same as comments above, can be drop `InAppMessage` from the name, since it is under this new class.

jkasten2 (changes requested):

emawby (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1172)* status: 3 of 23 files reviewed, 5 unresolved discussions (waiting on @fhboswell, @jennantilla, and @jkasten2)

___
*[`iOS_SDK/OneSignalSDK/Source/OSMessagingController.m` line 572 at r3](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1172#-NI_xPAzyrcBLDkuueJo-r3-572:-NIivpEM1FoJEDojt2t-:b-xwl6rv) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/8dd9a4b7cd9bf62ea79138b2f2e19856739d71be/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m#L572)):*

Why was this commented out? Don't we need to keep this logic so triggers work?
</blockquote></details>

added this back but instead of isRegisterUserFinished we are just checking if we have the push subscription

nan-li:
> I see a number of `respondsToSelector:@selector` checks on `OSNotificationsManager.delegate` 
> 
> However as far as I can tell all methods on this protocol are required. And we are not swizzling or doing anything dynamic here, `OSNotificationsManager` is our class, so it seems like we can remove these checks from what I can tell.

True, I added these back when the delegate could be null, just a null check would be sufficient.

jkasten2 (approved):
> Reviewed 8 of 11 files at r2, 4 of 8 files at r3, 3 of 4 files at r4, 2 of 3 files at r6, 1 of 1 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1172)* status: all files reviewed, 1 unresolved discussion (waiting on @fhboswell, @jennantilla, and @nan-li)

___
*[`iOS_SDK/OneSignalSDK/Source/OSMessagingController.m` line 192 at r8](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1172#-NIkESiR-nKSIw8d4-RP:-NIkESiR-nKSIw8d4-RQ:b2r5f8f) ([raw file](https://github.com/onesignal/onesignal-ios-sdk/blob/44d5657684730ec440f69b7f84f0d216ecde1d12/iOS_SDK/OneSignalSDK/Source/OSMessagingController.m#L192)):*
> ```objc
>     [OneSignalClient.sharedClient executeRequest:request onSuccess:^(NSDictionary *result) {
>         [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"getInAppMessagesFromServer success"];
>         if (result[@"in_app_messages"]) { // when there are no IAMs, will this still be there?
> ```

Yes, just updated our doc to define it should always be there. If there are no IAMs then it will be an empty array.

PR #1200, opened 2023-01-04
[user model] More fixes for alpha release (round 3)

nan-li:
# Description
## One Line Summary
More fixes as arose during testing (round 3).

### Misc
- No longer point to `staging`
- Refined More APIs for Swift
- Update Common Defines variables such as rename `OSUD_PLAYER_ID_TO` to `OSUD_PUSH_SUBSCRIPTION_ID` for clarity and remove some no longer used ones.
- Cleanup no longer used OneSignalRequests
- Add a `hydrating` check when a model is added to store (currently only used when adding email and sms)
- Need to prepare for a new user before setting the user to `nil` as the previous user may be needed in the preparation.
- On success, remove the create subscription request even if not hydrating. It can return just `success` with no subscription info if this subscription already exists.

### Separate model store just for the push subscription
- There are now 2 subscription model stores and 2 subscription model store listeners.
- Motivation: push sub behaves differently from other subscriptions and is not cleared from the store when a user changes. Separate it into its own store so that the email and sms subscriptions stores can be cleared but the push subscription store is not.

### Fix push subscription hydration logic _(still faulty)_
- Motivation: Hydrating the push subscription is problematic. We will hydrate it if:
  - the push subscription ID does not already exist locally
  - the type of the returned subscription is `iOSPush`
  - the token at the time the request is made matches the returned token
  - we hydrate using the first matching subscription that is returned (there may be multiple iOS push subscriptions in the response but we will accept the first one that fulfills our conditions above)
- This can still be faulty behavior but we get as close as we can to correct behavior for now.

### Connecting Deltas and Requests read from cache to the appropriate models
* Motivation: When initializing the delta queue and request queue, the executors need to hook up the Deltas' and Requests' models to the ones in the store, so that they can get the updated information needed to send the requests, and hydrate the correct models.
* When model stores, deltas, and requests are initialized from cache, they lose their references as they are new instances.
* To simplify uncaching, we maintain separate request queues for each request type in the executor

### Fix executors accessing `OneSignalUserManagerImpl.sharedInstance` too early
* Motivation: Must be careful of accessing `OneSignalUserManagerImpl.sharedInstance` before it's actually finished initializing, as the initializer for it does a lot of things!
* And we may need to note if any other areas can have the same issue.

### Fetch User Request after Create User Request succeeds
* Motivation: Create User's response won't send us the user's complete info if this user already existed.
* For now, parse the Create User response and make the Fetch User response, as we should parse to get the subscription_id from this request.
* TODO: Differentiate if we need to Fetch based on response code to not make extraneous call.

### Create email/sms subscription parameters update
* It seems we cannot send `enabled = true` without sending `notification_types` for creating an email or SMS. Even sending `-99` is ok.
* Or, we send neither.
* The server is still in flux. For now, don't send any of these, but we need to revisit

# Testing
## Manual testing
Lots on iPhone 13 physical.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1200)


emawby (approved):

iOS_SDK/OneSignalDevApp/OneSignalNotificationServiceExtension/OneSignalNotificationServiceExtension.entitlements
```diff
@@ -4,7 +4,7 @@
 <dict>
 	<key>com.apple.security.application-groups</key>
 	<array>
-		<string>group.com.onesignal.example.onesignal</string>
```
fhboswell, line 7: Staging group

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalCommonDefines.h
```diff
@@ -33,8 +33,8 @@
 // Networking
 #define OS_API_VERSION @"1"
 #define OS_API_ACCEPT_HEADER @"application/vnd.onesignal.v" OS_API_VERSION @"+json"
-#define OS_API_SERVER_URL @"https://api.staging.onesignal.com/"
-#define OS_IAM_WEBVIEW_BASE_URL @"https://staging.onesignal.com/"
```
fhboswell, line 37: The bundle id and group have staging in them but the app points at prod, just a heads up.

fhboswell (approved):

iOS_SDK/OneSignalDevApp/OneSignalNotificationServiceExtension/OneSignalNotificationServiceExtension.entitlements
nan-li: Not sure I understand?

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalCommonDefines.h
nan-li: Ah I see, the app points at prod for sending requests, but the bundle ID is different. That's the bundle ID for this app_id.

nan-li: Ohh ahh wow, good call. I wasn't thinking about that this app_id only exists in the staging environment. This is why I was having some issues.

PR #1203, opened 2023-01-09
[user model] replace user data with the fetch user response

nan-li:
# Description
## One Line Summary
_**Replace**_ the local user data with the response from a fetch user request.

### Motivation
* Motivation: Previously, when we hydrate after fetch user, we are adding to our models, but not doing any removals if local properties are not in the response.
* Instead, do a replacement of all the user's data with the response, but keep the same identity and property model since it is the same user.
* Currently, this user data is just the tags, aliases, and subscriptions (non-push)

# Testing
## Manual testing
On physical iPhone 13 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1203)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSUserRequests.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSUserRequests.swift
@@ -66,6 +66,13 @@ class OSUserExecutor {
      Used to parse Create User and Fetch User responses. The `originalPushToken` is the push token when the request was created, which may be different from the push token currently in the SDK. For example, when the request was created, there may be no push token yet, but soon after, the SDK receives a push token. This is used to determine whether or not to hydrate the push subscription.
      */
     static func parseFetchUserResponse(response: [AnyHashable: Any], identityModel: OSIdentityModel, originalPushToken: String?) {
+
+        // If this was a create user, it hydrates the onesignal_id of the request's identityModel
+        // The model in the store may be different, and it may be waiting on the onesignal_id of this previous model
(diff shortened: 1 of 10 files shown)


iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSUserRequests.swift
```diff
@@ -269,6 +266,9 @@ class OSUserExecutor {
         }
 
         OneSignalClient.shared().execute(request) { response in
+            // Clear local data in preparation for hydration
+            OneSignalUserManagerImpl.sharedInstance.clearUserData()
```
emawby, line 270: Should this be inside the`if let response = response` block? What would it mean to have a nil response?

emawby (changes requested):
Just one question

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSUserRequests.swift
nan-li: Yes good call! It makes more sense inside the block. I have no idea what it would mean to have a successful yet `nil` response. Better to not make any local changes in that case.

emawby (approved):

PR #1204, opened 2023-01-10
[user model] add more properties to push subscription object

nan-li:
# Description
## One Line Summary
Add more properties to the push subscription object as we realized we needed to pass `test_type` to the subscription object during Create User to receive push notifications _(but also fixed a couple other things below)_.

## Details
We should revisit all the properties to include in the push subscription model but for now, these are sufficient to work.

**Move `OneSignalMobileProvision` to OneSignalCore**
* This is used in the User Module to get the `test_type` for the push subscription model
* So it needs to move out of OneSignalFramework into Core
* Rename `UIApplicationReleaseMode` to `OSUIApplicationReleaseMode` since it is defined by us, not by Apple APIs

**Add more properties to push subscription model**
* Motivation: we need to send `test_type` in the push subscription model because this is what tells the backend the APNS environment (prod, sandbox)
* Also add a few other properties that used to exist like `device_model` and the `sdk` version.
* Add these properties to the subscription object to the Create User request

**Now that we fixed receiving notifications, testing showed that the `OSNotificationsManager` was not getting push subscription ID and appId correctly.**
* Problem: The `_appId` in the `OSNotificationsManager` was never being set, so it was always `nil`. Instead, let's use the appId from  `OneSignalConfigManager`, who should be controlling this data.
* Also, replace accessing `_pushSubscriptionId` with `[self pushSubscriptionId]` which will init it from user defaults if `nil`.

**Fix bug: need to update identity model even when user changed**
* Noticed a bug if someone calls `login` right away upon SDK init. I had been testing with a button tap, giving too much time in between to notice this bug earlier.
* The initial anonymous create user request is sent, then the model in the stores are replaced with new models when the `login` is called (which will result in an Identify User request). This request is waiting on the `onesignal_id` of the previous user in order to be sent.
* However, when the first create user request returns, we see that the user has changed, so we weren't hydrating anything.
* As a result, the identify user request could never be sent.
* Instead, always hydrate the identity model in the request first. Then we can hydrate the models in the stores, if the user is the same.

# Testing
## Manual testing
Physical iphone on production.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/1204)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
+++ b/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
@@ -68,6 +68,8 @@
 		3C2C7DC6288E00AA0020F9AE /* UserModelObjcTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 3C2C7DC5288E00AA0020F9AE /* UserModelObjcTests.m */; };
 		3C2C7DC8288F3C020020F9AE /* OSSubscriptionModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C2C7DC7288F3C020020F9AE /* OSSubscriptionModel.swift */; };
 		3C2D8A5928B4C4E300BE41F6 /* OSDelta.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C2D8A5828B4C4E300BE41F6 /* OSDelta.swift */; };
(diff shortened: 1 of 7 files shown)


emawby (approved):
awesome!

PR #1222, opened 2023-02-02
[5.0.0] Caching and error handling in client and executors

nan-li:
# Description
## One Line Summary
Improve **caching** and **uncaching** of requests when the app is started up, client **error** handling, and **retries** process.

### Motivation
This PR tries to improve the caching and uncaching of requests, error handling and retries in the client.

🚧  🔧 The caching, uncaching, error handling, retries process is still rudimentary in this PR and expected to undergo continued improvement.

### Scope
**Client will retry requests with a backoff strategy & retry 4 times**
* Keep max retries for now so we don't spam the server with requests when it is in trouble to avoid retrying forever.
* Let a request retry 4 times (so total 5 times) with backoff strategy calculated with powers of 3 times 5.
* For now, no strategy to trigger retries after these 5 times until app is started up again

**Updates to the Executors**
* When initializing requests from cache, hook models back up correctly so all references to the same model are correct. A request may be waiting on an identity model to fill in the `onesignal_id`, for example, or a `subscription_id` on a subscription model.
  * To aid this, the User Executor maintains a list of identity models it has processed, and the subscription executor maintains a list of subscription models it has processed.
* Removing requests from executors' queue and cache after success or an un-retryable failure (still WIP in terms of finer details)
  * While client is retrying, keep the request in the executors and don't flush the same request as the client retries.

**Update the push subscription object right before a request with it is sent**
* The push subscription may not be complete at the moment a push subscription is snapshotted for a request like Create User
* If it gets updated with the subscription_id, for example, we want that to be reflected.

**Update the Example App**
* Update methods and adding buttons for user model methods

**Remove `retain_previous_user` flag**
* No longer accepted by the server when transferring a push subscription.
* Server checks anonymous user for whether it has any aliases set. If so, we’ll keep the user. If there aren’t aliases, we’ll remove it. Done behind the scenes in Perseus.

# Testing
## Manual testing
iPhone 13
- testing scenario: Making CreateUser requests without internet, kill app, startup again with internet and those should go through correctly and be in the correct state.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1222)


emawby (commented):
> Reviewed 14 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1222)* status: 14 of 18 files reviewed, all discussions resolved (waiting on @brismithers, @fhboswell, @jennantilla, @jkasten2, and @rgomezp)

emawby (approved):

PR #1244, opened 2023-04-07
Enable push notifications for iOS simulators

emawby:
# Description
## One Line Summary
Always set test type to 1 for simulators

## Details
Our provision detection in OneSignalMobileProvision is not successfully able to find the embedded mobile provision for simulators. This might only affect Xcode managed provisioning profiles.

This change is a workaround to that issue to always set testType to 1 since simulators need to receive push in the sandbox environment.

### Motivation
Enable testing push notifications for developers that don't have physical iOS devices.

### Scope
iOS push on simulators

# Testing
## Manual testing
Tested using Xcode 14.3 with iOS 16 simulators

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] I have filled out all **REQUIRED** sections above
   - [x] PR does one thing
     - If it is hard to explain how any codes changes are related to each other then it most likely needs to be more than one PR
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [x] I have included test coverage for these changes, or explained why they are not needed
   - [x] All automated tests pass, or I explained why that is not possible
   - [x] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [x] Code is as readable as possible.
      - Simplify with less code, followed by splitting up code into well named functions and variables, followed by adding comments to the code.
   - [x] I have reviewed this PR myself, ensuring it meets each checklist item
      - WIP (Work In Progress) is ok, but explain what is still in progress and what you would like feedback on. Start the PR title with "WIP" to indicate this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1244)

PR #1297, opened 2023-08-10
[User model] enable push for simulators

emawby:
# Description
## One Line Summary
Enable push for simulators. Equivalent to #1244 

## Details
Our provision detection in OneSignalMobileProvision is not successfully able to find the embedded mobile provision for simulators. This might only affect Xcode managed provisioning profiles.

This change is a workaround to that issue to always set testType to 1 since simulators need to receive push in the sandbox environment.

### Scope
push on simulators

# Testing

## Manual testing
tested receiving push on a simulator

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1297)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
@@ -218,6 +218,11 @@ class OSSubscriptionModel: OSModel {
         // Set test_type if subscription model is PUSH
         if type == .push {
             let releaseMode: OSUIApplicationReleaseMode = OneSignalMobileProvision.releaseMode()
+            #if targetEnvironment(simulator)
+            if (releaseMode == OSUIApplicationReleaseMode.UIApplicationReleaseUnknown) {
+                self.testType = OSUIApplicationReleaseMode.UIApplicationReleaseDev.rawValue
+            }
+            #endif
             // Workaround to unsure how to extract the Int value in 1 step...
             if releaseMode == .UIApplicationReleaseDev {
                 self.testType = OSUIApplicationReleaseMode.UIApplicationReleaseDev.rawValue


iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
```diff
@@ -217,6 +217,9 @@ class OSSubscriptionModel: OSModel {
 
         // Set test_type if subscription model is PUSH
         if type == .push {
+            #if targetEnvironment(simulator)
+            self.testType = OSUIApplicationReleaseMode.UIApplicationReleaseDev.rawValue
+            #endif
```
jkasten2, line 222: `self.testType` will just be overwrite below. In order to prevent that we need to use an `#else` here or some kind of early return logic.

jkasten2 (changes requested):

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
emawby: That is fine if we are able to properly detect that it is an AdHoc or wildcard build on the simulator I want it to be overwritten. The issue is when the release mode is unknown

jkasten2: ah, that wasn't clear. I recommend a comment or explicitly only doing then logic when it is unknown then.

emawby: I updated it to only set if the mode is unknown

jkasten2 (approved):
👍

PR #1240, opened 2023-03-29
Fix direct session count

shepherd-l:
# Description
## One Line Summary
Reset direct sessions when the session ends.

### Motivation
Fix incorrect direct session count for notifications, prevents a direct session count from being sent up twice for the same notification.

# Testing
## Manual testing
Tested sending notifications from a direct session to another direct session, app build with Xcode 14.2 with the OneSignal example app on an iPhone 12 with iOS 15.5.

Steps
1. Send 2 notifications
2. Open notification 1 and immediately background the app
3. Wait for session to end (30+ seconds)
4. Open notification 2

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [x] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] I have filled out all **REQUIRED** sections above
   - [x] PR does one thing
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [x] I have included test coverage for these changes, or explained why they are not needed
   - [x] All automated tests pass, or I explained why that is not possible
   - [x] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [x] Code is as readable as possible.
   - [x] I have reviewed this PR myself, ensuring it meets each checklist item

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1240)

PR #1327, opened 2023-10-25
[v5] Fix crash with direct influence but nil direct id

nan-li:
# Description
## One Line Summary
Cherry-pick of https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1311 from `main`.

### Motivation
Fixes a crash when a session has a direct influence but the direct id is nil.

This does not resolve getting into this situation in the first place since we should not have a direction session without a direct id, but this will at least prevent a crash when the issue happens. Instead we will log an error.

# Testing
## Unit testing
Unit test was also cherry-picked even though it will not run.

## Manual testing
None on this PR

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [x] Outcomes
   - [x] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing
   - [x]I have personally tested this on my device, or explained why that is not possible

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1327)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalOutcomes/Source/Influence/OSChannelTracker.m
+++ b/iOS_SDK/OneSignalSDK/OneSignalOutcomes/Source/Influence/OSChannelTracker.m
@@ -139,9 +139,13 @@ THE SOFTWARE.
     
     if (_influenceType == DIRECT) {
         if ([self isDirectSessionEnabled]) {
-            NSArray *ids = [NSArray arrayWithObject:_directId];
-            builder.ids = ids;
-            builder.influenceType = DIRECT;
+            if (_directId) {
+                NSArray *ids = [NSArray arrayWithObject:_directId];
+                builder.ids = ids;
+                builder.influenceType = DIRECT;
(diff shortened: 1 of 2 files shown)


emawby (changes requested):
I think this was a change that we might've ended up reverting because it caused other issues. Could you check the current state of main? Is this change there or was it reverted?

nan-li:
> I think this was a change that we might've ended up reverting because it caused other issues. Could you check the current state of main? Is this change there or was it reverted?

Main contains these same changes. It was this PR https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1240 that was reverted in https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1251.

emawby (approved):

PR #1345, opened 2023-12-01
Fix legacy player ID was not cached properly when upgrading from v3

nan-li:
# Description
## One Line Summary
Correctly cache push subscription ID when upgrading from v3 for use in receive receipts and notification opened requests.

## Details

**Correctly cache push subscription ID when upgrading from v3**
We were not saving the `OSUD_PUSH_SUBSCRIPTION_ID` to user defaults when we detected a legacy player ID and proceed with upgrading.
It was only saved from create user responses that hydrate the subscription model, which does not happen on v3 -> v5 upgrade.
Receive receipts and notification opened requests use this value from the cache.

**Save push subscription ID on app start**
Because of the bug that this PR aims to fix, players that upgraded from v3 -> v5 were missing the player ID in cache for sending to notification opened and notification received endpoints.
Now, to remedy, every time app cold starts, as part of the push subscription model checking itself for updates, we will cache its subscription ID if non-empty.

**Future Work**
Consider some sort of migration in Core that the extension and user manager can utilize.
Also, this will not fix the issue for users that updated their app and did not open it yet.

### Motivation
It was brought up by other team members that many receive receipts and notification opened requests are missing the player ID.

### Scope
Sending non-null player ID to receive receipts and notification opened endpoints.

# Testing
## Unit testing
None, should consider adding

## Manual testing
iPhone 13 on iOS 17.1

**Scenario 1 Tested mimics v3 -> v5 -> this PR**
1. Install and open app on v3
  a. see player created in dashboard
2. Update app to `main` which is  v5.0.4, prior to this PR
  a. see legacy player get migrated in SDK
  b. player in dashboard remains the same but with sdk version updated to 5.0.4
  c. Receive receipts and notification opened will be missing player ID
3. Update app to this PR and open app
  a. Now receive receipts and notification opened will have player ID
 
**Scenario 2 Tested mimics v3 -> this PR**
1. Legacy player is migrated in SDK and requests have player ID

**Scenario 3 Tested using new app install**
1. New user created and requests have player ID

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [x] Open
      - [ ] Push Processing
      - [x] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1345)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
@@ -425,6 +425,9 @@ extension OSSubscriptionModel {
         netType = OSNetworkingUtils.getNetType() as? Int
         // sdkType ??
         // isRooted ??
+        if type == .push && !(subscriptionId ?? "").isEmpty {
+            OneSignalUserDefaults.initShared().saveString(forKey: OSUD_PUSH_SUBSCRIPTION_ID, withValue: subscriptionId)
+        }
     }
     
     enum OSPushPropertyChanged {


jkasten2 (approved):
This fixes the bug for new v3 -> v5 migrations for all cases, but it doesn't fix confirmed deliveries for those who already upgrade. Also it probably doesn't fix the not notification open REST API call if someone opens the app for the first time after the v3->v5 upgrade via push, due to the update() code running to late

PR #1352, opened 2024-01-03
Location sharing with OneSignal now defaults to false

nan-li:
# Description
## One Line Summary
Location sharing with OneSignal now defaults to `false` regardless of device permissions, so developers must explicitly set sharing to `true`.

### Motivation
It may be a poor developer experience to be opted-in by default. Previously, as long as device has location permission, we shared location with the OneSignal server. Now, sharing must be explicitly enabled. This aligns with recent similar Android behavior changes https://github.com/OneSignal/OneSignal-Android-SDK/pull/1942. 

### Scope
**⚠️ Behavior changes:**
- Developers must now explicitly call `OneSignal.Location.isShared = true` (Swift) or  `[OneSignal.Location setShared:true]` to share location with OneSignal regardless of permission or prompting for location.

**Fix location module bug** 
- Use constants for Location and IAM module classes that are used with `NSClassFromString`: `OneSignalLocationManager` and `OneSignalInAppMessages`
- These hardcoded classes can be easily missed. When the class `OneSignalLocation` was renamed to `OneSignalLocationManager`, some of the strings were missed in `NSClassFromString` calls. One consequence of this is the location module's `start` method never ran and location not shared.

# Testing
## Unit testing
None

## Manual testing
iPhone 13 on iOS 17.1.2

- Enable and then request permission → shares location when both steps are done
- Request permission and then enable → shares location when both steps are done

Something that is not working is mid-session disabling and enabling doesn't drive location sharing until next cold start.

1. Already sharing location
2. Disable sharing - location stops being tracked and shared
3. Enable sharing - location sharing doesn't start
4. New cold start - location is shared

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes
   - [x] Location sharing

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1352)

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -694,11 +694,11 @@ OneSignal.Notifications.addClickListener(myListener)
 
 ## Location Namespace
 
-The Location namespace is accessible via `OneSignal.Location` and provide access to location-scoped functionality.
+The Location namespace is accessible via `OneSignal.Location` and provide access to location-scoped functionality. As of `v5.1.0`, location sharing must be enabled by setting `isShared` to `true`.
 
 | **Swift**                                                                                      | **Objective-C**                                                                              | **Description**                                                                                                                                          |
 | ---------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `let isShared: Bool = OneSignal.Location.isShared`<br><br>`OneSignal.Location.isShared = true` | `BOOL isShared = [OneSignal.Location isShared]`<br><br>`[OneSignal.Location setShared:true]` | *Whether location is currently shared with OneSignal.*                                                                                                   |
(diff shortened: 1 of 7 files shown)


emawby (changes requested):
Should we update the migration guide to call out the new default as well?

nan-li:
> Should we update the migration guide to call out the new default as well?

Done. We will also need to update our docs calling this out for v5.1.0 and above.

jennantilla (approved):
Looks great! Confirmed that `isShared` now defaults to false.

emawby (approved):

PR #1378, opened 2024-03-05
Update cd.yml

emawby:
Use github.events.inputs instead of env for accessing workflow dispatch variables

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1378)


nan-li (approved):
I recall I used this to release before, and it did work. I ran into an unrelated issue.

PR #1412, opened 2024-04-18
Fix crashes when encoding user models

nan-li:
# Description
## One Line Summary
Fix crashes during model encoding due to concurrent data access.

## Details
- Models are encoded often: any change encodes them in the model store, when deltas are cached in the operation repo, when deltas are flushed in the operation repo, when deltas are enqueued to executors, when deltas are turned into requests, when requests are sent and removed, etc. These all have references to models.

- Identity Model and Properties Model have access to their aliases and tags dictionary wrapped in a lock. However, this lock was missing from the encode method, and is necessary.

- The changes made in this PR already avoid recursive locking, but let's also replace the lock that is used to a `NSRecursiveLock` for future-proofing. During development, I ran into this before trying to avoid recursive locking as models are encoded so often. This occurred when tags change on a model (in a lock), driving the model store to encode it's models, which eventually calls the model's own encode method (which is also in a lock).

- Another change in this PR is to reduce extra caching of the Identity Model.

### Motivation
Fixes any stack trace with this signature.

```
0  libobjc.A.dylib                0x1814 objc_msgSend + 20
1  Foundation                     0x12284 _encodeObject + 172
2  OneSignalUser                  0x12a14 __swift_destroy_boxed_opaque_existential_0 + 6928
3  OneSignalUser                  0x12a74 __swift_destroy_boxed_opaque_existential_0 + 7024
...
8  Foundation                     0x547f4 -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 344
...
11 OneSignalCore                  0xd69c -[OneSignalUserDefaults saveCodeableDataForKey:withValue:] + 88
```

### Scope
Locking in models

### Alternative to Consider

- Many objects have a model reference, and encodes the model when itself is being cached.
- However, most of these objects don't need all the data in the model.
- For example, an `OSDelta` may references an Identity Model, but only for the purpose of identifying the user and retrieving the OSID.
- An alternative would be models are cached in one place and objects with references would not cache the whole model.

# Testing
## Unit testing
Added tests reproducing crash when encoding models with concurrency
* Adds two tests that encode Identity Model and Properties Model while their aliases and tags dictionary are written to.
* The tests were run manually in Xcode with "Run "testName" Repeatedly..." **100 times**.
* The crash would be reproduced before the changes in this PR.

## Manual testing
Device: iPhone 13 with iOS 17.4
- reproduced occasionally following similar steps to the tests, before changes in the PR

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1412)


nan-li:
Just pushed one more change to remove an unused reference that was being cached.

emawby (approved):

PR #1417, opened 2024-04-24
Fix crash when handling a dialog result

emawby:
# Description
## One Line Summary
Fix a crash when handling a dialog result (often from requesting location or notification permission when permission has already been denied.)

## Details
The stack traces point to crashing when calling `[self.queue removeObjectAtIndex:0];` in `delayResult`. This was never reproducible as the queue should never be empty when calling delayResult, but this PR adds thread safety and a sanity check to protect and this crash.

- Synchronize queue reads and writes on self.queue

- Additionally verify that queue.count > 0 prior to removing object at index 0

### Motivation
Fix crash

### Scope
iOS dialogs

# Testing
## Unit testing
no infrastructure in place yet for mocking handling prompting actions

## Manual testing
Tested by spamming the permission dialog actions 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1417)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignalDialogController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignalDialogController.m
@@ -67,9 +67,8 @@
             //if so, we shouldn't present on top of existing dialog
             if (self.queue.count > 1)
                 return;
-            
-            [self displayDialog:request];
         }
+        [self displayDialog:request];
     });
 }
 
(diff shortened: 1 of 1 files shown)


nan-li (approved):
Do you know if there are consequences to calling `displayDialog` within the lock?

emawby:
> Do you know if there are consequences to calling `displayDialog` within the lock?

I moved the calls outside of the sync blocks since I don't think they need to be inside.

nan-li (approved):

PR #1418, opened 2024-04-25
WIP [Bug] Some pending properties can be sent to new user incorrectly, when users change

nan-li:
The WIP component is thorough testing still in progress.

# Description
## One Line Summary
Fixes bug when users change, old User Properties like tags can be sent to the new user.

### Motivation
**REQUIRED -** Why is this code change being made? Or what is the goal of this PR? Examples: Fixes a specific bug, provides additional logging to debug future issues, feature to allow X.

### Scope
- OSDelta has a new field containing the ID of the identity model it is meant for
- When uncaching, the old OSDeltas will be dropped since they will be missing the Identity Model ID
  - It is unlikely for them to still exist as we flush them into Requests when users change or backgrounding
- Adds a lightweight Identity Model Repo to manage instances that are still being used
  - An alternative considered is refactoring the Identity Model Store and would require it to basically override all of the OSModelStore methods

### OPTIONAL - Other
**OPTIONAL -** Feel free to add any other sections or sub-sections that can explain your PR better.

# Testing
## Unit testing
**OPTIONAL -**  Explain unit tests added, if not clear in the code.

## Manual testing
**RECOMMEND - OPTIONAL -** Explain what scenarios were tested and the environment.
Example: Tested opening a notification while the app was foregrounded, app build with Android Studio 2020.3 with a fresh install of the OneSignal example app on a Pixel 6 with Android 12.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] I have filled out all **REQUIRED** sections above
   - [ ] PR does one thing
     - If it is hard to explain how any codes changes are related to each other then it most likely needs to be more than one PR
   - [ ] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [ ] I have included test coverage for these changes, or explained why they are not needed
   - [ ] All automated tests pass, or I explained why that is not possible
   - [ ] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [ ] Code is as readable as possible.
      - Simplify with less code, followed by splitting up code into well named functions and variables, followed by adding comments to the code.
   - [ ] I have reviewed this PR myself, ensuring it meets each checklist item
[...]

PR #1427, opened 2024-05-06
[Bug] Fix processing of Identify User response if the user has changed since then

nan-li:
# Description
## One Line Summary
Fix processing of Identify User response when the user has changed since, so that pending updates will still be sent and transfer subscription happens only in correct situations.

## Details

Note that this bug only manifests if multiple logins to different users are called one after another, or the requests are not able to succeed (server is down, or there is no internet connection during which time multiple logins are called). This PR tests and addresses those scenarios.

### Motivation
Consider the scenario **Anon** -> Login User A (**Identify User**) -> Login User B (**Create User**)
- Dropping user updates
  - When the Identify User request returns, because the user is now different, we were not fetching User A to hydrate the OSID, so any updates made to it get dropped. Now, we mark the Identity Model to use the external_id for pending requests.
- Transferring push subscription incorrectly
  - In addition, transfer subscription had a bug. Create User B already contains the push subscription in the payload, but when Identify User A returns, we still enqueue a Transfer Subscription to User A request, which would create this flow: `Anon -> Login User A (Identify User) -> Login User B (Create User) -> Transfer Subscription to User A`
  - Now we only transfer if it's still the same user. Any Create User requests would already have the push subscription in it.
  - This was a recoverable bug as the next Fetch User B call would find the push subscription gone, and the SDK would call Create Subscription for User B with the push subscription data.

This PR also fixes the bugs mentioned under Manual Testing section for https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1418

### User State Change Observer
⚠️ The User Observer was never triggered for the 409'd middle user. That is still the case after this PR, as we never hydrate its `onesignal_id`.

### Scope
- Main logic change is in commit [[Bug] Fix Identify User processing if the user has changed](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1427/commits/f7f2a066cfa6bf46a9c266e6689b0ae69e177a4a) and [Avoid a fetch user after login returns 409](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1427/commits/f214e698219b129a0b395fd76b5f0e6646cacded)
- Other changes are tests-related

# Testing
## Unit testing
- Changes to mock client to process requests asynchronously
- Add test to reproduce behavior in [[tests] Add test to reproduce Identify User when in middle bug](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1427/commits/33244c09c961ef64a2a611cd1f6b5172ec34bcc5) which fails prior to the fix

## Manual testing
Tested on iPhone 13 with iOS 17.4

### ✅ Scenario 1:
1. On **new app install** after initialization, immediately call these methods.
2. Confirm tags are sent to `<some_existing_euid>` by way of external_id
```objc
// New App Install

[OneSignal initialize:<app_id> withLaunchOptions:launchOptions];

[OneSignal login:<some_existing_euid>]; // returns 409
[OneSignal.User addTagWithKey:@"a" value:@"a"];
[OneSignal login:<user_b>];
```

### ✅ Scenario 2:

1. On **new app install** after initialization, immediately call these methods.
2. Confirm every update is sent and is sent to the correct user.
```objc
// New App Install

[OneSignal initialize:<app_id> withLaunchOptions:launchOptions];

[OneSignal.User addTagWithKey:@"a" value:@"a"];
[OneSignal.User setLanguage:@"fr"];
[OneSignal.User addAliasWithLabel:@"a" id:@"a"];
[OneSignal.User addEmail:@"a@example.com"];

[OneSignal login:<some_existing_euid>];
[OneSignal.User addAliasWithLabel:@"existing" id:@"existing"];
[OneSignal.User addTagWithKey:@"existing" value:@"existing"];
[OneSignal.User setLanguage:@"zh"];
[OneSignal.User addEmail:@"existing@example.com"];

[OneSignal logout];
[OneSignal.User addTagWithKey:@"c" value:@"c"];
[OneSignal.User setLanguage:@"es"];
[OneSignal.User addEmail:@"c@example.com"];
[OneSignal.User addAliasWithLabel:@"anonymous" id:@"anonymous"];
```

### ✅ Scenario 3:

1. Check everything is correct when the current user is still the same as the Identified User (no regression).
2. On **new app install** after initialization, immediately call these methods.
3. Confirm every update is sent and is sent to the correct user.
4. The user in the SDK is correctly hydrated.
```objc
// New App Install

[OneSignal initialize:<app_id> withLaunchOptions:launchOptions];

[OneSignal.User addTagWithKey:@"a" value:@"a"];
[OneSignal.User setLanguage:@"fr"];
[OneSignal.User addAliasWithLabel:@"a" id:@"a"];
[OneSignal.User addEmail:@"a@example.com"];

[OneSignal login:<some_existing_euid>];
[OneSignal.User addAliasWithLabel:@"existing" id:@"existing"];
[OneSignal.User addTagWithKey:@"existing" value:@"existing"];
[OneSignal.User setLanguage:@"zh"];
[OneSignal.User addEmail:@"existing@example.com"];

// Don't logout or login to another user
```

### ✅ Scenario 4:
1. Turn off wifi and data (**no network connection**)
2. New app install and call the following methods
```objc
// New App Install

[OneSignal initialize:<app_id> withLaunchOptions:launchOptions];
[OneSignal login:<some_existing_euid>];
[OneSignal.User addTagWithKey:@"a" value:@"a"];
[OneSignal login:<user_b>];

// No requests go through
```
5. Wait 30 seconds and kill app
6. Turn on wifi and reopen app
7. Requests go to the correct users.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs - **none**

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1427)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
+++ b/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
@@ -116,6 +116,8 @@
 		3CA283A92B86A30400097465 /* OneSignalCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = DE7D17E627026B95002D3A5D /* OneSignalCore.framework */; };
 		3CA283AA2B86A30400097465 /* OneSignalCore.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = DE7D17E627026B95002D3A5D /* OneSignalCore.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		3CA6CE0A28E4F19B00CA0585 /* OSUserRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3CA6CE0928E4F19B00CA0585 /* OSUserRequest.swift */; };
+		3CA8B8822BEC2FCB0010ADA1 /* XCTest.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3C7A39D42B7C18EE0082665E /* XCTest.framework */; };
+		3CA8B8832BEC2FCB0010ADA1 /* XCTest.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 3C7A39D42B7C18EE0082665E /* XCTest.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		3CC063942B6D6B6B002BB07F /* OneSignalCore.m in Sources */ = {isa = PBXBuildFile; fileRef = 3CC063932B6D6B6B002BB07F /* OneSignalCore.m */; };
 		3CC063A22B6D7A8E002BB07F /* OneSignalCoreMocks.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3CC0639A2B6D7A8C002BB07F /* OneSignalCoreMocks.framework */; };
 		3CC063A72B6D7A8E002BB07F /* OneSignalCoreTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3CC063A62B6D7A8E002BB07F /* OneSignalCoreTests.swift */; };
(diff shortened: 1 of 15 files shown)


nan-li:
Followup PR https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1430 testing user switching is also testing these changes.

Note CodeQL is broken after push to start live activities merged to main, its autobuild action no longer works.

emawby (approved):

PR #1430, opened 2024-05-09
[Tests] Add more complex user switching tests

nan-li:
# Description
## One Line Summary
Add more complex user switching tests involving multiple calls to login / logout / user updates in a short span.

### Motivation

- Adding more testing, and prevent regressions in behavior. 
- We have seen bugs come up when changes have been made that changed expected behavior.

### Scope
Unit tests

# Testing
## Unit testing
Added new test class `SwitchUserTests`
1. `testIdentifyUserSuccessfully_thenLogin_sendsCorrectTags` - existing test migrated from other file
2. `testAnonUser_thenIdentifyUserWithConflict_sendsCorrectUpdatesAndFetchesUser` - Starts up with creating the anon user, then a login with conflict happens.
3. `testAnonUser_thenIdentifyUserWithConflict_thenLogout_sendsCorrectUpdatesWithNoFetch` - Starts up with creating the anon user, then a login with conflict happens, then a logout happens.
4. `testAnonUser_thenIdentifyUserWithConflict_thenLogin_sendsCorrectUpdatesAndFetchesUser` - Starts up with creating the anon user, then a login with conflict happens, then a login to another user happens.

## Manual testing
None, app runs.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1430)


iOS_SDK/OneSignalSDK/OneSignalCoreMocks/Extensions/NSDictionary+UnitTests.swift
```diff
@@ -1,13 +1,62 @@
...
+     This method goes one level deep into dictionaries
+     */
+    private func contains(key: String, value: Any) -> Bool {
         guard let dictVal = self[key] else {
             return false
         }
-
-        return equals(dictVal, value)
+        if let value = value as? [String: Any],
+        let dictVal = dictVal as? NSDictionary {
```
jkasten2, line 10: Nit: This `if let` formating seems very odd, is this meet standard Swift linting?

iOS_SDK/OneSignalSDK/OneSignalUserTests/SwitchUserTests.swift
```diff
@@ -0,0 +1,463 @@
+import XCTest
+import OneSignalCore
+import OneSignalCoreMocks
+import OneSignalUserMocks
+@testable import OneSignalUser
+
+final class SwitchUserTests: XCTestCase {
```
jkasten2, line 7: Nit: Might be good to state these are more integration style tests.

jkasten2 (approved):

iOS_SDK/OneSignalSDK/OneSignalCoreMocks/Extensions/NSDictionary+UnitTests.swift
nan-li: That's my bad, I did not align the `let`s.

Ideally this looks most correct to me:
```swift
if let value = value as? [String: Any],
   let dictVal = dictVal as? NSDictionary 
{
    return dictVal.contains(value)
} else {
    return equals(dictVal, value)
}
```

But running the swift auto-linter `swiftlint --fix`, the opening `{` does move to next line:
```swift
if let value = value as? [String: Any],
   let dictVal = dictVal as? NSDictionary {
    return dictVal.contains(value)
} else {
    return equals(dictVal, value)
}
```

Thoughts? There's a lot of these patterns and we can decide a better format we want to use.

nan-li: @jkasten2 That's my bad, I did not align the `let`s, and the swiftlint auto-linter did not fix it.

I think this looks most clear to me:
```swift
if let value = value as? [String: Any],
   let dictVal = dictVal as? NSDictionary 
{
    return dictVal.contains(value)
} else {
    return equals(dictVal, value)
}
```

But running the swift auto-linter `swiftlint --fix`, it moves the opening `{` up:
```swift
if let value = value as? [String: Any],
   let dictVal = dictVal as? NSDictionary {
    return dictVal.contains(value)
} else {
    return equals(dictVal, value)
}
```

Thoughts? There's a lot of these type of `if let` or `guard let` in the codebase, and we can decide a better format we want to use.

jkasten2: `swiftlint --fix` looks fine, I don't have any strong feels with that standard.

Original code really throw me off as; it was a combination of the where the 2nd `let` lined up with the `if` and not knowing you could use a `,` that allows multiple variables in the `if`.

iOS_SDK/OneSignalSDK/OneSignalUserTests/SwitchUserTests.swift
nan-li: Renamed class to `SwitchUserIntegrationTests`

iOS_SDK/OneSignalSDK/OneSignalCoreMocks/Extensions/NSDictionary+UnitTests.swift
nan-li: Sounds good to me. Fixed up the original commit.

jkasten2 (approved):

PR #1451, opened 2024-06-17
Fix clearing notifications when the application becomes active instead of when it enters foreground

emawby:
# Description
## One Line Summary
Fixes a bug where opening the notification center while in the app would clear notifications

## Details
The application active state is different than the application foreground state. The app resigns and becomes active if the app is interrupted by the notification center even though it remains in the foreground. We previously were dismissing notifications in didBecomeActive so when the notification center was opened while the app was foregrounded we would dismiss notifications.

There also seems to be an apple bug where the resign and become active triggers would fire twice as you are swiping down the notification center. This means that we were dismissing the notifications as you open the notification center which made the problem even worse.

I refactored where we handle notifications related tasks in response to application lifecycle events. The `OSNotificationsManager` now registers itself to listen to the lifecycle and handles changes on the foreground event.
I also moved checking if the app is using UIScenes to `OneSignalCore`.

### Motivation
bug fix

### Scope
notifications and application lifecycle

# Testing
## Unit testing
Created a new unit testing target for OneSignalNotifications and added it to the UnitTestApp target.

- Added 3 simple tests related to listening to foregrounding the app

## Manual testing

- Tested backgrounding and foregrounding with and without notifications with badges
- Tested changing the permission value in the background and then foregrounding
- Tested resigning and regaining active using the notification center while the the app was foregrounded

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1451)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalCore/Source/OSBundleUtils.h
+++ b/iOS_SDK/OneSignalSDK/OneSignalCore/Source/OSBundleUtils.h
@@ -1,10 +1,24 @@
-//
-//  OSBundleUtils.h
-//  OneSignal
-//
-//  Created by Elliot Mawby on 6/20/24.
-//  Copyright © 2024 Hiptic. All rights reserved.
-//
+/*
+ Modified MIT License
+ Copyright 2024 OneSignal
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+ 1. The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+ 2. All copies of substantial portions of the Software may only be used in connection
+ with services provided by OneSignal.
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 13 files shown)


iOS_SDK/OneSignalSDK/OneSignalNotificationsTests/OneSignalNotificationsTests.swift
```diff
@@ -0,0 +1,84 @@
...
+    func testDontclearBadgesWhenAppBecomesActive() throws {
+        // NotificationManager Start to register lifecycle listener
+        OSNotificationsManager.start()
+        // Set badge count > 0
+        UIApplication.shared.applicationIconBadgeNumber = 1
+        // Then resign active
+        OneSignalCoreMocks.resignActive()
+        // App becomes active the app
+        OneSignalCoreMocks.becomeActive()
+        // Ensure that badge count == 0
```
nan-li, line 51: nit, this comment should say "count == 1"?

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OSBundleUtils.h
```diff
@@ -0,0 +1,11 @@
+//
+//  OSBundleUtils.h
+//  OneSignal
```
nan-li, line 3: The new files created can clean up the headers?

nan-li (changes requested):

nan-li (approved):

Issue #1453, opened 2024-06-19
[Bug] Crash when logging in-app message impression


kajensen:
### What happened?

I'm seeing reports of a subset of my users crashing when logging a OneSignal in-app message impression.
<img width="579" alt="Screenshot 2024-06-19 at 10 59 47 AM" src="https://github.com/OneSignal/OneSignal-iOS-SDK/assets/6947273/2206a848-fd45-4b17-ad8b-54be6ca591b9">

This is a re-opening of #937

### Steps to reproduce?

```Markdown
I haven't reproduced myself but the crash logs indicate it happens when an in-app message is shown. Early indications show that it appears to be limited to iOS 17.5.1 devices.
```

### What did you expect to happen?

No crash

### OneSignal iOS SDK version

5.1.6

### iOS version

17

### Specific iOS version

```Markdown
iOS 17.5.1
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kajensen:
Update: also seeing crashes on 17.4.1

emawby:
@kajensen Thank you for reporting we will investigate

kunalzigma:
this should be resolved immediately, we are fed up of all these crashes. Already lost many users in another app due to crashes and cannot afford to lose more in this one where it is happening.
You might have seen other users removing onesignal from their apps on these sdk issues. You are already losing business due to this.
Dont force us to do the same. 
Or please provide alternatives to Onesignal. 
we cannot afford more crashes of such kind. 
instead of gaining business, we are losing after adding OneSignal. Never faced any such issue with any 3rd party SDK that too paid in past 13 years of development. This is too much and unbearable.
Cant you provide crash free experience for atleast a year or so? Is that really difficult? 
please say it to my face, leave onesignal. Dont harass me. 
This is torture and harassment. 
Instead of focussing on new features, you should be have made it bug-free.

edsonbittencourt:
I'm having problems migrating my app from iOS 17 to iOS 18 and Mac Sequoia, where the scenario is as follows: if the OneSignal framework is installed in the app, I can't run it in Xcode 16, as the app generates multiple errors/crashes.

By completely removing the OneSignal package, everything works normally, and I can build/run my app perfectly.

The problem is that I can't use Xcode 15.4 anymore unless I roll back my OS. Therefore, I can't send app updates.

I don't know if it's related to the problem, but it's quite complicated to maintain the use of the OneSignal framework.

[...]

PR #1457, opened 2024-06-26
[Bug] Make null values safe for IAM requests

nan-li:
# Description
## One Line Summary
Make null values safe for IAM payload such as subscription ID (most common) or other data required in the payload is `nil`.

## Details
* If subscription ID (most common) or other properties such as app ID are `nil`, the app crashes when requests for IAM impressions or clicks are created.
* The app will crash with Exception `[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[1]`
* If we omit these properties in the payload, the request will return a 400, which allows us to track these events
* Note that Android already omit the property in the payload or pass an empty string `""` and receives a 400 response.

### How could this happen?
- All the crash reports point to `null` subscription ID, but to get IAMs, we need a subscription ID
- It is possible the user or subscription has been deleted?
- When the Messaging Controller tries to get IAMs from the server but there is no subscription ID, [we use IAMs from the cache](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/f0552ddf17c6881eaa8c14ed187144485da6c8a9/iOS_SDK/OneSignalSDK/OneSignalInAppMessages/Controller/OSMessagingController.m#L243)
- However, in `shouldShowInAppMessage`, [there is an additional check for subscription ID](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/f0552ddf17c6881eaa8c14ed187144485da6c8a9/iOS_SDK/OneSignalSDK/OneSignalInAppMessages/Controller/OSMessagingController.m#L654) before presenting an IAM. No IAM should be shown when subscription ID is `null`.
- Update, see below under Manual Testing for a reproduction scenario.

### Motivation
Fix crashes, reported in #1453 and others. The crash reports all look like they are for player ID.

### Scope
Correctly handle and pass along null values.

# Testing
## Unit testing
Ported over 2 existing IAM-related request tests, and add a few more with null values and check it does not crash

## Manual testing
Physical iPhone 13 on iOS 17.4
- reproduced crashes reported by manually passing `nil` subscription ID to the `OSRequestInAppMessageViewed` and `OSRequestInAppMessageClicked`
- does not crash after these changes, no request are made.

**I was able to recreate this scenario "naturally" with this flow:**
1. SDK has perviously fetched IAMs so there are IAMs in the cache
2. Kill app and then delete the user via REST API
3. Manipulate the SDK code so that Create User requests cannot be sent
4. Open app, and notice that before IAMs are fetched, we actually evaluate IAMs on [onApplicationDidBecomeActive](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/f0552ddf17c6881eaa8c14ed187144485da6c8a9/iOS_SDK/OneSignalSDK/OneSignalInAppMessages/Controller/OSMessagingController.m#L1070), which at this point subscription ID still exists locally so the `shouldShowInAppMessage` check passes
6. Then the SDK fetches user to refresh and finds the user is gone, nullifies the local subscription ID, and enqueues a Create User request (that I prevent from being sent).
7. The IAM displays in a context with null subscription ID.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1457)


jkasten2 (changes requested):
Could we instead still make the request even if we are missing fields? Doing so allows the backend to have data on how many requests are missing what data. Only logging to the device doesn't give us this insight.

iOS_SDK/OneSignalSDK/OneSignalInAppMessagesTests/OneSignalInAppMessagesTests.swift
```diff
@@ -0,0 +1,33 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ */
+
+import XCTest
+
+final class OneSignalInAppMessagesTests: XCTestCase {
```
jkasten2, line 30: Can we remove this file? It doesn't contain any tests itself and I don't see it being used anywhere else.

jkasten2 (changes requested):

jkasten2 (approved):

PR #1464, opened 2024-07-24
[Feat] Use only OneSignal ID for requests

nan-li:
# Description
## One Line Summary
Use only OneSignal ID for requests, no longer using any External ID for requests.

## Details
- Fetch user will only occur with OneSignal ID, no more EUID
- Any user updates will only occur with OneSignal ID, no more EUID
- Remove Transfer Subscription Request except to uncache on upgrade only
  - Still relevant Transfer Subscription request will be converted into an equivalent Create User request
  - Cached Transfer Subscription requests should be rare - in order for this to be cached, it means the Identify User request returned 409 but the immediate Transfer Subscription did not send or resulted in a retry-able error response
- Identify User success and conflict responses result in different behavior now
  - Both will call Create User instead of marking pending requests to use EUID or fetching and transferring the push subscription

### Motivation
Move away from using External ID in request URLs, only using OneSignal ID.

### Scope
Making requests

# Testing
## Unit testing
- No new unit test are added, but existing unit tests are updated

## Manual testing
**Device:** iPhone 13 on iOS `17.5.1`

<details>
  <summary>Identify User with success (external ID applied successfully)</summary>

1. Start with anonymous user in the SDK
2. Add a tag and an email
3. Login to a new user (brand new external ID)
4. The Identify User request is successful
5. Fetch User by OSID happens
6. Local state is correct and contains the tag and email

</details>

<details>
  <summary>Identify User with 409 conflict</summary>

1. Start with anonymous user and login to an existing EUID
2. Receive 409 conflict response
3. If the user is the current user, it will make a CreateUser request with the external ID and push subscription in the payload
4. If the user is now different, it will make a CreateUser request with the external ID only for hydration of the OneSignal ID, for any pending requests for that previous user

</details>

<details>
  <summary>SDK Upgrade: There is a cached Transfer Subscription for the current user</summary>

1. Reproduce this on `main` by login to an existing EUID
2. The Identify User happens and receives 409 conflict response
3. The subsequent Fetch User by EUID and Transfer Subscription are generated and cached, but do not get sent
4. Upgrade SDK to this version
5. During uncaching, the Fetch User is dropped and the Transfer Subscription is converted to a CreateUser containing the external ID and push subscription.
6. User observer fires and everything looks correct.

</details>

<details>
  <summary>SDK Upgrade: There is a cached Transfer Subscription for a previous user</summary>

1. Reproduce this on `main` by login to an existing EUID `userA`
2. The Identify User happens and receives 409 conflict response
3. The subsequent Fetch User by EUID and Transfer Subscription are generated and cached, but do not get sent
4. Now login to another user `userB`, which enqueues a CreateUser request
5. Upgrade SDK to this version
6. During uncaching, the Fetch User is dropped and the Transfer Subscription is dropped, only the CreateUser remains
7. The CreateUser is sent and the state in the SDK seems correct for `userB`.

</details>

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1464)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
@@ -335,6 +335,19 @@ extension OSUserExecutor {
         OneSignalCoreImpl.sharedClient().execute(request) { _ in
             removeFromQueue(request)
 
+            guard let onesignalId = request.identityModelToIdentify.onesignalId else {
+                OneSignalLog.onesignalLog(.LL_ERROR, message: "executeIdentifyUserRequest succeeded but is now missing OneSignal ID!")
+                executePendingRequests()
+                return
+            }
+
+            // Need to hydrate the identity model for current user or past user with pending requests
+            let aliases = [
+                OS_ONESIGNAL_ID: onesignalId,
+                request.aliasLabel: request.aliasId
+            ]
+            request.identityModelToUpdate.hydrate(aliases)
+
             // the anonymous user has been identified, still need to Fetch User as we cleared local data
             // Fetch the user only if its the current user
             if OneSignalUserManagerImpl.sharedInstance.isCurrentUser(request.identityModelToUpdate) {
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
@@ -190,6 +190,20 @@ extension OSUserExecutor {
         executePendingRequests()
     }
 
+    /**
+     This Create User call expects an Identity Model with external ID to hydrate the OneSignal ID
+     */
+    static func createUser(identityModel: OSIdentityModel) {
+        guard identityModel.externalId != nil else {
+            OneSignalLog.onesignalLog(.LL_ERROR, message: "createUser(identityModel) called with missing external ID")
+            return
+        }
+
+        let request = OSRequestCreateUser(identityModel: identityModel, propertiesModel: nil, pushSubscriptionModel: nil, originalPushToken: nil)
+        appendToQueue(request)
+        executePendingRequests()
+    }
+
     static func executeCreateUserRequest(_ request: OSRequestCreateUser) {
         guard !request.sentToClient else {
             return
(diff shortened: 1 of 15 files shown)


iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
```diff
@@ -190,6 +190,20 @@ extension OSUserExecutor {
         executePendingRequests()
     }
 
+    /**
+     This Create User call expects an Identity Model with external ID to hydrate the OneSignal ID
+     */
+    static func createUser(identityModel: OSIdentityModel) {
+        guard identityModel.externalId != nil else {
+            OneSignalLog.onesignalLog(.LL_ERROR, message: "createUser(identityModel) called with missing external ID")
```
emawby, line 198: If we hit this case what happens? Will we try again in the same session?

nan-li: The purpose of this version of `ceateUser()` is to get an OSID for an EUID. The EUID must exist because the only usage of this method is after an Identify User 409 conflict, in which case we would have the EUID. The guard and error log is more for noticing that we do something wrong such as using this method inappropriately down the line. 

Thinking about this again, it may be clearer to make EUID a parameter instead of it being implicit.

emawby: Ya it should probably require it as a param. Would it retry next session?

nan-li: I edited my response after to talk about retrying. This call would be made if the user has changed since, so this would be for a previous user. It won't retry since no EUID, no way to get its OSID. That means if there are pending requests for this previous user, they will effectively be dropped.

emawby (approved):

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSIdentityModel.swift
```diff
@@ -29,23 +29,7 @@ import Foundation
...
-    case onesignal_id = "onesignal_id"
-    case external_id = "external_id"
-    // swiftlint:enable identifier_name
-}
-
 class OSIdentityModel: OSModel {
-    /**
-     Set either `onesignal_id` or `external_id`, representing the alias that will be used in requests.
-     */
-    var primaryAliasLabel: OSDefaultAlias = .onesignal_id
```
jkasten2, line 44: We will probably want to use this concept for JWT, since we will make requests with `external_id` when it is enabled. I am still ok with removing this code in this PR to if you want to, just making a note in-case this didn't come up.

jkasten2 (approved):

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSIdentityModel.swift
nan-li: @jkasten2 I thought about JWT before removing this. The issue with the `primaryAliasLabel` as it exists here is that every Identity Model needs to manage it. Therefore, when the SDK first detects JWT is on, it needs to go and update every Identity Model when that flag can be centralized.

PR #1465, opened 2024-07-26
[Bug] The user executor needs to uncache before other executors

nan-li:
# Description
## One Line Summary
The User Executor is no longer a static class, and will finish uncaching before other executors uncache.

## Details
**User executor is no longer a static class**
* _Primary Motivation_: After a dispatch queue was added to the user executor, it's uncaching can happen on a background thread. However, when the User Manager starts, the user executor needs to finish uncaching first before other executors begin uncaching. This is because their uncached requests rely on the identity models organized by the user executor. By happening on a background thread, this order is not guaranteed. By making the User Executor an instance instead of a static class, it can uncache in its initializer sans dispatch queue without concurrency dangers.
* Additional motivations include Identity Verification work that allows the user executor to conform to the same listener protocol as other executors.
* All other executors are instances owned by the User Manager.
* Make the User Executor also an instance owned by the User Manager, as it is the only class that calls it.
* Access to user executor is force unwrapped ‼️ because everything is guarded behind the User Manager's `start()` call. This will surface any issues or wrong usages early.
* The reason for the force unwrapping is because the User Manager and the executors have circular dependency currently, unfortunately

### Motivation
Bug Fix

### Scope

# Testing
## Unit testing
- no new tests added but minor updates made

## Manual testing
**Device**: iPhone 13 on iOS 17.5.1
Reproduce original wrong behavior:
1. Start with an anonymous user and turn off internet connection
2. Login to `userA` and add tag for userA
3. Login to `userB` and add tag for userB
4. Note this will enqueue an `IdentifyUser userA`, `CreateUser userB` in the user executor
5. Note this will enqueue two `UpdateUser` in the property executor
6. Swipe away the app, turn on internet, and re-open app
7. The user executor does not finish uncaching before the property executor uncaches
8. Therefore, `UpdateUser userA` is dropped as the property executor cannot connect the Identity model reference (needed to know when the OSID is hydrated)

After changes in this PR:
- Both tag updates are sent to their respective users

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1465)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSIdentityOperationExecutor.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSIdentityOperationExecutor.swift
@@ -39,7 +39,13 @@ class OSIdentityOperationExecutor: OSOperationExecutor {
     private let dispatchQueue = DispatchQueue(label: "OneSignal.OSIdentityOperationExecutor", target: .global())
 
     init() {
-        // Read unfinished deltas from cache, if any...
+        // Read unfinished deltas and requests from cache, if any...
+        uncacheDeltas()
+        uncacheAddAliasRequests()
+        uncacheRemoveAliasRequests()
+    }
+
+    private func uncacheDeltas() {
         if var deltaQueue = OneSignalUserDefaults.initShared().getSavedCodeableData(forKey: OS_IDENTITY_EXECUTOR_DELTA_QUEUE_KEY, defaultValue: []) as? [OSDelta] {
             // Hook each uncached Delta to the model in the store
             for (index, delta) in deltaQueue.enumerated().reversed() {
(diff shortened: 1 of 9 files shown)


emawby (changes requested):
Looks like some of the switch user integration tests failed in the CI. When I ran it locally they didn't fail but other tests are failing every time

nan-li:
> Looks like some of the switch user integration tests failed in the CI. When I ran it locally they didn't fail but other tests are failing every time

A live activity executor test, right?

emawby:
> > Looks like some of the switch user integration tests failed in the CI. When I ran it locally they didn't fail but other tests are failing every time
> 
> A live activity executor test, right?

No it is a notification enters foreground test. It seems like they are probably all unrelated to this PR and are just flaky

emawby (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1465)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jkasten2)

PR #1470, opened 2024-08-14
[Fix] Handle incorrect `404`, delay making requests to new IDs

nan-li:
# Description
## One Line Summary
Handle incorrect 404 responses from the OneSignal's backend by adding a delay after creates, when new subscription IDs or OneSignal IDs can be created on the backend.

## Details
See [Android Implementation here](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2059)

### Backend Issue
OneSignal sometimes returns a `404` on `GET` and `PATCH` requests if you are accessing something immediately after it was created. Normally the OneSignal backend can accept fetch/updates right way, but it all depends on it's server load and replication process. There isn't a guaranteed amount of time a client can wait, so SDKs has to work around this problem.

### SDK's work around strategy
Add a minimum delay after creating records (User or Subscription) before allowing any requests to fetch or update that specific record, based on the `onesignalId` or `subscriptionId`.
   * We picked **3 seconds** as this "cool down" period. This is a best guess, we are balancing SDK responsiveness with the chance of running into this issue, as well as timing for Operation Repo's 5 second flush interval.
   * Also noting that backend team said _"it looks like we can have peaks of up to 300ms for database replication lag. If you added 0.5s delay between each network request, that would likely resolve some of these issues"_

### Motivation
404/410 responses are being handled by recreating the missing record (User or Subscription), however this isn't always the right way to handle these, give the backend caveat noted above. Recreating these records can create side-effects and extra load we want to avoid.

### Scope
Add a delay processing user-based or subscription-based requests.

### Future Considerations
* We want to refactor Operation Repo and Executors relationship down the line, so this PR is minimal to prevent the primary errors. The 3 second delay should be sufficient to address the overwhelming majority of incorrect 404 cases, especially as backend team mentioned a delay of 0.5 seconds should be sufficient.
* As the 404/410 may still happen when the backend is under unusual load (a 3 second delay may not always be enough) we can also retry on 404/410 within a longer but still short window.
* Calls that do not go through the Operation Repo and Executors could have this same logic applied.
    - Live Activities requests, Fetch IAM, and Outcomes requests.
* The new parameter for the delay interval could be read from remote params, so it can be tweaked later.

# Testing
## Manual testing
Physical iPhone 13 on iOS 17.4
- New install
- New install with immediate login to **new** user
- New install with immediate login to **existing** user

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1470)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSNewRecordsState.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSNewRecordsState.swift
@@ -1,7 +1,7 @@
 /*
  Modified MIT License
 
- Copyright 2022 OneSignal
+ Copyright 2024 OneSignal
 
  Permission is hereby granted, free of charge, to any person obtaining a copy
  of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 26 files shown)


iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSNewRecordsState.swift
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public func add(_ key: String, _ overwrite: Bool = false) {
+        lock.withLock {
+            if overwrite || records[key] == nil {
+                records[key] = Date()
+            }
+        }
+    }
+
+    public func canAccess(_ key: String?) -> Bool {
```
emawby, line 57: Should this parameter be optional? Should it return true for a nil string?

iOS_SDK/OneSignalSDK/OneSignalUser/Source/Requests/OSRequestAddAliases.swift
```diff
@@ -37,15 +38,16 @@ class OSRequestAddAliases: OneSignalRequest, OSUserRequest {
     var identityModel: OSIdentityModel
     let aliases: [String: String]
 
-    // requires a `onesignal_id` to send this request
-    func prepareForExecution() -> Bool {
-        if let onesignalId = identityModel.onesignalId, let appId = OneSignalConfigManager.getAppId() {
+    /// requires a `onesignal_id` to send this request
+    func prepareForExecution(newRecordsState: OSNewRecordsState) -> Bool {
```
emawby, line 42: Why are we passing in this state as a parameter to prepareForExecution instead of the class's initializer? Do we expect that state object to change between calls to prepareForExecution?

emawby (changes requested):

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSNewRecordsState.swift
nan-li: There is only 1 place that `nil` can be passed in, it was to simplify this check - if a subscription ID exists, we should check it. Otherwise, the request can be sent.

```swift
guard let appId = OneSignalConfigManager.getAppId(),
      newRecordsState.canAccess(pushSubscriptionModel?.subscriptionId)
else {
    return false
}
```

But I agree it is clearer to handle this at the use site:
```swift
guard let appId = OneSignalConfigManager.getAppId() {
    return false
}

if let subscriptionId = pushSubscriptionModel?.subscriptionId,
   !newRecordsState.canAccess(subscriptionId)
{
    return false
}
```

iOS_SDK/OneSignalSDK/OneSignalUser/Source/Requests/OSRequestAddAliases.swift
nan-li: If the requests are uncached, they won't have the OSNewRecordsState.

emawby (approved):

nan-li:
Badge test started failing this week, but can look at later.

It looks like the runner-image has changed, and the new tests are run on ios 18.1.

**Previous passing runner- image used:**

> Included Software: https://github.com/actions/runner-images/blob/macos-14/20240811.3/images/macos/macos-14-Readme.md
> Image Release: https://github.com/actions/runner-images/releases/tag/macos-14%2F20240811.3

**Failing runner-image used:**

> Included Software: https://github.com/actions/runner-images/blob/macos-14/20240818.2/images/macos/macos-14-Readme.md
> Image Release: https://github.com/actions/runner-images/releases/tag/macos-14%2F20240818.2

PR #1524, opened 2024-11-25
[Fix] Don't evaluate in app messages when paused

nan-li:
# Description
## One Line Summary
Don't evaluate in app messages when paused to improve accuracy once they are unpaused.

### Motivation
A SDK consumer reached out with the following issue:
1. Set up a regular IAM and a duration since last IAM
2. Pause IAMs and open app
3. Now unpause IAMs
4. The first IAM will show, then the duration IAM will show immediately after, ignoring the duration

* This is caused by message evaluation after fetching.
* Since no IAM has shown yet in the paused state, the duration IAM passes the duration check and is queued for display
* Thus, evaluating IAMs while paused will lead to inaccuracies once IAMs are unpaused.
* In this case, duration-since-last IAMs will be evaluated incorrectly and then queue for display once unpaused.
* When IAMs are unpaused, IAM evaluation is re-triggered anyway and will queue the messages for display at that time. There is no need to evaluate while paused.

### Scope
More accurate In App Message evaluation

# Testing
## Unit testing
Will be part of a subsequent task to build IAM tests.

## Manual testing
iPhone simulator 18.1
- Reproduced issue with 2 IAMs (a display on app open, and a 30 second duration since last)
- In app messaging starts out paused
- Then unpause
- First IAM should display then 30 seconds later the next one should display

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1524)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalInAppMessages/Controller/OSMessagingController.m
+++ b/iOS_SDK/OneSignalSDK/OneSignalInAppMessages/Controller/OSMessagingController.m
@@ -683,6 +683,10 @@ static BOOL _isInAppMessagingPaused = false;
  Checks to see if any messages should be shown now
  */
 - (void)evaluateMessages {
+    if (_isInAppMessagingPaused) {
+        [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"Not evaluating in app messages while paused"];
+        return;
+    }
     [OneSignalLog onesignalLog:ONE_S_LL_VERBOSE message:@"Evaluating in app messages"];
     for (OSInAppMessageInternal *message in self.messages) {
         if ([self.triggerController messageMatchesTriggers:message]) {


nan-li:
[Here](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/9d099bc62145b9c3c71de33042f142821446f914/iOS_SDK/OneSignalSDK/OneSignalInAppMessages/Controller/OSMessagingController.m#L187) is where `evaluateMessages` is called when IAMs are unpaused.

jinliu9508:
Logic looks good to fix this issue. Just curious about a scenario when there is a condition to show IAM when either a trigger is satisfied OR duration > 30 seconds. It looks like the trigger won't work when IAM is paused. Is this what we expect?

nan-li:
> Logic looks good to fix this issue. Just curious about a scenario when there is a condition to show IAM when either a trigger is satisfied OR duration > 30 seconds. It looks like the trigger won't work when IAM is paused. Is this what we expect?

I think this is the expected behavior, since we don't show any IAMs while in the paused state. Are you saying maybe it won't work if IAMs are then unpaused?

PR #707, opened 2020-07-06
Add new remote params

Jeasmine:
* Add locationShared remote param
   * Add requiresUserPrivacyConcent remote param
   * Keep public method
   * Add tests for new getters

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/707)

PR #361, opened 2018-04-16
Gdpr privacy

Nightsd01:
• Adds ability to require privacy consent and lets developers add privacy consent/revoke to their apps
• Uses `Onesignal_require_privacy_consent` info.plist key to let developers require consent
• Adds `consentGranted(boolean)` to indicate if the user gives their consent. Can also be used to revoke previously granted consent

PR #1541, opened 2025-02-13
Fix requiring privacy consent blocks confirmed deliveries indefinitely

nan-li:
# Description
## One Line Summary
Setting consent required and then providing consent still blocked confirmed deliveries from being sent - this PR fixes that.

## Details
The consent required value was being written to / read from the shared user defaults but the consent granted value was written to / read from standard user defaults (which is not accessible to the NSE). 
- Looking at the history, the consent granted logic was added in https://github.com/OneSignal/OneSignal-iOS-SDK/pull/361 in 2018 when few things were shared with the NSE. 
- In 2020, the requires consent value started to get cached from remote params in https://github.com/OneSignal/OneSignal-iOS-SDK/pull/707, which saved this to shared user defaults.

This PR also fixes a migration bug (and creates more migrations to fix that and fix the bug in this PR) that arose after modularizing the SDK.

### Motivation
Reported by customers, and we should send confirmed deliveries once consent is granted.

### Scope
- where consent is set and read
- migrations changes including fixing a migration bug after modularization

# Testing
## Unit testing
None added, little infrastructure

## Manual testing
iPhone 13 with iOS 18.1
**Fix the bug in this PR**
1. On `main`, set consent required, and then provide consent
2. Confirmed deliveries fail due to not detecting consent in the extension
3. Upgrade to this PR, and confirmed deliveries succeed, and other requests, such as for the user, succeed too (the migration was necessary to maintain behavior in the app).

**Fix the existing migration bug**
1. Mock an SDK version upgrade but don't open the app. 
2. Let's say the previous version was `2.0.0` and the new version is `5.2.10`.
3. Send a notification
4. The NSE receives notification and calls a migration in `OSOutcomes`
5. The old SDK version of `2.0.0` is read from cache to determine if migration is needed, and the new SDK version is cached. Outcomes is migrated now.
6. Next time the app is launched, the Migration Controller reads the new version from cache `5.2.10` and does not do a migration, even if it is necessary. The migration controller invokes in app messages to migrate.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [x] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1541)

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
@@ -40,6 +40,7 @@ THE SOFTWARE.
 @implementation OSMigrationController
 
 - (void)migrate {
+    [OneSignalCoreImpl migrate];
     [OSOutcomes migrate];
     let oneSignalInAppMessages = NSClassFromString(ONE_SIGNAL_IN_APP_MESSAGES_CLASS_NAME);
     if (oneSignalInAppMessages != nil && [oneSignalInAppMessages respondsToSelector:@selector(migrate)]) {
--- a/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
+++ b/iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
@@ -40,47 +40,11 @@ THE SOFTWARE.
 @implementation OSMigrationController
 
 - (void)migrate {
-    [self migrateToVersion_02_14_00_AndGreater];
+    [OSOutcomes migrate];
     let oneSignalInAppMessages = NSClassFromString(ONE_SIGNAL_IN_APP_MESSAGES_CLASS_NAME);
     if (oneSignalInAppMessages != nil && [oneSignalInAppMessages respondsToSelector:@selector(migrate)]) {
         [oneSignalInAppMessages performSelector:@selector(migrate)];
     }
-    [self saveCurrentSDKVersion];
-}
-
-/**
- * Support renaming of decodable classes for cached data
- */
-- (void)migrateToVersion_02_14_00_AndGreater {
-    let influenceVersion = 21400;
-    let uniqueCacheOutcomeVersion = 21403;
-    long sdkVersion = [OneSignalUserDefaults.initShared getSavedIntegerForKey:OSUD_CACHED_SDK_VERSION defaultValue:0];
-    if (sdkVersion < influenceVersion) {
-        [OneSignalLog onesignalLog:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Migrating OSIndirectNotification from version: %ld", sdkVersion]];
-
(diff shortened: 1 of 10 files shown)


iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
```diff
@@ -40,47 +40,12 @@ + (OSOutcomeEventsCache *)outcomeEventsCache;
...
- */
-- (void)migrateToVersion_02_14_00_AndGreater {
-    let influenceVersion = 21400;
-    let uniqueCacheOutcomeVersion = 21403;
-    long sdkVersion = [OneSignalUserDefaults.initShared getSavedIntegerForKey:OSUD_CACHED_SDK_VERSION defaultValue:0];
-    if (sdkVersion < influenceVersion) {
-        [OneSignalLog onesignalLog:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Migrating OSIndirectNotification from version: %ld", sdkVersion]];
-
-        [NSKeyedUnarchiver setClass:[OSIndirectInfluence class] forClassName:@"OSIndirectNotification"];
-        NSArray<OSIndirectInfluence *> * indirectInfluenceData = [[OSInfluenceDataRepository sharedInfluenceDataRepository] lastNotificationsReceivedData];
```
jkasten2, line 62: I don't see this code or a replacement equivalent in this PR. While the version is pretty old there can definitely be some devices that may go from 2.14 to 5.1.x.

jkasten2 (changes requested):

iOS_SDK/OneSignalSDK/Source/OSMigrationController.m
nan-li: Oh yeah, see commit https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1541/commits/9a1ec13d1aaae242cbc91ccc85a295dc4ecd9094

This method `migrateToVersion_02_14_00_AndGreater` is the exact same method as `migrateToVersion_02_14_00_AndGreater` in `[OSOutcomes migrate]` ([see code](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/a1eea2b98b79544a6af7b31b2a8cbcac1254e7d4/iOS_SDK/OneSignalSDK/OneSignalOutcomes/OSOutcomes.m#L60)), which is called by the NSE after we did the modularization. So let's just use `[OSOutcomes migrate]`.

jkasten2: Ah my bad, you are right.

jkasten2 (approved):

PR #1554, opened 2025-04-28
Synchronize to fix crashes in the user module

nan-li:
# Description
## One Line Summary
Make changes to synchronize and prevent crashes that have been reported.

### Motivation
Address synchronization, race conditions, and user-submitted crash reports.

### Scope
1. Add lock to OSModelStore models dictionary, and confirmed codes paths would not result in deadlock.
2. Make a change to prevent generating an update subscription server call when the SDK nullifies the push subscription ID in 404 scenarios. It will be received from the server on the expected subsequent User Create call.
3. Synchronize the User Manager start up process, and confirm codes paths do not result in deadlock. 
4. To avoid deadlock from changes above, there is one side effect that `purchases` and `session_time` will be dropped if the user manager has not started up by that point, which is a scenario that should not be happening anyway.

# Testing
## Unit testing
Added one test to reproduce a crash via multiple concurrent threads.

## Manual testing
Tested on iPhone 13 running iOS 18.4.1
- The User Manager `start()` method essentially has 3 paths: has cached user, has legacy player to migrate, and new install. Test all code paths to ensure correctness and no deadlocking.
- Manual testing involved triggering concurrent threads doing variety of things.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes
   - [x] Initialization

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1554)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSModel.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSModel.swift
@@ -52,11 +52,11 @@ open class OSModel: NSObject, NSCoding {
     }
 
     // We can add operation name to this... , such as enum of "updated", "deleted", "added"
-    public func set<T>(property: String, newValue: T) {
+    public func set<T>(property: String, newValue: T, preventServerUpdate: Bool = false) {
         let changeArgs = OSModelChangedArgs(model: self, property: property, newValue: newValue)
 
         changeNotifier.fire { modelChangeHandler in
-            modelChangeHandler.onModelUpdated(args: changeArgs, hydrating: self.hydrating)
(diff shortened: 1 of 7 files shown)


nan-li:
The test `testClearBadgesWhenAppEntersForeground` is always failing. Need to fix in another PR, but I think some of the involved APIs might be async is why?

jkasten2 (approved):

PR #1650, opened 2026-03-02
Feat: Add option to disable swizzling

nan-li:
# Description
## One Line Summary
Add option to disable swizzling and add manual integration APIs, when disabled.

### Motivation
Allow disabling swizzling for developers to have more control over notification delegate methods.

### New Manual Integration APIs

If swizzling is enabled but the new forwarding APIs are used, those APIs would no-op and log a warning.

All new manual integration APIs are called on the `OneSignal.Notifications namespace`:

|Objective-C | Swift (NS_SWIFT_NAME) | Purpose|
|-- | -- | --|
|didRegisterForRemoteNotificationsWithDeviceToken: | didRegisterForRemoteNotifications(deviceToken:) | Forward APNs token|
|didFailToRegisterForRemoteNotificationsWithError: | didFailToRegisterForRemoteNotifications(error:) | Forward registration failure|
|didReceiveRemoteNotification:completionHandler: | didReceiveRemoteNotification(userInfo:completionHandler:) | Forward background/silent notification|
|willPresentNotificationWithPayload:completion: | willPresentNotification(payload:completion:) | Forward foreground notification|
|didReceiveNotificationResponse: | didReceiveNotificationResponse(_:) | Forward tap/action
setBadgeCount: | setBadgeCount(_:) | Set badge count|

### How to test
Commit [dev app: setup to test swizzle disabled](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1650/commits/3e444886ad40f7301fb140b42129d9ae18a12882) is setup to allow testing

### Scope
Swizzling disable option

# Testing
## Unit testing
None

## Manual testing
iPhone on iOS 26.3 with sizzling disabled via Info.plist for:
- init, APNS token, notification displays, listeners, click handling, permissions, badge count, NSE, etc

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-iOS-SDK/1650)

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
+++ b/iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
@@ -252,9 +252,6 @@ static NSString *_pushSubscriptionId;
         @selector(onesignalSetApplicationIconBadgeNumber:)
     );
     [OneSignalNotificationsUNUserNotificationCenter setup];
-    
-    [self registerLifecycleObserver];
-    
 }
 #pragma clang diagnostic pop
 
@@ -444,8 +441,7 @@ static NSString *_pushSubscriptionId;
     return true;
 }
 
-+ (void)didRegisterForRemoteNotifications:(UIApplication *)app
-                              deviceToken:(NSData *)inDeviceToken {
++ (void)didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)inDeviceToken {
     let parsedDeviceToken = [NSString hexStringFromData:inDeviceToken];
 
     [OneSignalLog onesignalLog:ONE_S_LL_INFO message: [NSString stringWithFormat:@"Device Registered with Apple: %@", parsedDeviceToken]];
@@ -465,7 +461,7 @@ static NSString *_pushSubscriptionId;
     [self sendPushTokenToDelegate];
 }
 
-+ (void)handleDidFailRegisterForRemoteNotification:(NSError*)err {
++ (void)didFailToRegisterForRemoteNotificationsWithError:(NSError*)err {
     OSNotificationsManager.waitingForApnsResponse = false;
     
     if (err.code == 3000) {
@@ -651,9 +647,10 @@ static NSString *_lastnonActiveMessageId;
     return nil;
 }
 
-+ (void)handleWillPresentNotificationInForegroundWithPayload:(NSDictionary *)payload withCompletion:(OSNotificationDisplayResponse)completion {
++ (void)willPresentNotificationWithPayload:(NSDictionary *)payload completion:(OSNotificationDisplayResponse)completion {
     // check to make sure the app is in focus and it's a OneSignal notification
-    if (![OneSignalCoreHelper isOneSignalPayload:payload]
+    if ([OSPrivacyConsentController shouldLogMissingPrivacyConsentErrorWithMethodName:nil]
+        || ![OneSignalCoreHelper isOneSignalPayload:payload]
         || UIApplication.sharedApplication.applicationState == UIApplicationStateBackground) {
         completion([OSNotification new]);
         return;
@@ -662,10 +659,16 @@ static NSString *_lastnonActiveMessageId;
 
     OSDisplayableNotification *osNotification = [OSDisplayableNotification parseWithApns:payload];
     if ([osNotification additionalData][ONESIGNAL_IAM_PREVIEW]) {
+        [self notificationReceived:payload wasOpened:NO];
         completion(nil);
         return;
     }
-    [self handleWillShowInForegroundForNotification:osNotification completion:completion];
+
+    OSNotificationDisplayResponse wrappedCompletion = ^(OSNotification *notification) {
+        [self notificationReceived:payload wasOpened:NO];
+        completion(notification);
+    };
+    [self handleWillShowInForegroundForNotification:osNotification completion:wrappedCompletion];
 }
 
 + (void)handleWillShowInForegroundForNotification:(OSDisplayableNotification *)notification completion:(OSNotificationDisplayResponse)completion {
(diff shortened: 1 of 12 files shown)


iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
```diff
@@ -252,9 +262,6 @@ + (void)start {
         @selector(onesignalSetApplicationIconBadgeNumber:)
     );
     [OneSignalNotificationsUNUserNotificationCenter setup];
-    
-    [self registerLifecycleObserver];
```
abdulraqeeb33, line 256: Shouldn't we continue to registerLifecycleObserver?

iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
```diff
@@ -651,9 +673,19 @@ + (NSString*)checkForProcessedDups:(NSDictionary*)customDict lastMessageId:(NSSt
...
+        completion([OSNotification new]);
+        return;
+    }
+    [self processWillPresentNotificationWithPayload:payload completion:completion];
+}
+
++ (void)processWillPresentNotificationWithPayload:(NSDictionary *)payload completion:(OSNotificationDisplayResponse)completion {
     // check to make sure the app is in focus and it's a OneSignal notification
-    if (![OneSignalCoreHelper isOneSignalPayload:payload]
+    if ([OSPrivacyConsentController shouldLogMissingPrivacyConsentErrorWithMethodName:nil]
```
abdulraqeeb33, line 686: is there a way we added this new check now?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -784,8 +786,6 @@ + (void)load {
     [[OSMigrationController new] migrate];
 //    sessionLaunchTime = [NSDate date];
     // TODO: sessionLaunchTime used to always be set in load
-    
-    [OSDialogInstanceManager setSharedOSDialogInstance:[OneSignalDialogController sharedInstance]];
```
abdulraqeeb33, line 788: why are we getting rid of this?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -797,12 +797,6 @@ - (void)onesignalSetApplicationIconBadgeNumber:(NSInteger)badge {
     [self onesignalSetApplicationIconBadgeNumber:badge];
 }
 
-+(BOOL) shouldDisableBasedOnProcessArguments {
```
abdulraqeeb33, line 800: did we have something to enable/disable swizzling earlier?

iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
```diff
@@ -161,6 +161,16 @@ +(void)setDelegate:(id<OneSignalNotificationsDelegate>)delegate {
 static UIBackgroundTaskIdentifier _mediaBackgroundTask;
 static BOOL _disableBadgeClearing = NO;
 
++ (BOOL)isSwizzlingDisabled {
+    static BOOL _cached = NO;
+    static BOOL _disabled = NO;
+    if (!_cached) {
```
abdulraqeeb33: how is this protected from multiple threads accessing this method?

iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
nan-li: Yes, this `start` method is just setting up swizzling so it is not called in a non-swizzled implementation. Therefore `[OSNotificationsManager registerLifecycleObserver]` moved to line 753 of OneSignal.m so that it is registered for both swizzled and non-swizzled implementations.
I will rename this method to be more clear as `start` is ambiguous.

iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
nan-li: If you look at `UNUserNotificationCenter+OneSignalNotifications.m` in line 279 for the method 
```
onesignalUserNotificationCenter:(UNUserNotificationCenter *)center willPresentNotification:(UNNotification *)notification withCompletionHandler:(void (^)(UNNotificationPresentationOptions options))completionHandler
```
this method is called by swizzling. In it, we check for `shouldLogMissingPrivacyConsentErrorWithMethodName` and `isOneSignalPayload` and return early if needed. If it doesn't return early there, it then calls `processWillPresentNotificationWithPayload` this method here. However, `processWillPresentNotificationWithPayload` is also called by the manual integration API so we need to check for the same guards. It's because the early return differs when swizzling is used so we cannot remove the early return in `onesignalUserNotificationCenter:(UNUserNotificationCenter *)center willPresentNotification:(UNNotification *)notification ...`

iOS_SDK/OneSignalSDK/Source/OneSignal.m
nan-li: Yeah, I know its confusing since we have an existing `[self shouldDisableBasedOnProcessArguments]` that seems like it is a swizzling guard, but that is actually never used and dead code, so the early return would never actually happen before. But the code you are referring to, it moved up to line 752 since swizzling returns early in line 757 now. We need to do all the things that should be done in both swizzled and non-swizzled implementations, those moved up before the return.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
nan-li: yeah this is dead code, we never documented `DISABLE_ONESIGNAL_SWIZZLING` and the SDK wouldn't have actually worked had someone tried to do that.

iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
nan-li: Good point, will address, but two threads could both see _cached == NO and both compute the value. In practice it's harmless since the plist value is immutable and both threads would write the same result, but it's not good practice

iOS_SDK/OneSignalSDK/OneSignalNotifications/OSNotificationsManager.m
```diff
@@ -161,6 +161,15 @@ +(void)setDelegate:(id<OneSignalNotificationsDelegate>)delegate {
 static UIBackgroundTaskIdentifier _mediaBackgroundTask;
 static BOOL _disableBadgeClearing = NO;
 
++ (BOOL)isSwizzlingDisabled {
+    static BOOL _disabled = NO;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
```
abdulraqeeb33, line 167: 💯

abdulraqeeb33 (approved):

PR #1669, opened 2026-06-02
fix: Preserve user identity across iOS prewarm launches 

nan-li:
# Description
## One Line Summary
Defer SDK ops during iOS prewarm (before first unlock) so the cached user isn't overwritten by an orphan; mirror identifiers to `NSFileProtectionNone` storage for NSE/locked-storage reads.

### Motivation
Fixes SDK-4725: when iOS prewarms the app before first unlock, shared App Group `UserDefaults` reads return nil. The SDK then loads the cached user as empty, falls through to `createUser`, mints a new server-side user, and overwrites the prior identity / external_id / tags on disk once UD becomes readable.

### Scope
- New `OSResilientStorage` — file-backed `NSFileProtectionNone` JSON mirror of opaque identifiers (`app_id`, `subscription_id`, `receive_receipts_enabled`, `has_prior_session`).
- `OneSignalConfig.shouldAwaitAppIdAndLogMissingPrivacyConsent` gets a third branch via an `isProtectedDataAvailableProvider` closure the main app injects at `+OneSignal.init`. NSE leaves the provider nil → treats storage as available (NSE uses the mirror directly).
- `OneSignal.m` seeds an atomic-BOOL one-way latch via a UD probe + `UIApplication.isProtectedDataAvailable` tiebreaker, registers a `UIApplicationProtectedDataDidBecomeAvailable` observer (CAS-gated, runs at most once per process) that re-drives `start()` / `sendPushTokenToDelegate` / LA / IAM / `startNewSession:YES` post-unlock.
- `OSModelStore.refresh()` re-reads UserDefaults if `models` is empty so post-unlock `start()` sees disk state.
- `OneSignalIdentifiers.subscriptionId` / `storedAppId` fall back to the mirror when UD reads are empty.
- NSE `isReceiveReceiptsEnabled` falls back to the mirror.
- LA request prepareForExecution paths switched from `_user?.pushSubscriptionModel.subscriptionId` → `OneSignalIdentifiers.subscriptionId` so LA requests built during prewarm still resolve a sub id.
- `handleAppIdChange` nil-guards `prevAppId` (no destructive clear on a fresh install) and extends the clear set to keep UD + mirror in lockstep.

### What does NOT change
- Public SDK API surface.
- Normal unlocked-device launches: the seed reports YES and `+init` runs synchronously as before.
- NSE behavior on unlocked devices is unchanged (UD reads return values, mirror fallback is never consulted).

# Testing
## Unit testing
- `OSResilientStorageTests` — 8 tests on the public API.
- `OSModelStoreRefreshTests` — 4 tests on the post-unlock refresh path.
- `OneSignalIdentifiersFallbackTests` — 6 tests on UD→mirror fallback.
- `OneSignalUserTests.testStartDefersUntilProtectedDataAvailableThenProceeds` — regression test for the `start()` defer-then-proceed contract.

## Manual testing
On a real device (simulators don't reproduce locked-storage reliably):
- Reboot → don't unlock → send a push → NSE fires receive receipt.
- Reboot → wait for prewarm → unlock → OSID unchanged, no new orphan user server-side.
- Normal cold-launch on an unlocked device → `OneSignal.User.addTag(...)` immediately after `initialize` lands.
- Lock + unlock with app running → SDK calls keep working through the lock; no duplicate `session_count` / `fetchUser` on the unlock.
- Upgrade from `main` to this branch → OSID/cached user preserved, no orphan created, data loaded correctly

**To reproduce original behavior on `main`, use physical device:**
1. start a live activity on screen (causes prewarm to happen consistently)
2. then reboot phone but don't unlock yet
3. the app is likely to go into a prewarm
4. Look on dashboard and see new user and subscription from that device

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [x] Confirm Deliveries
   - [ ] Outcomes
   - [x] Sessions
   - [x] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OneSignalUserManagerImpl.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OneSignalUserManagerImpl.swift
@@ -215,6 +215,16 @@ public class OneSignalUserManagerImpl: NSObject, OneSignalUserManager {
 
             OneSignalLog.onesignalLog(.LL_VERBOSE, message: "OneSignalUserManager calling start")
 
+            // The model stores load their in-memory `models` dict once in their initializer.
+            // If the singleton was first touched while protected data was unavailable (iOS app
+            // prewarm), that read returned nil and the dicts are empty. The gate above ensures
+            // `start()` only proceeds when protected data is available, but the stores need to
+            // be refreshed here so Path 1's cache check can see what's on disk.
+            identityModelStore.refresh()
+            propertiesModelStore.refresh()
+            subscriptionModelStore.refresh()
+            pushSubscriptionModelStore.refresh()
+
             OSNotificationsManager.delegate = self
 
             var hasCachedUser = false
@@ -228,6 +238,11 @@ public class OneSignalUserManagerImpl: NSObject, OneSignalUserManager {
                 _user = OSUserInternalImpl(identityModel: identityModel, propertiesModel: propertiesModel, pushSubscriptionModel: pushSubscription)
                 addIdentityModelToRepo(identityModel)
                 OneSignalLog.onesignalLog(.LL_VERBOSE, message: "OneSignalUserManager.start called, loaded the user from cache.")
+
+                // Backfill the mirror so SDK upgraders to v5.5.2 populate it on first normal launch.
+                if let subId = pushSubscription.subscriptionId, !subId.isEmpty {
+                    OSResilientStorage.setString(subId, forKey: OSResilientStorage.keySubscriptionId)
+                }
             }
 
             // TODO: Update the push sub model with any new state from NotificationsManager
@@ -256,6 +271,7 @@ public class OneSignalUserManagerImpl: NSObject, OneSignalUserManager {
                 OneSignalLog.onesignalLog(.LL_DEBUG, message: "OneSignalUserManager: creating user linked to legacy subscription \(legacyPlayerId)")
                 createUserFromLegacyPlayer(legacyPlayerId)
                 OneSignalUserDefaults.initShared().saveString(forKey: OSUD_PUSH_SUBSCRIPTION_ID, withValue: legacyPlayerId)
+                OSResilientStorage.setString(legacyPlayerId, forKey: OSResilientStorage.keySubscriptionId)
                 OneSignalUserDefaults.initStandard().removeValue(forKey: OSUD_LEGACY_PLAYER_ID)
                 OneSignalUserDefaults.initShared().removeValue(forKey: OSUD_LEGACY_PLAYER_ID)
             } else {
@@ -275,6 +291,13 @@ public class OneSignalUserManagerImpl: NSObject, OneSignalUserManager {
                 _user?.update()
             }
             hasCalledStart = true
+            // Mark this app install as having completed `start()` at least once — see
+            // `OSResilientStorage.keyHasPriorSession`. The `snapshot()` below forces the
+            // async write queue to drain before returning, so an OS kill in the
+            // microsecond window can't lose the sentinel and cause the next prewarm
+            // launch to misclassify us as a fresh install.
+            OSResilientStorage.setString("1", forKey: OSResilientStorage.keyHasPriorSession)
+            _ = OSResilientStorage.snapshot()
         }
     }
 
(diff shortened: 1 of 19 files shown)


iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -528,8 +604,7 @@ + (void)handleAppIdChange:(NSString*)appId {
     NSString *prevAppId = OneSignalIdentifiers.storedAppId;
 
     // Handle changes to the app id, this might happen on a developer's device when testing
-    // Will also run the first time OneSignal is initialized
-    if (appId && ![appId isEqualToString:prevAppId]) {
+    if (appId && prevAppId && ![appId isEqualToString:prevAppId]) {
```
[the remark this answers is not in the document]
nan-li: Won't fix, negligible, requires: (a) pre-PR install (b) app-id change in the upgrade release (c) first launch is prewarm

1. The `prevAppId` nil-guard is pre-existing (carried over from v3), not introduced by this PR. The corruption path the bot describes existed before and is unchanged here.
2. Runtime app-id changes aren't really a fully supported scenario in the SDK. Developers who change their OneSignal app-id are expected to do so as part of a fresh install / reinstall path, not via in-place upgrade.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OneSignalUserManagerImpl.swift
```diff
@@ -275,6 +291,13 @@ public class OneSignalUserManagerImpl: NSObject, OneSignalUserManager {
                 _user?.update()
             }
             hasCalledStart = true
+            // Mark this app install as having completed `start()` at least once — see
+            // `OSResilientStorage.keyHasPriorSession`. The `snapshot()` below forces the
+            // async write queue to drain before returning, so an OS kill in the
+            // microsecond window can't lose the sentinel and cause the next prewarm
+            // launch to misclassify us as a fresh install.
+            OSResilientStorage.setString("1", forKey: OSResilientStorage.keyHasPriorSession)
+            _ = OSResilientStorage.snapshot()
```
nan-li, line 300: remove this,

nan-li:
Manual testing can look for logs:
- "deferred: device-protected storage is not yet available"
- any new onesignal IDs or subscription IDs
- requests made "network request" ...
- when the observer fires that protected data is avail: "refresh hydrated \(stored.count) model(s) from UserDefaults"

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -451,12 +452,83 @@ + (void)delayInitializationForPrivacyConsent {
...
+            [OneSignal startNewSession:YES];
+        }];
+
+        OneSignalConfig.isProtectedDataAvailableProvider = ^BOOL {
+            return atomic_load(&gProtectedDataAvailable);
+        };
+
+        BOOL initialState = ComputeInitialStorageReadable();
+        atomic_store(&gProtectedDataAvailable, initialState);
+        atomic_store(&gObserverShouldRecover, !initialState);
```
sherwinski, line 520: It looks like there could be a race with respect to setting up the observer before the latch and provider. In reality it's probably not likely to happen but in theory could the block see `gObserverShouldRecover == NO` if the notification is sent between `addObserverForName` and the two stores?

Claude suggests:  compute `initialState`, store both atomics, set the provider, and register the observer last

nan-li: **Practical likelihood:** for the notification to fire during the few microseconds inside dispatch_once, the device would have to unlock at exactly that instant during app init. Vanishingly rare. If this were to happen, it is not disastrous. The SDK mostly heals in the same session such as on any `OneSignal.User.X` call, only small gaps would be missed such as missed session_count incrementer or refresh user call. Fully self-heals on next launch.

The reorder narrows but doesn't fully eliminate the race. With the new order, there's still a window between the atomic_store and the addObserverForName where the notification could post and be missed entirely.

fadi-george:
I dont see duplciate record but i notice live activity didnt show after restarting device

fadi-george (approved):

sherwinski (approved):

PR #1677, opened 2026-06-15
fix: [SDK-4757] recover startup when initialized before UIApplicationMain

nan-li:
# Description
## One Line Summary
Fix anonymous user creation when OneSignal initializes from SwiftUI `App.init()` before `UIApplicationMain` exists.

### Motivation
OneSignal iOS SDK 5.5.2 could permanently gate startup on clean installs when initialized from a pure SwiftUI `App.init()`. In that lifecycle, `UIApplication.sharedApplication` is nil before `UIApplicationMain`; the previous protected-data readiness check silently treated that as `false`, deferred the user manager/session startup, and could wait forever for a protected-data notification that never fires on an already-unlocked device. The result was no anonymous user or push subscription creation.

### Scope
This change is limited to SDK startup readiness around protected-data availability. It explicitly handles pre-`UIApplicationMain` initialization, distinguishes actual iOS prewarm using `ActivePrewarm`, and adds lifecycle-based recovery once `UIApplication` exists and reports protected data is available. It does not change public APIs.

### Other
The branch history intentionally includes a temporary demo repro commit followed by a restore commit, so reviewers can check out the repro state if needed. The net PR diff only changes `iOS_SDK/OneSignalSDK/Source/OneSignal.m`.

# Testing
## Unit testing
No new unit tests were added for the UIKit/SwiftUI launch-order path. 

## Manual testing
Manually verified on an iOS 26.5 iPhone 17 Pro simulator:
- Reproduced the issue by initializing the demo app from SwiftUI `App.init()`: before the fix, `UIApplication.sharedApplication` was nil, startup stayed gated, and `OSRequestCreateUser` did not fire.
- Verified the fixed SwiftUI-init repro: `ActivePrewarm=0` was treated as readable, `OSRequestCreateUser` returned 201, and an `onesignal_id` was hydrated.
- Verified the normal AppDelegate-based demo path still creates the user normally.
- Verified deferred recovery by preserving `onesignal_identity.json` while removing UserDefaults-backed state, then relaunching and observing the lifecycle recovery path start deferred OneSignal components and send `OSRequestCreateUser`.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [x] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/Source/OneSignal.m
+++ b/iOS_SDK/OneSignalSDK/Source/OneSignal.m
@@ -463,10 +463,24 @@ static BOOL ComputeInitialStorageReadable(void) {
     if (hasPriorSession) { // returning user during prewarm; defer
         return NO;
     }
-    if ([NSThread isMainThread]) {
-        return UIApplication.sharedApplication.isProtectedDataAvailable;
+    if (![NSThread isMainThread]) {
+        return YES; // off-main: can't safely read UIApplication
     }
-    return YES; // off-main: can't safely read UIApplication
+    UIApplication *sharedApp = UIApplication.sharedApplication;
+    if (sharedApp) {
+        return sharedApp.isProtectedDataAvailable;
+    }
+    // sharedApplication is nil: we are running before UIApplicationMain, e.g. a SwiftUI
+    // App.init() or an ObjC +load. (Messaging nil here would silently return NO and leave
+    // the SDK permanently gated, since the recovery notification below may never fire.)
+    // A user-initiated launch implies the device is unlocked, so storage is readable.
+    // Only a prewarm-created process can reach this point while storage may still be
+    // locked; iOS marks those processes with the ActivePrewarm environment variable,
+    // which is cleared after launch completes — but pre-UIApplicationMain is always
+    // before that, so the read is reliable here.
+    BOOL isPrewarm = [NSProcessInfo.processInfo.environment[@"ActivePrewarm"] isEqualToString:@"1"];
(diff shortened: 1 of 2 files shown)


fadi-george (approved):

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -463,54 +463,98 @@ static BOOL ComputeInitialStorageReadable(void) {
...
+                            object:nil
+                             queue:nil
+                        usingBlock:^(NSNotification * _Nonnull note) {
+            recoverIfDeferred();
         }];
 
+        // Lifecycle signals: if init deferred while storage was actually readable,
+        // the unlock notification never posts. Once UIApplicationMain has run,
+        // `sharedApplication` exists and we can consult the real `isProtectedDataAvailable`.
+        for (NSNotificationName name in @[UIApplicationDidFinishLaunchingNotification,
```
abdulraqeeb33, line 546: why do we need to loop thru the notifications?

nan-li: We need to observe both lifecycle notifications with the same recovery block: `didFinishLaunching` is the earliest point after UIApplicationMain where sharedApplication exists, and `didBecomeActive` gives us a second chance if storage wasn’t readable at didFinish. The loop is just to avoid duplicating the same observer body twice.

abdulraqeeb33 (approved):

PR #1678, opened 2026-06-25
fix: restore local tags cleared by a concurrent FetchUser

nan-li:
# Description

## One Line Summary
Restore locally set tags that a concurrent `FetchUser` can wipe, by re-asserting server-confirmed tags when an `UpdateProperties` request succeeds.

### Motivation
`getTags()` could return `{}` or missing recently-added tags, even when the tags exist on the server. `addTags` writes tags optimistically to the local model and sends them via `OSRequestUpdateProperties`. If a `FetchUser` runs concurrently (e.g. a new session after backgrounding/foregrounding) and its response is missing the just-written tags, its `clearUserData()` + hydrate wipes the local tag cache and nothing restores it — so `getTags()` keeps returning only a subset of tags. The write itself succeeds; only the SDK's local mirror ends up wrong.

### Fix
On a successful `OSRequestUpdateProperties` (202) response, merge the tags the server just confirmed (echoed back in the response) into the local properties model. This is a **merge, never a replace**:
- only the keys present in the response are touched, so tags managed elsewhere (e.g. the app's backend via the REST API) are left intact;
- a `""` value is treated as a removal, matching how the server echoes removed tags.

### Scope
Affects only user tags, on `UpdateProperties` success. The `FetchUser` flow and other user properties are unchanged. No public API changes.

### Alternatives considered
- **Restore the full local tag snapshot during `FetchUser` hydration**: rejected — it would re-introduce stale local-only tags and drop tags the app's backend set via the REST API, clobbering backend state.
- **Preserve only the pending tag writes during `FetchUser` hydration**: viable and race-equivalent to the chosen approach, but needs more cross-executor plumbing (inspecting the operation repo + property executor queues). Merge-on-confirm is simpler, self-contained, and only ever re-asserts tags the server just confirmed.

### Trade-offs / concessions / limitations
- This is a best-effort, eventually-consistent, last-writer-wins system. Inherent staleness (e.g. an offline write that lands later and overwrites a newer backend change) is **out of scope** and not solvable client-side. This fix only addresses the SDK destroying its *own* successfully-written tags locally.
- The `UpdateProperties` response echoes only the tags that were sent (confirmed from logs), not the user's full tag set — so this is necessarily a merge, not a full resync.
- **Relies on the response echoing tags:** the merge reads `properties.tags` from the 202 response. If a response omits it, the merge simply no-ops for that round (no restoration); a later clean fetch reconciles.
- **Assumes removals are echoed as `key: ""`:** removed tags come back as a non-empty `tags` object with empty-string values, which the merge treats as deletions. If the backend ever collapsed an all-removals update to `tags: {}`, the `!confirmedTags.isEmpty` guard would skip the merge; a later clean fetch still reconciles.
- **Residual race (accepted):** a stale `FetchUser` applied *after* the write is already confirmed is not covered in that instant; it self-corrects on the next fetch user call.
- **Minor wart (accepted):** removing a tag whose earlier add is still in flight in a *separate* request can briefly re-add it until the removal confirms; self-corrects on next fetch user. This is an inherent problem outside the scope of this PR. If 2 concurrent requests updating the same tag key are sent to the server, the value persisted on server just depends on random chance.

# Testing
## Unit testing
- `testMergeConfirmedTags_mergesAndRemovesWithoutTouchingOtherTags` — merge semantics: updates, `""` removals, additions, and other tags left untouched (never a wholesale replace).
- `testUpdatePropertiesResponse_restoresTagsClearedByConcurrentFetch` — regression: add tags, simulate a concurrent fetch clearing the local cache, then assert the tags are restored once the update confirms.
- `OneSignalUserTests`, `UserExecutorTests`, `UserConcurrencyTests`, and `SwitchUserIntegrationTests` pass.

## Manual testing
Not performed on-device: the race is timing-dependent and not naturally reproducible, so it is covered by the regression test above, which simulates the clobber via the property model's `clearData()` — the same clearing a stale `FetchUser` performs.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs (no public API changes)

## Testing
   - [x] I have personally tested this on my device, or explained why that is not possible (race is timing-dependent and not naturally reproducible; covered by an automated regression test)

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSPropertyOperationExecutor.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSPropertyOperationExecutor.swift
@@ -246,8 +246,7 @@ class OSPropertyOperationExecutor: OSOperationExecutor {
         }
 
         OneSignalCoreImpl.sharedClient().execute(request) { response in
-            // On success, remove request from cache, and we do need to hydrate
-            // TODO: We need to hydrate after all ? What why ?
+            // On success, remove request from cache
             self.dispatchQueue.async {
                 self.updateRequestQueue.removeAll(where: { $0 == request})
                 OneSignalUserDefaults.initShared().saveCodeableData(forKey: OS_PROPERTIES_EXECUTOR_UPDATE_REQUEST_QUEUE_KEY, withValue: self.updateRequestQueue)
@@ -255,6 +254,16 @@ class OSPropertyOperationExecutor: OSOperationExecutor {
                     OSBackgroundTaskManager.endBackgroundTask(backgroundTaskIdentifier)
                 }
             }
+
+            // Re-assert the tags the server just confirmed by merging them back into the local model,
+            // to remedy a concurrent FetchUser whose response is missing the just-written tags
+            if OneSignalUserManagerImpl.sharedInstance.isCurrentUser(request.identityModel),
+               let properties = response?["properties"] as? [String: Any],
+               let confirmedTags = properties["tags"] as? [String: String],
+               !confirmedTags.isEmpty {
+                OneSignalUserManagerImpl.sharedInstance._user?.propertiesModel.mergeConfirmedTags(confirmedTags)
+            }
+
             if let onesignalId = request.identityModel.onesignalId {
                 if let rywToken = response?["ryw_token"] as? String
                 {
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSPropertiesModel.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSPropertiesModel.swift
@@ -151,6 +151,25 @@ class OSPropertiesModel: OSModel {
         self.set(property: "tags", newValue: tagsToSend)
     }
 
+    /**
+     Merges server-confirmed tags into the local model, without enqueuing a new delta.
+     A value of `""` removes the tag.
+
+     Used to re-assert tags in order to remedy a concurrent FetchUser whose response
+     may be missing the just-written tags.
+     */
+    func mergeConfirmedTags(_ serverTags: [String: String]) {
+        tagsLock.withLock {
+            for (key, value) in serverTags {
+                if value.isEmpty {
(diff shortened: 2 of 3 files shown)


nan-li:
@claude review

nan-li:
ran multi-model-review, which flagged intentional gaps so concessions and limitations addressed in PR description

iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSPropertyOperationExecutor.swift
```diff
@@ -246,15 +246,24 @@ class OSPropertyOperationExecutor: OSOperationExecutor {
         }
 
         OneSignalCoreImpl.sharedClient().execute(request) { response in
-            // On success, remove request from cache, and we do need to hydrate
-            // TODO: We need to hydrate after all ? What why ?
```
abdulraqeeb33, line 250: you removed this TODO? did we have anything todo around this before deleting it?

abdulraqeeb33 (approved):

iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSPropertyOperationExecutor.swift
nan-li: I'm unsure what the _"We need to hydrate after all"_ reasoning was... that was written in the early days of creating v5. Then I added _"What why ?"_  almost 3 years ago in this commit (https://github.com/OneSignal/OneSignal-iOS-SDK/commit/0fda0b1c3faf3f342aee4424620a73c370781ad9) since I didn't know what this TODO was for. I would say that syncing the tags back in is a form of hydration and tags are the most important property we should be synced with the backend on.

PR #1680, opened 2026-06-30
fix: replace deprecated NSKeyedArchiver/NSKeyedUnarchiver APIs

### LINKED ISSUE #919: OneSignalUserDefaults uses NSKeyedUnarchiver method flagged as unsafe
**Description:**

The class `OneSignalUserDefaults` relies on a deprecated method of `NSKeyedUnarchiver`, [`+ unarchiveObjectWithData:`](https://developer.apple.com/documentation/foundation/nskeyedunarchiver/1413894-unarchiveobjectwithdata?language=objc), which has been deemed unsafe. This was called out by submitting our application code, including the OneSignal iOS SDK, to a scanning service. For reference, the report includes a callout to [CVE-676](http://cwe.mitre.org/data/definitions/676.html) with respect to this case.

Apple's recommendation is to use [`+unarchivedObjectOfClass:fromData:error:`](https://developer.apple.com/documentation/foundation/nskeyedunarchiver/2962884-unarchivedobjectofclass?language=objc), introduced in iOS 11.0. The practicality of this change is not currently clear without further review of the code and some experimentation.

**Environment**
1. SDK version 3.2.1
2. Added to project using Carthage
3. Xcode 12
4. App targeting iOS 12.0 and newer

**Steps to Reproduce Issue:**
N/A

**Anything else:**
I can work on devising a way to use `+unarchivedObjectOfClass:fromData:error:` when running on iOS 11.0 or newer, but it would help to know if there are any design requirements of `OneSignalUserDefaults` that should be kept in mind.

garvitkaushik-123:
## Summary

- Replaces deprecated `+[NSKeyedUnarchiver unarchiveObjectWithData:]` with instance-based `NSKeyedUnarchiver` using `initForReadingFromData:error:` and `decodeTopLevelObjectAndReturnError:`
- Replaces deprecated `+[NSKeyedArchiver archivedDataWithRootObject:]` with `archivedDataWithRootObject:requiringSecureCoding:NO error:`
- Both modern APIs are available since iOS 11.0, which matches the SDK's minimum deployment target

The deprecated methods are flagged as unsafe by security scanners (CWE-676). Setting `requiresSecureCoding = NO` maintains backward compatibility with existing model classes that implement `NSCoding`.

Resolves #919

## Test plan

- [ ] Verify existing unit tests pass (archiving/unarchiving behavior is unchanged)
- [ ] Verify migration flows (IAM redisplay cache, class name migrations) work correctly
- [ ] Confirm no deprecation warnings from `NSKeyedArchiver`/`NSKeyedUnarchiver` in build output

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalUserDefaults.m
+++ b/iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalUserDefaults.m
@@ -152,14 +152,21 @@
 }
 
 - (id _Nullable)getSavedCodeableDataForKey:(NSString * _Nonnull)key defaultValue:(id _Nullable)value {
-    if ([self keyExists:key])
-        return [NSKeyedUnarchiver unarchiveObjectWithData:[self.userDefaults objectForKey:key]];
-    
+    if ([self keyExists:key]) {
+        NSData *data = [self.userDefaults objectForKey:key];
+        NSKeyedUnarchiver *unarchiver = [[NSKeyedUnarchiver alloc] initForReadingFromData:data error:nil];
+        unarchiver.requiresSecureCoding = NO;
+        id result = [unarchiver decodeTopLevelObjectAndReturnError:nil];
+        [unarchiver finishDecoding];
+        return result;
+    }
+
     return value;
 }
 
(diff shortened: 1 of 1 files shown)


nan-li:
@claude review

PR #1681, opened 2026-06-30
fix: [SDK-4814] Live Activities Confirmed Receipts exceeding Delivered count

nan-li:
# Description
## One Line Summary
Stop Live Activities from reporting a receive receipt more than once per device, which inflated the dashboard's Confirmed Receipts count above Delivered.

## Details

Reproduced this manually by starting a live activity on the device, opening the app, swiping it away, reopening the app… It looks like ActivityKit re-emits content updates on app relaunches even when the content of the live activity does not change. The SDK had been treating each update as a separate message and was not deduplicating.

### Motivation
For Live Activities, the dashboard's Confirmed Receipts count could exceed Delivered (SDK-4814). A receive receipt for a given `notificationId` was reported to OneSignal more than once per device because:

- The receipt was attributed to the Activity's *current* content snapshot (`activity.content.state`) instead of the content actually delivered in that update, so a burst of content updates could mis-attribute the `notificationId`.
- On success the receipt request was forgotten with no record that it had already been sent, so a later content update — or ActivityKit re-emitting the active activity's content on app relaunch — re-sent it.

### Scope
- Receive receipts are attributed to the iterated `content.state` of each update rather than the live `activity.content` snapshot.
- The receive-receipts request cache is now the single source of truth for dedup: sent receipts are kept (`shouldForgetWhenSuccessful = false`) as markers, so both in-session re-emits and post-relaunch re-emits for the same `notificationId` are suppressed.
- The cache TTL is reduced from 30 to 3 days, bounding how long sent markers and unsent retries are retained.
- No change to Live Activity start/update token handling, notification display, or open behavior.

# Testing
## Unit testing
- Added `testReceiveReceiptsNotResentForSameNotificationIdAfterRelaunch`, a regression test that sends a receipt through one executor, then has a second executor (simulating an app relaunch sharing the same persisted cache) re-enqueue the same `notificationId` and verifies it is not re-sent.
- Updated `testReceiveReceiptsWithSuccessfulRequest` and `testReceiveReceiptsRequestNotExecutedWithSameNotificationId` to assert the sent receipt is retained in the cache as a dedup marker (`requestSuccessful == true`) rather than removed.

## Manual testing
On-device, with a Live Activity on the Lock Screen, swiped the app away to terminate it and reopened it — the path where ActivityKit re-emits the active activity's current content on launch. Before the fix this sent a duplicate Confirmed Receipt for the same `notificationId`; after the fix the duplicate no longer reproduces.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [x] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalLiveActivities/Source/Executors/OSLiveActivitiesExecutor.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalLiveActivities/Source/Executors/OSLiveActivitiesExecutor.swift
@@ -115,11 +115,11 @@ class StartRequestCache: RequestCache {
 }
 
 class ReceiveReceiptsRequestCache: RequestCache {
-    // Keep receive receipts requests for up to 30 days.
-    static let OneMonthInSeconds = TimeInterval(60 * 60 * 24 * 30)
+    // Only pending receipts live here; de-duplication is handled by the executor's confirmedReceiptIds.
+    static let OneWeekInSeconds = TimeInterval(60 * 60 * 24 * 7)
 
     init() {
-        super.init(cacheKey: OS_LIVE_ACTIVITIES_EXECUTOR_RECEIVE_RECEIPTS_KEY, ttl: ReceiveReceiptsRequestCache.OneMonthInSeconds)
+        super.init(cacheKey: OS_LIVE_ACTIVITIES_EXECUTOR_RECEIVE_RECEIPTS_KEY, ttl: ReceiveReceiptsRequestCache.OneWeekInSeconds)
     }
 }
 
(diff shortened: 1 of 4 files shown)


nan-li:
Ran multi-model-review with GPT5.5, Opus 4.8, Codex

fadi-george (approved):

abdulraqeeb33 (approved):

fadi-george (approved):

PR #1687, opened 2026-07-20
fix: [SDK-4874] prevent stale UpdateSubscription leaving users Never Subscribed

nan-li:
# Description
## One Line Summary
Prevent a stale pre-permission UpdateSubscription (`notification_types: -19`) from overwriting a later permission grant and leaving users stuck Never Subscribed.

### Motivation
Rapid permission prompt acceptance can queue an UpdateSubscription while the push subscription is still in the “prompted but never answered” state (`enabled: false`, `notification_types: -19`). That request may stay waiting to be sent. When it later flushes with another UpdateSubscription (enabled = true) request, its payload can race with or overwrite the most recent update due to timing of requests accepted by the server, leaving the server subscription disabled / Never Subscribed even though the user accepted. 

Fixes [SDK-4874](https://linear.app/onesignal/issue/SDK-4874/users-stuck-in-never-subscribed-state-due-to-rapid-prompt-acceptance).

### Scope
- `OSRequestUpdateSubscription` rebuilds the PATCH body from the live subscription model in `prepareForExecution` (and drops the unused `subscriptionObject`).
- `OSSubscriptionOperationExecutor` keeps at most one **unsent** UpdateSubscription per subscription model (last wins).
- Adds regression tests in `SubscriptionUpdateRaceTests`.
- Does **not** change public APIs, prompt UX, or skip sending `-19` entirely. Does **not** serialize already in-flight PATCHes (follow-up if needed).

# Testing
## Unit testing
Added `SubscriptionUpdateRaceTests`:
- Refresh in `prepareForExecution` picks up live `enabled` / `notification_types` after grant.
- A pending pre-permission update + post-grant delta coalesces and sends only the subscribed payload.

## Manual testing
Previously reproduced the race with demo app forced to enqueue conflicting UpdateSubscription requests; after the changes in this PR, the requests carry the current subscribed payload. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing
     - Ran `SubscriptionUpdateRaceTests` locally (pass). Full CI pending on this PR.
     - Unit coverage for the race; earlier DevApp verification of live refresh. Full CI / device pass still pending.

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
+++ b/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
@@ -163,6 +163,7 @@
 		3CA6CE0A28E4F19B00CA0585 /* OSUserRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3CA6CE0928E4F19B00CA0585 /* OSUserRequest.swift */; };
 		3CA8B8822BEC2FCB0010ADA1 /* XCTest.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3C7A39D42B7C18EE0082665E /* XCTest.framework */; };
 		3CA8B8832BEC2FCB0010ADA1 /* XCTest.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 3C7A39D42B7C18EE0082665E /* XCTest.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		3CA93BC4300AEFFA000724B3 /* SubscriptionUpdateRaceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3CA93BC3300AEFFA000724B3 /* SubscriptionUpdateRaceTests.swift */; };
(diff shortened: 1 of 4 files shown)


abdulraqeeb33 (approved):

fadi-george:
Core approach looks good. Two concerns:

refreshParametersFromLiveModel() omits device_model, net_type, and test_type, so those updates are no longer sent.
Retryable and in-flight requests can bypass coalescing and still send stale data.
Also, fixed test sleeps may be flaky in CI.

nan-li:
> Core approach looks good. Two concerns:
> refreshParametersFromLiveModel() omits device_model, net_type, and test_type, so those updates are no longer sent. 

"No longer sent" is not quite accurate, this is existing and it is not sent currently. And for subscription updates, those things don't really need updating, they are not really used nor change after an app is already installed.

> Retryable and in-flight requests can bypass coalescing and still send stale data. Also, fixed test sleeps may be flaky in CI.

Called out in PR that it does not serialize already in-flight PATCHes (follow-up if needed). Omitted as out of scope for this PR, as it is possible but very hard to hit, in favor of a larger planned refactor of operation repo / executors.

fadi-george (approved):

PR #1693, opened 2026-07-23
refactor: [SDK-4911] rebuild OperationRepo unmatched delta queue on flush

nan-li:
## Summary
- Refactors `OSOperationRepo.flushDeltaQueue` to rebuild `deltaQueue` as the list of unmatched deltas instead of removing matched entries while iterating.
- Same keep-until-executor-available semantics; clearer than `for-in` + `remove(at:)` on the same array (Swift `Array` value semantics make the old loop safe, not undefined behavior).

## Test plan
- [x] `OSOperationRepoFlushTests`: matched deltas dispatch and clear the repo queue; unmatched stay queued in order.
- [x] Deterministic interleaved matched/unmatched case asserts dispatched deltas and retained queue order.
- [x] Existing `UserConcurrencyTests.testOperationRepoFlushingConcurrency` still relevant (Full plan); it does not assert this routing behavior.


fadi-george (approved):
maybe add tests?

nan-li:
Added tests

Issue #1588, opened 2025-08-31
[Bug]: Crash in v5.2.14: OneSignal.OSOperationRepo - OSSubscriptionModel.encode(with:) + 298 (OSSubscriptionModel.swift:298)


ethanwa:
### What happened?

Got this crash appearing in Crashlytics:

```
Crashed: OneSignal.OSOperationRepo
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x00000003f531c9f8

Crashed: OneSignal.OSOperationRepo
0  libobjc.A.dylib                0x7b98 objc_retain + 16
1  libobjc.A.dylib                0x7b98 objc_retain_x0 + 16
2  libswiftCore.dylib             0x3fb018 objcRetainAndReturn(void*) + 32
3  OneSignalUser                  0x32d1c OSSubscriptionModel.encode(with:) + 298 (OSSubscriptionModel.swift:298)
4  OneSignalUser                  0x331ec @objc OSSubscriptionModel.encode(with:) + 30792 (<compiler-generated>:30792)
5  Foundation                     0x5f670 _encodeObject + 1176
6  OneSignalOSCore                0x113a4 OSDelta.encode(with:) + 62 (OSDelta.swift:62)
7  OneSignalOSCore                0x114b8 @objc OSDelta.encode(with:) + 25532 (<compiler-generated>:25532)
8  Foundation                     0x5f670 _encodeObject + 1176
9  Foundation                     0x5f018 -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 460
10 Foundation                     0x293a8 -[NSArray(NSArray) encodeWithCoder:] + 584
11 Foundation                     0x5f670 _encodeObject + 1176
12 Foundation                     0x2c39dc +[NSKeyedArchiver archivedDataWithRootObject:] + 104
13 OneSignalCore                  0xd434 -[OneSignalUserDefaults saveCodeableDataForKey:withValue:] + 162 (OneSignalUserDefaults.m:162)
14 OneSignalOSCore                0x4fd4 closure #1 in OSOperationRepo.enqueueDelta(_:flush:) + 2748
15 OneSignalOSCore                0x5064 thunk for @escaping @callee_guaranteed @Sendable () -> () + 2892
16 libdispatch.dylib              0x1aac _dispatch_call_block_and_release + 32
17 libdispatch.dylib              0x1b584 _dispatch_client_callout + 16
[...]

PR #1694, opened 2026-07-24
fix: [SDK-4899] lock OSSubscriptionModel state to stop encode crashes

nan-li:
# Description
## One Line Summary
Lock-protect `OSSubscriptionModel` stored state so concurrent encode can no longer crash.

### Motivation
Fixes production `EXC_BAD_ACCESS` crashes in `OSSubscriptionModel.encode(with:)` when the model is archived on a background queue (operation repo delta queue / subscription executor request caches) while another thread mutates its String-typed properties. Same crash as [GitHub #1588](https://github.com/OneSignal/OneSignal-iOS-SDK/issues/1588).

### Scope
- `OSSubscriptionModel` only: stored properties move into a value-type `State` guarded by one `NSLock`, held only for raw reads/writes (never across change events, observers, or UserDefaults).
- `encode(with:)`, `jsonRepresentation()`, and `currentPushSubscriptionState` read one consistent snapshot.
- Existing setter side effects are preserved.

Thanks [@chunhuidu-dev](https://github.com/chunhuidu-dev) for [#1685](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1685) that informed this fix.

# Testing
## Unit testing
Added `SubscriptionModelConcurrencyTests`:
- Deterministic snapshot test: pause mid-encode, mutate the model, assert the archive is consistent
- Concurrent encode + mutation hammer
- `notificationTypes` pin-to-`-2` while disabled without firing a change event, such as when already opted out and something tries to write a non-negative-two value (e.g. a permission refresh with 15). Disabled is supposed to supersede the real notification type. While disabled, don’t let a stray write of 15 fire a change.

## Manual testing
Not personally reproduced on device (customer crash; not reliably reproducible). Covered by the unit tests above.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
+++ b/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
@@ -164,6 +164,7 @@
 		3CA8B8822BEC2FCB0010ADA1 /* XCTest.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3C7A39D42B7C18EE0082665E /* XCTest.framework */; };
 		3CA8B8832BEC2FCB0010ADA1 /* XCTest.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 3C7A39D42B7C18EE0082665E /* XCTest.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		3CA93BC4300AEFFA000724B3 /* SubscriptionUpdateRaceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3CA93BC3300AEFFA000724B3 /* SubscriptionUpdateRaceTests.swift */; };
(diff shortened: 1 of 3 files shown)


nan-li:
@claude review once

fadi-george (approved):
The change looks safe, but the original loop was not undefined behavior due to Swift Array value semantics. Could we frame this as a readability refactor rather than a crash fix and update the PR description accordingly?
Could we also add a deterministic test with interleaved matched and unmatched deltas, verifying dispatched deltas and the retained queue order? The existing concurrency test has no assertions for this behavior.

nan-li:
> The change looks safe, but the original loop was not undefined behavior due to Swift Array value semantics. Could we frame this as a readability refactor rather than a crash fix and update the PR description accordingly? Could we also add a deterministic test with interleaved matched and unmatched deltas, verifying dispatched deltas and the retained queue order? The existing concurrency test has no assertions for this behavior.

I think this comment is for #1693? Done in that PR.

PR #1695, opened 2026-07-30
fix: [SDK-4818] resolve TOCTOU race on the current user

nan-li:
# Description
## One Line Summary
Fix TOCTOU race where an `isCurrentUser` check then a separate `_user` mutation can apply data to the wrong user after a concurrent login/logout.

### Motivation
Executor callbacks checked whether a request belonged to the current user, then mutated `_user` in a separate step. A concurrent `login()`/`logout()` between those steps could clear or hydrate the wrong user's data.

### Scope
- Adds lock-guarded `_user` access and `currentUser(matching:)` so the check and use share one read.
- Routes fetch/hydrate, logout-on-missing, and tags-merge paths through that accessor.
- Removes `isCurrentUser` so "is this the current user?" has a single source of truth (`_user`).
- No public API changes.

# Testing
## Unit testing
- TOCTOU accessor tests: matching returns the captured user after a concurrent switch; nil when no longer current or no user.
- Existing fetch-user regression tests updated (stale fetch must not clear the new user; current-user fetch still clears before hydrate).

## Manual testing
Not run here — unit coverage exercises the race window; please run `OneSignalUserTests` locally.

# Affected code checklist
   - [ ] Notifications
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
@@ -120,7 +120,7 @@ class OSUserExecutor {
             // Translate the last request into a Create User request, if the current user is the same
             if let request = transferSubscriptionRequestQueue.last,
                let userInstance = OneSignalUserManagerImpl.sharedInstance._user,
-               OneSignalUserManagerImpl.sharedInstance.isCurrentUser(request.aliasId) {
+               userInstance.identityModel.externalId == request.aliasId {
                 createUser(userInstance)
             }
         }
@@ -251,7 +251,7 @@ extension OSUserExecutor {
 
                 // If this user already exists and we logged into an external_id, fetch the user data
                 // Fetch the user only if its the current user and non-anonymous
-                if OneSignalUserManagerImpl.sharedInstance.isCurrentUser(request.identityModel),
+                if OneSignalUserManagerImpl.sharedInstance.currentUser(matching: request.identityModel.modelId) != nil,
                    let identity = request.parameters?["identity"] as? [String: String],
                    let onesignalId = request.identityModel.onesignalId,
                    identity[OS_EXTERNAL_ID] != nil {
@@ -320,7 +320,7 @@ extension OSUserExecutor {
                 request.identityModel.hydrate(identityObject)
 
                 // Fetch this user's data if it is the current user
-                guard OneSignalUserManagerImpl.sharedInstance.isCurrentUser(request.identityModel)
+                guard OneSignalUserManagerImpl.sharedInstance.currentUser(matching: request.identityModel.modelId) != nil
                 else {
                     self.executePendingRequests()
                     return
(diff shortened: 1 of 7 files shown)


iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
```diff
@@ -120,7 +120,7 @@ class OSUserExecutor {
             // Translate the last request into a Create User request, if the current user is the same
             if let request = transferSubscriptionRequestQueue.last,
                let userInstance = OneSignalUserManagerImpl.sharedInstance._user,
```
sherwinski, line 122: This still till reads `_user` directly and matches on `externalId` instead of going through `currentUser(matching:)`, which I assume we'd want to make consistent.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OneSignalUserManagerImpl.swift
```diff
@@ -413,27 +422,27 @@ public class OneSignalUserManagerImpl: NSObject, OneSignalUserManager {
...
+     land between the check and the use and apply one user's data to another.
+     */
+    func currentUser(matching modelId: String) -> OSUserInternal? {
+        guard let user = _user, user.identityModel.modelId == modelId else {
+            return nil
         }
-
-        return userInstance.identityModel.externalId == externalId
+        return user
     }
```
sherwinski: Agent review flagged this but I'm not sure how realistic of an edge case this is. I'd leave it up to you to decide:

> **`OneSignalUserManagerImpl.swift:425-435` — the doc comment promises more than the accessor delivers.**
> > "a concurrent `login()`/`logout()` cannot land between the check and the use and apply one user's data to another"
> That holds for the returned instance's *own* `identityModel`/`propertiesModel`. It does not hold for the global stores that the same guarded blocks mutate. In `OSUserExecutor.swift:452-477`, once the guard passes:
> - `clearUserData(user)` → `subscriptionModelStore.clearModelsFromStore()` (`OneSignalUserManagerImpl.swift:448`) wipes whatever user is current *now*.
> - `parseFetchUserResponse` → `subscriptionModelStore.add(...)` (`OSUserExecutor.swift:565`) writes the stale user's email/SMS subscriptions into the current user's store.
> 
> So a login landing microseconds after the guard still gets its email/SMS subscription store cleared and repopulated from another user's response. (`pushSubscriptionModel` is device-scoped, so that one is fine.) The window is much narrower than before, which is a real improvement — but the comment reads as if the class of bug is closed. Either soften it to "the returned instance's models" or note the residual on `clearUserData`.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSUserExecutor.swift
nan-li: The cached request only has an external ID, so we can't use `currentUser(matching modelId)`, this is the only place that matches on external ID and is a one time migration so reading _user is ok. `currentUser` checks _user under the hood.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OneSignalUserManagerImpl.swift
nan-li: Fair point — I'll scope the comment to the returned instance's identity and properties models. The one thing that isn't scoped is the email/SMS subscription store clear, but login() already clears that store itself, so in practice the stale clear lands on an already-empty store. And the repopulate half is covered: parseFetchUserResponse re-checks currentUser(matching:) before writing email/SMS. Not worth restructuring the store for, so just fixing the comment.

sherwinski (approved):

PR #1702, opened 2026-08-05
feat: [SDK-4976] add Swift adapters for KMP logger

fadi-george:
# Description
## One Line Summary
Add Swift host adapters for the shared KMP logger ports.

### Motivation
The KMP logger owns encoding, batching, and uploads but requires iOS implementations for console logging, platform metadata, protobuf transport, and durable crash storage.

### Scope
- Implements `ILogger`, `ILoggerPlatformProvider`, `ILogHttpSender`, and `ILogFileStore`.
- Persists crash payloads synchronously using atomic rename and `fsync`.
- Adds KMP byte-array conversion and adapter unit coverage.
- Updates the KMP submodule to [OneSignal-KMP-SDK#14](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/14).
- Lifecycle wiring and native crash handlers remain out of scope.

# Testing
## Unit testing
- `OSLoggerAdaptersTests`: 5 tests passed covering logging, metadata, HTTP headers/body, durable file writes, and cleanup.
- `./build_kmp_xcframework.sh`

## Manual testing
- Built `OneSignalFramework` in Release for a generic iOS device.
- A force-kill demo test requires the follow-up lifecycle wiring ticket, so it was not run here.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


nan-li (dismissed):
My main concern is the module placement. It lives in source, which is the OneSignal framework umbrella, which doesn’t feel right. It is one giant file with no documentation in it and unclear hardcodes.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+                OneSignalLog.onesignalLog(
+                    .LL_WARN,
+                    message: "OSLogFileStore delete failed: \(error.localizedDescription)"
+                )
+            }
+            completionHandler(nil)
+        }
+    }
+
+    func deleteUnrecognizedEntries(
```
abdulraqeeb33: this is probably not needed, i had written it in Android SDK as we might have older files of Otel that we cant process and send in the new logger.

abdulraqeeb33 (commented):
Reviewed this against the Android counterparts (`AndroidLogger`, `LoggerPlatformProviderAdapter`, `OneSignalLogHttpSender`, `FileLogStore` in `onesignal/core/.../logger/android/`) and the KMP port contracts in `OneSignal-KMP-SDK/kmp/src/commonMain/kotlin/com/onesignal/logger/`. The port coverage is complete and the durable-write path (temp + `fsync` + atomic rename + directory `fsync`) is actually stronger than Android's. Most of what follows is about the seams between Swift and the shared Kotlin pipeline.

Nothing here is live: the diff only touches the adapter file, its tests, the pbxproj, and the submodule pointer, and nothing constructs these classes yet (lifecycle wiring is scoped out, as the description says). So every runtime finding below is latent. That said, two of them fail specifically on the crash path, which is the hardest place to debug after the fact, so I'd rather fix them here than carry them into the wiring PR.

### Blockers

1. **`appState` touches `UIApplication.shared` off the main thread.** `LogFieldsPerEvent.getAttributes()` reads `appState`, `processUptime`, and `currentThreadName` for every record, and `LogTelemetryRemoteImpl` runs on `CoroutineScope(SupervisorJob() + Dispatchers.Default)`. It's also read from the crashing thread via `LogCrashReporter.saveCrash`'s `runBlocking`. Details inline.
2. **`lazy var installId` is not thread-safe** and is read concurrently from `Dispatchers.Default` and the crash path.
3. **The `KotlinByteArray` ↔ `Data` conversions box one Kotlin object per byte.** This is the one I'd fix first — it's free on Android and roughly 200k allocations per 100 KB payload here.

### Parity gaps with Android

- `Content-Type` precedence is inverted relative to `OneSignalLogHttpSender`.
- Non-2xx responses discard the body and report `message: nil`; Android drains `errorStream` and surfaces it behind the diagnostics flag. There's also no iOS equivalent of Android's `isDiagnosticsEnabled` closure, so `OSLogHttpSender` can never log anything.
- `enabledFeatureFlags` and `isExporterLoggingEnabled` are hardcoded where Android has real values.
- `remoteLogLevel` re-reads `OSRemoteParamController.shared().remoteParams` per event from background threads while that dictionary is replaced when remote params land. Worth confirming whether that property is synchronized, and snapshotting if not.
- `processUptime` subtracts two wall-clock values (`Date()` minus `sysctl` `p_starttime`), so an NTP correction or user clock change skews it. Android uses monotonic `SystemClock.uptimeMillis()`; `ProcessInfo.processInfo.systemUptime` is the closer analogue.

### On module placement

Agreeing with @nan-li here, and I think Android already has the fix. The reason this landed in `Source/` is the `import OneSignalUser` — `OneSignalUserManagerImpl` and `OneSignalIdentifiers` sit above Core, so Core can't reach them. Android hit the identical wall and solved it by injection rather than placement: `OtelIdResolver` reads IDs straight out of SharedPreferences instead of depending on the user module, and app state arrives as a closure (`getIsInForeground`) in `OtelPlatformProviderConfig`.

iOS has both ingredients already. `OSUD_PUSH_SUBSCRIPTION_ID` is cached in `OneSignalUserDefaults`, and `OneSignalLifecycleObserver` already observes `UIApplicationDidBecomeActiveNotification` plus the `UIScene` variants. Resolve the OneSignal ID from a cached default (or inject it as a closure the umbrella sets at init) and this whole file can move into `OneSignalCore` next to `OneSignalLog` and `OSRemoteParamController`. That also resolves the main-thread bug for free.

On the "one giant file with no documentation" point: I'd split it into one file per port, matching Android. Each of the four Android classes carries a KDoc block explaining the non-obvious constraints — why `save` is deliberately blocking, why `.otlp` denotes ownership, why the temp-then-rename. None of that rationale survived the port, and it's exactly the context the next reader will need.

### Build configuration

The `FRAMEWORK_SEARCH_PATHS` entries point at `.../XCFrameworks/release/...` in the **Debug** configurations too, so Debug builds link the Release KMP slice — was that intentional? Separately, that path is a locally-built artifact that only exists after `./build_kmp_xcframework.sh` runs, so a clean clone or a CI job that doesn't invoke the script won't resolve `import OneSignalKMP`. Worth confirming CI runs it before this merges.

### Smaller things

- `currentThreadName` returns `""` when `pthread_getname_np` succeeds with an empty name; it should fall through to `"unknown"`.
- `writeDurably` with empty `data` hits the nil `baseAddress` early return and writes a zero-byte `.otlp` that later uploads as an empty body.
- No data-protection class is set on the crash files. The Caches default is `CompleteUntilFirstUserAuthentication`, so a crash before first unlock can't persist — probably acceptable, but worth being deliberate about.
- Prefixes mix `OSLog*` and `OSLogger*`, and `OSLog*` also reads like Apple's own `OSLog`. Worth settling on one.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+    var onesignalId: String? {
+        OneSignalUserManagerImpl.sharedInstance.onesignalId
+    }
+
+    var pushSubscriptionId: String? {
+        OneSignalUserManagerImpl.sharedInstance.pushSubscriptionId
+    }
+
+    var appState: String {
+        switch UIApplication.shared.applicationState {
```
abdulraqeeb33: **Blocker: main-thread violation, including on the crash path.**

`UIApplication.shared.applicationState` is main-thread-only, but this is read from at least two background contexts:

- `LogFieldsPerEvent.getAttributes()` reads `appState` for **every** record, and `LogTelemetryRemoteImpl` runs on `CoroutineScope(SupervisorJob() + Dispatchers.Default)` — a background worker on Kotlin/Native.
- On the crash path, `LogCrashReporter.saveCrash` does `runBlocking { crashTelemetry.emit(record) }` on the crashing thread, which is whatever thread threw.

This will trip the Main Thread Checker and can hang, and doing it inside a crash handler is the worst place for that.

Android avoids it entirely — `OtelPlatformProviderConfig` carries a `getIsInForeground: (() -> Boolean?)?` closure so the provider never touches a UI API. iOS has the same source of truth available: `OneSignalLifecycleObserver` already observes `UIApplicationDidBecomeActiveNotification`/`DidEnterBackground` and the `UIScene` equivalents. Cache a flag there and read it atomically.

Same shape of concern for `onesignalId` and `pushSubscriptionId` just above, which hit `OneSignalUserManagerImpl.sharedInstance` per event from background threads.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+
+    private lazy var installId: String = {
+        let defaults = OneSignalUserDefaults.initShared()
+        if let saved = defaults.getSavedString(forKey: Self.installIdKey, defaultValue: nil) {
+            return saved
+        }
+        let generated = UUID().uuidString
+        defaults.saveString(forKey: Self.installIdKey, withValue: generated)
+        return generated
+    }()
```
abdulraqeeb33: **Blocker: `lazy` is not thread-safe, and this is read on the crash path.**

Swift's `lazy` has no synchronization. `getInstallId` is called from `LogFieldsTopLevel.getAttributes()` on `Dispatchers.Default` and, concurrently, from the crashing thread inside `LogCrashReporter`'s `runBlocking`. Racing this initializer can double-generate and double-persist the UUID, or fault mid-initialization — a crash inside the crash handler.

Resolve it eagerly at construction, or guard it with an `os_unfair_lock`.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+        }
+        defer { close(descriptor) }
+        guard fsync(descriptor) == 0 else {
+            throw POSIXError(POSIXErrorCode(rawValue: errno) ?? .EIO)
+        }
+    }
+}
+
+final class OSLoggerPlatformProvider: NSObject, ILoggerPlatformProvider {
+    private static let installIdKey = "PREFS_OS_INSTALL_ID"
```
abdulraqeeb33: This key is Android's, copied verbatim — it's `PreferenceOneSignalKeys.PREFS_OS_INSTALL_ID` there. It appears nowhere else in this repo, and iOS convention is an `OSUD_*` constant declared in `OneSignalCommonDefines.h`.

More substantively: iOS has no install-id concept today (no `OneSignal-Install-Id` header anywhere in the repo), so this quietly mints a new persistent device identifier inside a logging adapter and writes it to the app-group-shared defaults via `initShared()`. On Android that identity is owned by `InstallIdService` in core and drives an outgoing header.

Both the introduction of the identifier and the choice of shared-vs-standard defaults feel like decisions that deserve to be made explicitly rather than as a side effect of the logger port.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+    var data: Data {
+        Data((0..<size).map { UInt8(bitPattern: get(index: $0)) })
+    }
+}
+
+private extension Data {
+    var kotlinByteArray: KotlinByteArray {
+        KotlinByteArray(size: Int32(count)) { index in
+            KotlinByte(value: Int8(bitPattern: self[Int(index.int32Value)]))
+        }
```
abdulraqeeb33: **Blocker (performance): one boxed Kotlin object allocated per byte.**

This closure allocates a `KotlinByte` for every byte and receives a boxed `KotlinInt` index for every byte — roughly 200k allocations for a 100 KB crash payload, per file, on every `listReadable` pass. The other direction (`KotlinByteArray.data` on line 441) is n cross-boundary `get(index:)` calls; no boxing, but still O(n) interop.

On Android this is a plain `ByteArray` and costs nothing, so the shared pipeline was designed without this cost in mind. It matters most in `save()`, which runs on the crashing thread with limited time before the process dies.

The KMP module declares `iosX64`/`iosArm64`/`iosSimulatorArm64` but has no `iosMain` source set at all right now. Adding `ByteArray.toNSData()` / `NSData.toByteArray()` there with `usePinned { memcpy(...) }` and exporting them makes both directions a single memcpy. That's a KMP-side change, so worth deciding now whether it belongs in [OneSignal-KMP-SDK#14](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/14) or a follow-up.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+            return saved
+        }
+        let generated = UUID().uuidString
+        defaults.saveString(forKey: Self.installIdKey, withValue: generated)
+        return generated
+    }()
+
+    func getInstallId(completionHandler: @escaping (String?, Error?) -> Void) {
+        completionHandler(installId, nil)
+    }
```
abdulraqeeb33: Worth a comment here: this handler **must** be invoked synchronously.

`LogCrashReporter` bridges to the suspend sink with `runBlocking` on the crashing thread, and this only works because the callback fires inline. If someone later dispatches it to a queue — which looks entirely reasonable for something that touches `NSUserDefaults` — you either deadlock or lose the crash record.

The KMP KDoc states the constraint ("Crash-path callers may block on this; prefer returning a cached value") but nothing on the Swift side does.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+                nil
+            )
+            return
+        }
+
+        var urlRequest = URLRequest(url: url)
+        urlRequest.httpMethod = "POST"
+        urlRequest.httpBody = request.body.data
+        request.headers.forEach { urlRequest.setValue($0.value, forHTTPHeaderField: $0.key) }
+        urlRequest.setValue(request.contentType, forHTTPHeaderField: "Content-Type")
```
abdulraqeeb33: `Content-Type` precedence is inverted relative to Android. `OneSignalLogHttpSender` sets `Content-Type` first and then applies `request.headers`, so a header entry wins there; here headers are applied first and then overwritten by `request.contentType`.

No impact today since the shared code sends them consistently, but it's a silent divergence.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+                completionHandler(
+                    LogHttpResponse(success: false, statusCode: -1, message: "Missing HTTP response"),
+                    nil
+                )
+                return
+            }
+
+            completionHandler(
+                LogHttpResponse(
+                    success: (200...299).contains(response.statusCode),
```
abdulraqeeb33: The response body is discarded (`_` in the closure) and failures report `message: nil`, so a 4xx from the log endpoint is undiagnosable from the shared pipeline.

Android drains `errorStream` — both to surface the body under the diagnostics flag and so the connection closes cleanly — and passes the message through on `LogHttpResponse`. Related: Android's sender takes an `isDiagnosticsEnabled: () -> Boolean` closure wired to `isExporterLoggingEnabled`; there's no iOS equivalent, so `OSLogHttpSender` can never emit any diagnostics at all.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+        }
+        return level != "NONE"
+    }
+
+    var remoteLogLevel: String? {
+        let config = OSRemoteParamController.shared().remoteParams["logging_config"] as? [String: Any]
+        return (config?["log_level"] as? String)?.uppercased()
+    }
+
+    let isExporterLoggingEnabled = false
```
abdulraqeeb33: This and `enabledFeatureFlags = []` on line 339 are the hardcodes I think @nan-li was pointing at. Android drives both from real sources (`isOtelExporterLoggingEnabled` from remote config, `enabledFeatureFlags` from `IFeatureManager`).

A one-line comment each saying iOS has no feature-flag manager or exporter-diagnostics toggle yet would make these read as deliberate placeholders rather than oversights.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    func deleteUnrecognizedEntries(
+        minAgeMillis: Int64,
+        completionHandler: @escaping (KotlinInt?, Error?) -> Void
+    ) {
+        ioQueue.async {
+            var deleted = 0
+            do {
+                for url in try self.fileURLs() where !url.lastPathComponent.hasSuffix(Self.ownedSuffix) {
```
abdulraqeeb33: Following up on @abdulraqeeb33's note — the premise is right, but I'd stop short of deleting this.

On Android this exists because `FileLogStore` shares a crash directory with the legacy OTel `disk-buffering` module and needs to reclaim bare-millis filenames. iOS never shipped that module, and `crashStoragePath` here is a brand-new `caches/onesignal/logger/crashes` that nothing else writes to. `ILogFileStore.deleteUnrecognizedEntries` already defaults to `= 0`, so iOS could simply not override it.

But it isn't dead — it's the only thing reaping orphaned `<id>.otlp.tmp` files. `writeDurably` creates that temp before the atomic rename, and an interrupted write leaves it behind (on the crash path, interruption is the likely case). It doesn't match `.otlp`, so `listReadable` skips it and nothing else removes it.

So either keep the override but narrow it to the `.tmp` suffix with a comment noting iOS has no legacy OTel directory and this only reaps interrupted writes, or drop the override and sweep stale temps inside `save`/`listReadable`. The first is smaller and keeps parity with the `LogCrashUploader` `finally` call site.

Either way `testFileStoreDeletesOnlyOldUnrecognizedEntries` should change — it seeds a file named `legacy-crash`, which is an Android-only scenario.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
```diff
@@ -0,0 +1,451 @@
...
+
+        do {
+            try data.withUnsafeBytes { rawBuffer in
+                guard var pointer = rawBuffer.baseAddress else {
+                    return
+                }
+                var remaining = rawBuffer.count
+                while remaining > 0 {
+                    let count = Darwin.write(descriptor, pointer, remaining)
+                    guard count > 0 else {
```
abdulraqeeb33: Two small edges in `writeDurably`:

If `data` is empty, `baseAddress` is nil and the closure returns early, producing a zero-byte `.otlp` that `listReadable` will later hand to the uploader as an empty body. Probably worth guarding at the top of `save`.

Also, `Darwin.write` can legitimately return `-1` with `EINTR`; `guard count > 0` treats that as fatal and drops the record. A retry on `EINTR` would be cheap insurance on a path that only runs while the process is dying.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Narrowed cleanup to stale `.otlp.tmp` files only. Unrelated files are preserved, and the test now covers interrupted writes.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Injected app state and user identifier providers. OSCore no longer touches `UIApplication.shared` or `OneSignalUser`, and the provider contract requires arbitrary-thread safety.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Removed the lazy install ID. It is now injected as a synchronously available value, so the adapter owns no racing initialization state.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Removed the Android preference key and ID generation entirely. Install identity ownership is now explicit at composition through an injected provider.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Added a pinned-memory `NSData` bridge in OneSignal-KMP-SDK#16 and switched Swift conversion to it. Both directions now use a single native copy.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Documented that completion must run inline and that injected providers must be safe for synchronous crash-handler access.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Aligned precedence with Android by setting the content type first and applying request headers afterward.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: The sender now consumes the response body, returns it on failures, and supports opt-in diagnostics through an injected logger and toggle. Added coverage for the failure path.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Replaced both hardcodes with injected feature-flag and exporter-diagnostics providers.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Kept cleanup but limited it to stale `.otlp.tmp` files left by interrupted writes. The Android-only legacy fixture was removed.

iOS_SDK/OneSignalSDK/Source/OSLoggerAdapters.swift
fadi-george: Added an empty-payload guard and retry on `EINTR` in the write loop.

fadi-george:
Addressed the review feedback in `65e43f05`:

- moved and split the adapters under `OneSignalOSCore/Source/Logging`
- injected crash-path state and configuration, including install ID ownership, instead of reading UI/user/remote state in OSCore
- fixed HTTP parity, temp cleanup, empty writes, `EINTR`, monotonic uptime, thread-name fallback, and file protection
- standardized the Swift types on the `OSLogger*` prefix
- replaced per-byte boxing with the pinned-memory bridge in [OneSignal-KMP-SDK#16](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/16)

The KMP artifact path is intentionally `XCFrameworks/release`: KMP produces one static XCFramework consumed by both host configurations. CI builds it first in `.github/workflows/ci.yml`.

Verified with KMP formatting/tests/XCFramework validation, focused iOS adapter tests, and Debug and Release `OneSignalOSCore` builds.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
```diff
@@ -0,0 +1,188 @@
...
+    var pushSubscriptionId: String? {
+        pushSubscriptionIdProvider()
+    }
+
+    var appState: String {
+        appStateProvider()
+    }
+
+    var processUptime: Int64 {
+        Int64(ProcessInfo.processInfo.systemUptime * 1_000)
```
abdulraqeeb33: **This is measuring the wrong thing, and it's my fault.** I pointed at `ProcessInfo.processInfo.systemUptime` as "the closer analogue" last round and it got substituted directly.

`systemUptime` is time since the *device* booted, not since the process started. On a phone that's been up three days, a freshly launched app now reports `process.uptime` as roughly 259,200,000 ms instead of a few hundred. The previous `sysctl` version was right in intent — it only had a wall-clock skew problem, which is much smaller than this.

Android computes `SystemClock.uptimeMillis() - Process.getStartUptimeMillis()`, so the iOS equivalent needs process start expressed in the same monotonic timebase. Subtracting `kern.boottime` from the `kp_proc.p_starttime` you already had gives you process start in `systemUptime`'s frame; then it's `systemUptime - that`. The `max(0, …)` clamp also disappeared and is worth keeping.

Note that `XCTAssertGreaterThanOrEqual(provider.processUptime, 0)` passes either way, so the test wouldn't catch a regression here.

iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
```diff
@@ -5378,6 +5410,14 @@
 			buildSettings = {
 				APPLICATION_EXTENSION_API_ONLY = YES;
 				BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
+				"FRAMEWORK_SEARCH_PATHS[sdk=iphoneos*]" = (
```
abdulraqeeb33, line 5413: **This will fail the next release build.**

These entries are conditioned on `[sdk=iphoneos*]` and `[sdk=iphonesimulator*]`. `build_all_frameworks.sh` archives every framework a third time for Catalyst:

```
xcodebuild archive -scheme ${BUILD_SCHEME} -destination='generic/platform=macOS,variant=Mac Catalyst' ...
```

and line 51 runs that for `OneSignalOSCore`. The subsequent `-create-xcframework` hard-references all three archive paths, so a missing Catalyst archive fails the step outright. Since the adapter files compile into OSCore and `import OneSignalKMP`, and no conditional matches a Catalyst destination, that archive fails with "no such module 'OneSignalKMP'".

Adding a conditional wouldn't help — there's no slice to point at. Kotlin/Native has no Mac Catalyst target; it isn't in the [supported target list](https://kotlinlang.org/docs/native-target-support.html) and [KT-40442](https://youtrack.jetbrains.com/issue/KT-40442) is still open. Catalyst needs `-macabi` binaries, so neither the device nor simulator slice can substitute.

This is green right now only because `ci.yml` doesn't invoke `build_all_frameworks.sh` — that only runs from `create-release-prs.yml`.

Cheapest fix that changes nothing for existing integrations: keep shipping the Catalyst slice exactly as it is and just leave the logger out of it, guarding the four adapter files with `#if !targetEnvironment(macCatalyst)` and skipping the KMP link for that variant.

I'd avoid the alternative of dropping the Catalyst slice unless someone confirms nobody uses it — every other framework depends on OSCore, so OSCore losing its Catalyst slice effectively ends Catalyst support for the whole SDK, which is a breaking change for existing integrations.

For what it's worth, device and simulator coverage is correct: `iosX64` and `iosSimulatorArm64` fold into the one `ios-arm64_x86_64-simulator` slice, so Intel and Apple Silicon Macs are both handled, and Apple's "Designed for iPad" path runs the `ios-arm64` device slice directly. Catalyst is genuinely the only gap.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerAdapter.swift
```diff
@@ -0,0 +1,50 @@
...
+ THE SOFTWARE.
+ */
+
+import Foundation
+import OneSignalCore
+@_implementationOnly import OneSignalKMP
+
+/// Routes diagnostic messages from the shared logger through the iOS SDK's
+/// existing logging pipeline.
+final class OSLoggerAdapter: ILogger {
```
abdulraqeeb33: All four adapters dropped `: NSObject` this round — this was `final class OSLoggerAdapter: NSObject, ILogger`.

Kotlin/Native exports Kotlin interfaces as Objective-C protocols, and its interop guidance is that Swift types implementing them subclass `NSObject`. It clearly compiles, so this may well be fine, but every test calls the adapters directly from Swift — nothing dispatches from Kotlin into them — so a runtime interop problem wouldn't surface here.

Worth one round-trip through the KMP pipeline to confirm before the wiring PR builds on it. The asymmetry is visible in the test file: `LoggerAdapterListener` is still `NSObject` because `OSLogListener` is an ObjC protocol from OneSignalCore, while `TestLogger: ILogger` isn't.

Related to the debug-variant point in the summary: the Kotlin/Native debug runtime is what would surface this as a clear diagnostic rather than an obscure crash, and Debug builds currently link the release binary.

abdulraqeeb33 (commented):
This is a strong response to the first round — all ten inline comments are addressed, and several went past what I asked for.

Splitting into four documented adapters under `OneSignalOSCore/Source/Logging/` resolved @nan-li's placement concern and the "one giant file" concern in a single move, and I think OSCore is the right home: the KMP framework is built with `isStatic = true`, so it links into OSCore's binary rather than becoming a runtime dependency, and linking a static Kotlin/Native framework into exactly one widely-depended-upon module is what avoids duplicating the Kotlin runtime later. The seven injected closures on `OSLoggerPlatformProvider` are a cleaner seam than the Android design they're modeled on — one mechanism instead of `OtelPlatformProviderConfig` lambdas plus `OtelIdResolver` preference reads. And deleting the install-ID generation outright instead of renaming the key was the right instinct; that decision now gets made explicitly at composition rather than inherited by accident.

The tests improved more than the +73 lines suggest. `testHttpSenderPostsEncodedBytesAndPassesHeaders` now passes a conflicting `Content-Type` through `headers` and asserts the header wins, which actually pins the Android-matching precedence instead of just exercising the happy path. `testFileStoreDeletesOnlyInterruptedTemporaryWrites` asserts the foreign file is *preserved*, which is the half of that contract that's easy to regress. The `EINTR` retry, empty-payload guard, `completeUntilFirstUserAuthentication`, and thread-name fallback are all in. And the bridge in [OneSignal-KMP-SDK#16](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/16) is correct — `NSData.create(bytes:length:)` maps to `dataWithBytes:length:`, which copies, so pinning only for the duration of the block is safe.

Also worth closing out from round one: CI *does* build the framework. `ci.yml` runs `build_kmp_xcframework.sh` before the Xcode build, so the clean-clone concern I raised doesn't apply.

Three things left, only one of which is a code bug.

### 1. Mac Catalyst will break the release build

This is the one with a deadline. `build_all_frameworks.sh` archives every framework for Catalyst unconditionally, including `OneSignalOSCore`, and Kotlin/Native has no Catalyst target ([KT-40442](https://youtrack.jetbrains.com/issue/KT-40442)). Details inline on the search-path block. It's green today only because `ci.yml` doesn't run that script — it fails at the next release.

### 2. `processUptime` regression

The only actual runtime bug in the current code, and it came from my round-one suggestion being applied literally. Details inline. Sorry for the misdirection there.

### 3. Dropping `NSObject` needs verifying

Not necessarily a problem, but currently unknown. Details inline.

### Worth doing while you're here

I flagged the Debug configurations pointing at `XCFrameworks/release/` as though it were an oversight. Having traced the build, it isn't — `verifyOneSignalKMPXCFramework` declares `dependsOn 'assembleOneSignalKMPReleaseXCFramework'` and hardcodes the `XCFrameworks/release` directory, so the release variant is the only artifact anything produces. Pointing Debug at it is currently load-bearing.

The cost is that every Xcode Debug build links an optimized Kotlin/Native binary: breakpoints inside KMP won't bind, KMP frames won't symbolicate usefully, and you lose the debug runtime's interop checks — which is exactly what would turn item 3 into a clear diagnostic instead of an obscure crash. Debugging the wiring PR against a release-only KMP binary is going to be unpleasant.

If you want the debug variant, it's three files: parameterize the verify task over `['debug', 'release']` in `kmp/build.gradle` (the XCFramework plugin already registers `assembleOneSignalKMPDebugXCFramework`), give `build_kmp_xcframework.sh` an optional variant argument, and replace the hardcoded `release` in the pbxproj with a `KMP_XCFRAMEWORK_VARIANT` setting defined per configuration. Since that block repeats across roughly ten configurations — most of the 152 added pbxproj lines — an `.xcconfig` would define it once and shrink the diff a lot. One wrinkle: CI builds the `UnitTestApp` scheme in Debug, so once Debug points at `debug`, CI has to produce that variant too. Worth deciding whether it builds both or just the one it's about to compile against.

### Minor

In `writeDurably`, the `setAttributes` call for the protection key sits between the `defer` and the `do`/`catch` that cleans up the temp file, so if it throws you leak an `.otlp.tmp`. Self-healing via `deleteUnrecognizedEntries` now, but moving it inside the `do` is free.

And carrying forward for the wiring PR rather than this one: the `OSRemoteParamController.remoteParams` thread-safety question didn't go away, it moved. Whoever writes `remoteLogLevelProvider` has to satisfy the "safe to invoke from arbitrary threads" contract in the class doc. Same for `appId`, which still reads `OneSignalIdentifiers.currentAppId` directly. And `osName`/`osVersion` are `let`s evaluated from `UIDevice.current` at construction, so the initializer has a main-thread affinity the doc comment doesn't mention.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
fadi-george: Fixed in `abed34cd`. Process start is now converted to the boot-relative timebase using `kern.boottime`, then subtracted from `systemUptime` with the zero clamp preserved. Added deterministic coverage for the subtraction and clamp.

iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
fadi-george: Fixed in `abed34cd`. All KMP-dependent Swift files are excluded for Mac Catalyst, while the existing OSCore Catalyst slice remains intact. Verified with the same Catalyst archive settings used by `build_all_frameworks.sh`.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerAdapter.swift
fadi-george: Verified without `NSObject` in `abed34cd`. The new test builds crash telemetry and a reporter through `LoggerFactory`, then Kotlin calls the Swift platform provider, file store, and logger end to end. It persists the record and receives both Kotlin-originated log callbacks. Keeping these as pure Swift also avoids exposing the implementation-only KMP protocol through `OneSignalOSCore-Swift.h`.

fadi-george:
Also moved the file-protection attribute write inside the temp-file cleanup scope, so a protection error cannot leave an unmanaged temporary file. The remote-configuration/app-state composition concerns remain scoped to SDK-4977.

nan-li (commented):
no longer blocking

nan-li (dismissed):
-

abdulraqeeb33 (approved):
Approving. All three items from the last round are properly closed out, and the fixes are better than what I asked for in a couple of places.

**`processUptime`** now derives process start in the boot-relative timebase (`kp_proc.p_starttime` minus `kern.boottime`) and subtracts it from `systemUptime`, with the zero clamp restored. Pulling the arithmetic into a static `processUptimeMillis(systemUptime:processStartUptime:)` was a nice touch — it makes the subtraction and the clamp testable without mocking a clock, which is exactly why the old vacuous `>= 0` assertion missed the regression. The `?? systemUptime` fallback degrading to ~0 rather than device uptime is the right failure mode too.

**Catalyst** is handled with the minimum blast radius: five files guarded, the OSCore Catalyst slice still ships, and no KMP symbols are referenced in that variant so the absent search-path conditional stops mattering. Verifying against the same archive settings `build_all_frameworks.sh` uses is the check that counts, since that path isn't exercised by `ci.yml`.

**The `NSObject` question is genuinely answered**, not just asserted. `testKmpPipelineInvokesSwiftAdapters` goes through `LoggerFactory` and `saveNonFatal` so Kotlin dispatches into the Swift platform provider, file store, and logger for real, and asserts both the persisted record and the Kotlin-originated log callbacks. That's the round-trip that was missing. Your point about not exposing the implementation-only KMP protocol through `OneSignalOSCore-Swift.h` is a good independent reason to keep them pure Swift.

Nothing blocking left. Three things for whenever:

The debug XCFramework variant didn't make this commit, and I'd suggest deciding on it explicitly rather than letting it carry — it's the difference between having usable breakpoints in KMP while you build the lifecycle wiring and not. Happy for it to be its own PR.

`TestLogger` picked up `: NSObject` in the test file, which is now unnecessary — it's only ever handed to `OSLoggerHttpSender` and never crosses into Kotlin, and the new pipeline test proves pure Swift conformance works. Purely cosmetic.

One nuance on the uptime fix that I'd leave alone: `systemUptime` excludes time the device spent asleep, while the `boottime`/`p_starttime` difference is wall-clock and includes it. Negligible on iOS and it's the standard idiom, but it does mean the numbers won't line up exactly with Android's `SystemClock.uptimeMillis()`-based value if anyone ever compares them across platforms.

Still carrying forward to the wiring PR, restating so it isn't lost: whoever writes `remoteLogLevelProvider` has to satisfy the "safe to invoke from arbitrary threads" contract in the class doc, since `OSRemoteParamController.remoteParams` is replaced when remote params land — that race moved rather than disappeared. Same for `appId` reading `OneSignalIdentifiers.currentAppId` directly. And the provider needs constructing on the main thread, which the doc comment now says.

Nice work turning this around — the injected-provider design ended up cleaner than the Android original it's ported from.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerAdapter.swift
```diff
@@ -0,0 +1,55 @@
...
+// Kotlin/Native does not produce a Mac Catalyst framework slice.
+#if !targetEnvironment(macCatalyst)
+
+import Foundation
+import OneSignalCore
+@_implementationOnly import OneSignalKMP
+
+/// Routes diagnostic messages from the shared logger through the iOS SDK's
+/// existing logging pipeline.
+final class OSLoggerAdapter: ILogger {
```
nan-li: nit we should have naming consistencies with android so components can be equated more easily.  In Android the concrete class is called AndroidLogger and here OSLoggerAdapter

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/KotlinByteArray+Data.swift
```diff
@@ -0,0 +1,46 @@
...
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ */
+
+// Kotlin/Native does not produce a Mac Catalyst framework slice.
+#if !targetEnvironment(macCatalyst)
```
nan-li, line 29: Not specific to this PR, but so does this mean KMP in general is going to be a problem for Mac Catalyst as we migrate more components into KMP.. something to consider, as there will be a product change if we drop that support

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
```diff
@@ -0,0 +1,245 @@
...
+import UIKit
+
+/// Supplies iOS SDK state and platform metadata to the shared logger.
+///
+/// Mutable SDK state is injected by the composition layer because
+/// `OneSignalUser` depends on `OneSignalOSCore`; importing it here would create
+/// a circular module dependency. Providers must be safe to invoke from arbitrary
+/// threads, including synchronously from a crash handler. Construct this adapter
+/// on the main thread because its static device metadata comes from `UIDevice`.
+final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
```
nan-li, line 44: not a blocker, can be cleaned up later, there's different ways to get these data. Some are injected closures, some are reading direct SDK reads, some computed properties here, not all of which are clear why or justification

nan-li (approved):

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerAdapter.swift
fadi-george: Aligned in `9a286be8`: `IOSLogger`, `OneSignalLogHttpSender`, and `FileLogStore` now match their Android counterparts. I kept `OSLoggerPlatformProvider` because iOS provides values directly, while Android’s `LoggerPlatformProviderAdapter` wraps an existing OTEL provider.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/KotlinByteArray+Data.swift
fadi-george: Yes, this affects any future KMP component used by Catalyst. The current guard preserves existing Catalyst support rather than dropping the SDK slice. I created [SDK-4998](https://linear.app/onesignal/issue/SDK-4998/support-kmp-logger-on-mac-catalyst) to track a supported `macabi` KMP path and CI coverage. Kotlin/Native currently has no official Catalyst target.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
fadi-george: Agreed. The value ownership and access strategy should be normalized when SDK-4977 composes these providers and defines their threading guarantees. I kept that out of this adapter-definition PR.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerAdapter.swift
fadi-george: Agreed, done

PR #1703, opened 2026-08-05
feat: [SDK-4977] wire KMP remote logging lifecycle

fadi-george:
# Description
## One Line Summary
Wire iOS SDK logs and lifecycle events into the shared KMP remote logger behind `SDK_CUSTOM_LOGGING` and remote log-level gates.

### Motivation
The SDK-4976 adapters do not emit telemetry until iOS composes `LoggerFactory`, routes `OneSignalLog` events, and flushes the shared logger during app lifecycle transitions.

### Scope
- Creates and owns KMP remote telemetry after remote parameters load.
- Requires both `sdk_custom_logging` and a non-`NONE` `logging_config.log_level`; the flag-off path keeps existing local logging unchanged.
- Applies remote severity filtering, forwards SDK logs through `LogLoggingHelper`, flushes on app/scene backgrounding, and shuts down on termination/reset.
- Emits startup diagnostics with SDK/KMP versions, active path, feature flag, and crash directory.
- Includes KMP PR [#15](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/15) as a stacked dependency.

# Testing
## Unit testing
- Focused `OSLoggerAdaptersTests` and `OSRemoteLoggingControllerTests` pass on the iOS simulator.
- Coverage verifies dual-gate behavior, severity filtering, routing, background flush, and shutdown.

## Manual testing
- `OneSignalFramework` Release build succeeds for generic iOS.
- No physical-device `/sdk/log` smoke test was performed; rollout remains disabled unless remote config enables both gates.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
+++ b/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
@@ -75,6 +75,11 @@
 		3C14E3B32FAE54C006ED053 /* OSLoggerPlatformProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AC2FAE54C006ED053 /* OSLoggerPlatformProvider.swift */; };
 		3C14E3B42FAE54C006ED053 /* KotlinByteArray+Data.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AD2FAE54C006ED053 /* KotlinByteArray+Data.swift */; };
 		3C14E3B52FAE54C006ED053 /* OSLoggerAdaptersTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AE2FAE54C006ED053 /* OSLoggerAdaptersTests.swift */; };
+		497700000000000000000003 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000001 /* OSRemoteLoggingController.swift */; };
+		497700000000000000000004 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000001 /* OSRemoteLoggingController.swift */; };
+		497700000000000000000005 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000001 /* OSRemoteLoggingController.swift */; };
+		497700000000000000000007 /* OSRemoteLoggingControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000006 /* OSRemoteLoggingControllerTests.swift */; };
+		497700000000000000000009 /* OSRemoteLogger.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000008 /* OSRemoteLogger.swift */; };
 		3C19C6322E919F0C00D6731E /* OSRequestLiveActivityClicked.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C19C6312E919F0C00D6731E /* OSRequestLiveActivityClicked.swift */; };
 		3C23A21B2FCE0A52001D32E3 /* OneSignalIdentifiersFallbackTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C23A21A2FCE0A52001D32E3 /* OneSignalIdentifiersFallbackTests.swift */; };
 		3C23A21D2FCE0A83001D32E3 /* OSModelStoreRefreshTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C23A21C2FCE0A83001D32E3 /* OSModelStoreRefreshTests.swift */; };
@@ -1789,6 +1794,10 @@
 		3C14E3AC2FAE54C006ED053 /* OSLoggerPlatformProvider.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OSLoggerPlatformProvider.swift; sourceTree = "<group>"; };
 		3C14E3AD2FAE54C006ED053 /* KotlinByteArray+Data.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "KotlinByteArray+Data.swift"; sourceTree = "<group>"; };
 		3C14E3AE2FAE54C006ED053 /* OSLoggerAdaptersTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OSLoggerAdaptersTests.swift; sourceTree = "<group>"; };
+		497700000000000000000001 /* OSRemoteLoggingController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OSRemoteLoggingController.swift; sourceTree = "<group>"; };
+		497700000000000000000002 /* OSRemoteLoggingController.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = OSRemoteLoggingController.h; sourceTree = "<group>"; };
+		497700000000000000000006 /* OSRemoteLoggingControllerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OSRemoteLoggingControllerTests.swift; sourceTree = "<group>"; };
+		497700000000000000000008 /* OSRemoteLogger.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OSRemoteLogger.swift; sourceTree = "<group>"; };
 		DEF5CCF12539321A0003E9CC /* UnitTestApp.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = UnitTestApp.app; sourceTree = BUILT_PRODUCTS_DIR; };
 		DEF5CCF32539321A0003E9CC /* AppDelegate.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = AppDelegate.h; sourceTree = "<group>"; };
 		DEF5CCF42539321A0003E9CC /* AppDelegate.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = AppDelegate.m; sourceTree = "<group>"; };
@@ -2285,6 +2294,7 @@
 				3C14E3AA2FAE54C006ED053 /* OneSignalLogHttpSender.swift */,
 				3C14E3A92FAE54C006ED053 /* IOSLogger.swift */,
 				3C14E3AC2FAE54C006ED053 /* OSLoggerPlatformProvider.swift */,
+				497700000000000000000008 /* OSRemoteLogger.swift */,
 			);
 			path = Logging;
 			sourceTree = "<group>";
@@ -2665,6 +2675,7 @@
 				4746E2A62B86B64100D6324C /* LiveActivitiesSwiftTests.swift */,
 				4746E2AA2B8775C400D6324C /* LiveActivitiesObjcTests.m */,
 				3C4319082F4CE9D90075492D /* SessionEndOutcomesRequestTests.swift */,
+				497700000000000000000006 /* OSRemoteLoggingControllerTests.swift */,
 			);
 			path = UnitTests;
 			sourceTree = "<group>";
@@ -2697,6 +2708,8 @@
 				3C448B9B2936ADFD002F96BC /* OSBackgroundTaskHandlerImpl.h */,
 				3C448B9C2936ADFD002F96BC /* OSBackgroundTaskHandlerImpl.m */,
 				DEC08AFF2947D4E900C81DA3 /* OneSignalSwiftInterface.swift */,
+				497700000000000000000002 /* OSRemoteLoggingController.h */,
+				497700000000000000000001 /* OSRemoteLoggingController.swift */,
 				3CC9A6352AFA26E7008F68FD /* PrivacyInfo.xcprivacy */,
 			);
 			path = Source;
@@ -4379,6 +4392,7 @@
 				9124120E1E73342200E41FD7 /* OneSignal.m in Sources */,
 				9124121E1E73342200E41FD7 /* OneSignalJailbreakDetection.m in Sources */,
 				3C14E3A42AFAE54C006ED053 /* OneSignalSwiftInterface.swift in Sources */,
+				497700000000000000000003 /* OSRemoteLoggingController.swift in Sources */,
 				912412471E73369600E41FD7 /* OneSignalHelper.m in Sources */,
 				CA8E19062193C76D009DA223 /* OSInAppMessagingHelpers.m in Sources */,
 				7AAA60682485D0420004FADE /* OSMigrationController.m in Sources */,
@@ -4424,6 +4438,7 @@
 				3C14E3B12FAE54C006ED053 /* OneSignalLogHttpSender.swift in Sources */,
 				3C4F9E4428A4466C009F453A /* OSOperationRepo.swift in Sources */,
 				3C14E3B02FAE54C006ED053 /* IOSLogger.swift in Sources */,
+				497700000000000000000009 /* OSRemoteLogger.swift in Sources */,
 				3C11518B289ADEEB00565C41 /* OSEventProducer.swift in Sources */,
 				3C115165289A259500565C41 /* OneSignalOSCore.docc in Sources */,
 				5BC1DE5E2C90B80E00CA8807 /* OSCondition.swift in Sources */,
@@ -4530,6 +4545,7 @@
 				9124123F1E73342200E41FD7 /* UIApplicationDelegate+OneSignal.m in Sources */,
 				7AECE59F23675F6300537907 /* OSFocusTimeProcessorFactory.m in Sources */,
 				DEC08B012947D4E900C81DA3 /* OneSignalSwiftInterface.swift in Sources */,
+				497700000000000000000004 /* OSRemoteLoggingController.swift in Sources */,
 				CA1A6E7120DC2E73001C41B9 /* OneSignalDialogRequest.m in Sources */,
 				912412331E73342200E41FD7 /* OneSignalTracker.m in Sources */,
 				7A674F1C2360D82E001F9ACD /* OSBaseFocusTimeProcessor.m in Sources */,
@@ -4624,6 +4640,7 @@
 				CA42CAC320D99CB90001F2F2 /* ProvisionalAuthorizationTests.m in Sources */,
 				5B58E4F8237CE7B4009401E0 /* UIDeviceOverrider.m in Sources */,
 				3C4319092F4CE9D90075492D /* SessionEndOutcomesRequestTests.swift in Sources */,
+				497700000000000000000007 /* OSRemoteLoggingControllerTests.swift in Sources */,
 				CA8E19022193C6B0009DA223 /* InAppMessagingIntegrationTests.m in Sources */,
 				CAB4112B20852E4C005A70D1 /* DelayedConsentInitializationParameters.m in Sources */,
 				7AECE59223674A9700537907 /* OSAttributedFocusTimeProcessor.m in Sources */,
@@ -4654,6 +4671,7 @@
 				03CCCC852835F291004BF794 /* UIApplicationDelegateSwizzlingTests.m in Sources */,
 				4529DEEA1FA8360C00CEAB1D /* UIApplicationOverrider.m in Sources */,
 				DEC08B022947D4E900C81DA3 /* OneSignalSwiftInterface.swift in Sources */,
+				497700000000000000000005 /* OSRemoteLoggingController.swift in Sources */,
 				7A93269E25AF4F0300BBEC27 /* OSPendingCallbacks.m in Sources */,
 				7AECE59823674AB700537907 /* OSUnattributedFocusTimeProcessor.m in Sources */,
 				7A5A818224897693002E07C8 /* MigrationTests.m in Sources */,
--- a/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
+++ b/iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
@@ -75,11 +75,11 @@
 		3C14E3B32FAE54C006ED053 /* OSLoggerPlatformProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AC2FAE54C006ED053 /* OSLoggerPlatformProvider.swift */; };
 		3C14E3B42FAE54C006ED053 /* KotlinByteArray+Data.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AD2FAE54C006ED053 /* KotlinByteArray+Data.swift */; };
 		3C14E3B52FAE54C006ED053 /* OSLoggerAdaptersTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AE2FAE54C006ED053 /* OSLoggerAdaptersTests.swift */; };
-		497700000000000000000003 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000001 /* OSRemoteLoggingController.swift */; };
-		497700000000000000000004 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000001 /* OSRemoteLoggingController.swift */; };
-		497700000000000000000005 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000001 /* OSRemoteLoggingController.swift */; };
-		497700000000000000000007 /* OSRemoteLoggingControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000006 /* OSRemoteLoggingControllerTests.swift */; };
-		497700000000000000000009 /* OSRemoteLogger.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000008 /* OSRemoteLogger.swift */; };
+		B96A3B6BA8CC49EE4796D9BF /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A72F938F8A3808AC1FF7F3C /* OSRemoteLoggingController.swift */; };
+		25898119922BDCDA7AF0B9CC /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A72F938F8A3808AC1FF7F3C /* OSRemoteLoggingController.swift */; };
+		9EAF92032D0429FA35E04417 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A72F938F8A3808AC1FF7F3C /* OSRemoteLoggingController.swift */; };
+		ACE2175908241BB46C9F1829 /* OSRemoteLoggingControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 658E6E9E6BC6BBF702BCBD33 /* OSRemoteLoggingControllerTests.swift */; };
+		7732574D325D34CC7C498199 /* OSRemoteLogger.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6972EE491A57C79EFE56D4C8 /* OSRemoteLogger.swift */; };
 		3C19C6322E919F0C00D6731E /* OSRequestLiveActivityClicked.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C19C6312E919F0C00D6731E /* OSRequestLiveActivityClicked.swift */; };
 		3C23A21B2FCE0A52001D32E3 /* OneSignalIdentifiersFallbackTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C23A21A2FCE0A52001D32E3 /* OneSignalIdentifiersFallbackTests.swift */; };
 		3C23A21D2FCE0A83001D32E3 /* OSModelStoreRefreshTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C23A21C2FCE0A83001D32E3 /* OSModelStoreRefreshTests.swift */; };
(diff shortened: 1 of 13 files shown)


iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,271 @@
...
+                + "crash_dir=\(provider.crashStoragePath)"
+        )
+    }
+
+    private func message(from event: OneSignalLogEvent) -> String {
+        let prefix = "\(OSRemoteLoggingConfiguration.levelName(event.level)): "
+        return event.entry.hasPrefix(prefix)
+            ? String(event.entry.dropFirst(prefix.count))
+            : event.entry
+    }
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. OneSignalLogEvent now carries the raw message separately, and the controller forwards it directly. Added WARN coverage to verify the body has no console prefix.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,271 @@
...
+struct OSRemoteLoggingConfiguration {
+    static let featureFlagName = "SDK_CUSTOM_LOGGING"
+    private static let featureFlagKey = "sdk_custom_logging"
+
+    let isFeatureEnabled: Bool
+    let logLevel: String?
+
+    var isRemoteLoggingEnabled: Bool {
+        isFeatureEnabled && logLevel != nil && logLevel != "NONE"
+    }
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. The level is parsed once into a typed threshold, and both enablement and filtering use that value. Invalid levels now keep telemetry disabled.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,271 @@
...
+                exceptionStacktrace: nil,
+                completionHandler: { _ in }
+            )
+        }
+    }
+
+    func forceFlush() {
+        stateQueue.async { [weak self] in
+            self?.telemetry?.forceFlush(completionHandler: { _ in })
+        }
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. Lifecycle flushes now hold an OSBackgroundTaskManager task until the KMP completion callback, and termination flushes before shutdown.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,271 @@
...
+        ]
+        for name in backgroundNames {
+            notificationTokens.append(
+                notificationCenter.addObserver(forName: name, object: nil, queue: nil) { [weak self] _ in
+                    self?.forceFlush()
+                }
+            )
+        }
+        notificationTokens.append(
+            notificationCenter.addObserver(
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. Registration now follows OSBundleUtils.isAppUsingUIScene() and observes either scene or application lifecycle notifications, not both.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,271 @@
...
+    private func logStartupDiagnostic(activePath: String) {
+        let provider = OSLoggerPlatformProvider()
+        OneSignalLog.onesignalLog(
+            .LL_WARN,
+            message: "OneSignal logging initialized: sdk=\(provider.sdkBaseVersion), "
+                + "kmp=\(LoggerBuildInfo.shared.KMP_VERSION), path=\(activePath), "
+                + "\(OSRemoteLoggingConfiguration.featureFlagName)=\(configuration.isFeatureEnabled), "
+                + "crash_dir=\(provider.crashStoragePath)"
+        )
+    }
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. The local-path startup warning was removed. The diagnostic is emitted only when the KMP logger is activated.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,271 @@
...
+            ?? (remoteParams["sdkRemoteFeatureFlags"] as? [String])
+            ?? (remoteParams["feature_flags"] as? [String])
+            ?? []
+        isFeatureEnabled =
+            Self.boolValue(remoteParams[Self.featureFlagKey])
+            ?? Self.boolValue(featureFlags?[Self.featureFlagKey])
+            ?? Self.boolValue(loggingConfig?[Self.featureFlagKey])
+            ?? enabledFlagNames.contains { $0.caseInsensitiveCompare(Self.featureFlagKey) == .orderedSame }
+        logLevel = (loggingConfig?["log_level"] as? String)?.uppercased()
+    }
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. Parsing is now limited to the backend contract used by Android: sdk_remote_feature_flags plus logging_config.log_level, with centralized keys.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -735,6 +737,7 @@ + (void)downloadIOSParamsWithAppId:(NSString *)appId {
         }
 
         [[OSRemoteParamController sharedController] saveRemoteParams:result];
+        [OSRemoteLoggingController configure];
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. The branch is rebased onto main. Privacy consent params are now applied before logger configuration, consent revocation shuts logging down, granting consent reconfigures it, and app ID changes reset telemetry.

iOS_SDK/OneSignalSDK/UnitTests/OSLoggerAdaptersTests.swift
```diff
@@ -156,6 +157,68 @@ final class OSLoggerAdaptersTests: XCTestCase {
...
+        OneSignalLog.onesignalLog(.LL_ERROR, message: "uploaded")
+        notificationCenter.post(name: UIApplication.didEnterBackgroundNotification, object: nil)
+
+        wait(for: [telemetry.emitExpectation!, telemetry.flushExpectation!], timeout: 2)
+        controller.shutdown()
+
+        XCTAssertEqual(telemetry.records.count, 1)
+        XCTAssertEqual(telemetry.records.first?.body, "uploaded")
+        XCTAssertEqual(telemetry.records.first?.severity, LogSeverity.error)
+        XCTAssertEqual(telemetry.shutdownCount, 1)
```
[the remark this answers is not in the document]
fadi-george: Fixed in 7a3e9af3. Coverage now includes WARN raw bodies, invalid levels, disable-after-enable teardown, termination flush and shutdown, background-task completion, and exclusive scene lifecycle registration. Test teardown always removes listeners.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,376 @@
...
+                self.stopRemoteLogging()
+                return
+            }
+
+            if previousLogLevel != newConfiguration.logLevel {
+                self.stopRemoteLogging()
+            }
+
+            if self.remoteLogger == nil {
+                self.remoteLogger = self.remoteLoggerFactory(self.makeProviders(configuration: newConfiguration))
```
abdulraqeeb33: **Act on (P2) — construct the remote logger on the main thread.**

`configure(with:)` is reached from `downloadIOSParams`’s `URLSession` success callback (background queue), then this factory runs on `stateQueue` and builds `OSRemoteLogger` → `OSLoggerPlatformProvider`. That provider’s docs require main-thread construction because `init` reads `UIDevice.current.systemName` / `systemVersion`.

Hop creation (or at least UIDevice metadata capture) onto `DispatchQueue.main` before assigning `remoteLogger`; keep log routing on `stateQueue`. A regression test that configures from a background queue would lock this in.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,376 @@
...
+                self.logStartupDiagnostic()
```
abdulraqeeb33: **Consider (P3) — avoid re-entrant `OneSignalLog` under `stateQueue.sync`.**

`logStartupDiagnostic()` runs inside the same `stateQueue.sync` as setup and calls `OneSignalLog.onesignalLog(.LL_WARN, ...)`. A host `OSLogListener` that calls `OSRemoteLoggingController.reset` / `configure` would deadlock on this serial queue.

Emit the diagnostic after releasing the queue (e.g. `stateQueue.async` after setup, or outside the `sync` block).

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.h
```diff
@@ -0,0 +1,39 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ */
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OSRemoteLoggingController : NSObject
+
```
abdulraqeeb33, line 33: **Consider (P3) — hand-written ObjC header can drift from the Swift `@objc` surface.**

This manually declares `+configure` / `+reset` beside `@objc(OSRemoteLoggingController)`. It works for `OneSignal.m` today, but duplicate declarations are a drift hazard if signatures change. Prefer the module’s generated Swift header (or a single source of truth) once CI is green.

abdulraqeeb33 (changes requested):
## Multimodal adversarial review (round 2)

Thanks for the harden pass — the prior Act-On blockers from the first review are fixed (raw `OneSignalLogEvent.message`, typed dual-gate, background-task flush, scene XOR app observers, consent/app-id reset, focused tests, KMP #15 bump). Merge base is clean.

**Verdict: request changes** for one remaining composition bug; everything else looks solid pending green CI.

### Act on
1. **Construct `OSRemoteLogger` / `OSLoggerPlatformProvider` on the main thread** — `configure()` runs from the iOS params `URLSession` success callback (background), then creates the provider which reads `UIDevice` in `init`. The provider’s own docs require main-thread construction.

### Consider
2. Move `logStartupDiagnostic()` off the `stateQueue.sync` section so a host `OSLogListener` that calls `reset`/`configure` cannot deadlock.
3. Prefer one ObjC surface for `OSRemoteLoggingController` (generated Swift header vs hand-written `.h`) to avoid signature drift.

### Residual
- No physical-device `/sdk/log` smoke yet (ticket Done-when).
- CI was still pending at review time — wait for green after the main-thread fix.
- Keep `exporterLoggingEnabled` false until there is a non-reentrant diagnostic sink (`IOSLogger` → `OneSignalLog` → remote listener).

Full canvas write-up is in the Cursor session for this review.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: Fixed in a150dce3. Logger construction now hops to main outside the lifecycle queue, then installs the result only if the configuration is still current. Added a background-configuration regression test that asserts main-thread construction.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: Fixed in a150dce3. The startup diagnostic is emitted after releasing stateQueue and before registering the remote listener. Added coverage where an existing listener resets the controller from the diagnostic callback.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.h
fadi-george: Keeping the small header for now. This Swift controller is compiled into multiple umbrella targets with different generated Swift header names, so OneSignal.m does not have one stable generated header to import. The header exposes only the two fixed ObjC lifecycle methods.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -748,6 +758,11 @@ + (void)downloadIOSParamsWithAppId:(NSString *)appId {
             [[OSRemoteParamController sharedController] savePrivacyConsentRequired:required];
             [OSPrivacyConsentController setRequiresPrivacyConsent:required];
         }
+        if ([OSPrivacyConsentController requiresUserPrivacyConsent]) {
```
abdulraqeeb33: Where does this requirement come from?

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    func allows(_ level: ONE_S_LOG_LEVEL) -> Bool {
+        guard isRemoteLoggingEnabled, let threshold else {
+            return false
+        }
+        return level != .LL_NONE && level.rawValue <= threshold.rawValue
+    }
+
+    static func levelName(_ level: ONE_S_LOG_LEVEL) -> String {
```
abdulraqeeb33, line 69: was it worth moving this into KMP?

iOS_SDK/OneSignalSDK/Source/OneSignal.m
abdulraqeeb33: Not from SDK-4977 — this came from round-1 review ([r3736328806](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1703#discussion_r3736328806)): raw `URLSession` in `OneSignalLogHttpSender` bypasses `OneSignalClient`’s non-GET privacy-consent check. Android’s logger path has the same bypass today (`HttpURLConnection`, no lifecycle consent gate).

**Prefer Option A:** keep consent, and inject a shared raw/binary POST transport into `OneSignalLogHttpSender` so we reuse the SDK connection **and** apply consent at send-time (stronger than only gating `configure`).

- Don’t route through today’s `IOneSignalClient.executeRequest` — it rejects `dataRequest` and JSON-parses responses; OTLP `/sdk/log` is absolute-URL + protobuf.
- Extend the shared client with an opaque POST (url + headers + body → status) that reuses the shared `NSURLSession` + existing consent gate, then wire it through `OSRemoteLogger` (sender already has `RequestSender` / `URLSession` seams).
- **No KMP change** — `ILogHttpSender` is already the port; composition stays native.
- File an Android parity follow-up if we keep this.

**Option B** if product decides remote diagnostic logs are outside GDPR consent (match Android): drop the `OneSignal.m` / `setConsentGiven` branches and keep flag + `log_level` gates only.

Happy with either once product picks — A is the cleaner long-term shape.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
abdulraqeeb33: Leaving `levelName` native on purpose — KMP already owns encode/batch/upload via `LoggerFactory`. This is just `ONE_S_LOG_LEVEL` ↔ string at the bridge (same class of glue as Android’s `LogLevel.fromString`). Not worth a shared move in this PR.

Separate cleanup later if we want one shared severity parse used by both mobile config parsers.

abdulraqeeb33 (commented):
## Multimodal review (round 3)

Thanks for the harden passes — raw message, typed dual-gate, background-task flush, scene XOR app, and main-thread construction all look solid. `OSRemoteLoggingControllerTests` (8/8) are green; the red Build-and-Test job is failing in unrelated IAM/User/LiveActivities suites.

**Blocker before merge:** please answer the open consent thread ([r3750752628](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1703#discussion_r3750752628)) with an explicit product call (keep vs drop). That gate is review-invented, not in SDK-4977 Done-when, and Android does not mirror it today.

**If keeping consent (recommended):** inject a shared raw/binary POST transport into `OneSignalLogHttpSender` so `/sdk/log` reuses the SDK `NSURLSession` and consent runs at send-time. No KMP change needed (`ILogHttpSender` is already the injection port). Don’t force OTLP through today’s JSON-only `executeRequest`.

Also replied on the `levelName` → KMP question: leave native.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
fadi-george: It came from the earlier automated review, not SDK-4977 requirements, and Android does not apply an equivalent gate. Agreed this should not be introduced here, so I removed it in 8674e553. Consent-aware binary transport can be handled separately if product requires it.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: Yes, the shared pipeline and export behavior still benefit from living in KMP. This small mapping should remain native because ONE_S_LOG_LEVEL is an iOS type at the host boundary, while KMP accepts the canonical level name. Moving it would expand the KMP API without removing platform-specific conversion.

abdulraqeeb33 (approved):
Looks build/test fails, but looks great otherwise. thank you for addressing all the items.

nan-li (changes requested):
reviewing

iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
```diff
@@ -75,6 +75,11 @@
 		3C14E3B32FAE54C006ED053 /* OSLoggerPlatformProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AC2FAE54C006ED053 /* OSLoggerPlatformProvider.swift */; };
 		3C14E3B42FAE54C006ED053 /* KotlinByteArray+Data.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AD2FAE54C006ED053 /* KotlinByteArray+Data.swift */; };
 		3C14E3B52FAE54C006ED053 /* OSLoggerAdaptersTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3C14E3AE2FAE54C006ED053 /* OSLoggerAdaptersTests.swift */; };
+		497700000000000000000003 /* OSRemoteLoggingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 497700000000000000000001 /* OSRemoteLoggingController.swift */; };
```
nan-li: nit on these generated object IDs. It should stylistically match the other ones because there's nothing that is different about these new files.

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalLog.h
```diff
@@ -39,6 +39,7 @@ typedef NS_ENUM(NSUInteger, ONE_S_LOG_LEVEL) {
 @interface OneSignalLogEvent : NSObject
 @property(readonly)ONE_S_LOG_LEVEL level;
 @property(readonly, nonnull)NSString *entry;
+@property(readonly, nonnull)NSString *message;
```
nan-li: this is an API add we should callout and make a separate PR for release notes, as a client may be interested in getting `event.message` rather than the existing `event.entry`

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public func shutdown() {
+        telemetry.shutdown()
+    }
+}
+
+#else
+
+public final class OSRemoteLogger: OSRemoteLoggerProtocol {
```
nan-li, line 139: do we have to have this stub?

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,425 @@
...
+struct OSRemoteLoggingConfiguration {
+    static let featureFlagName = "SDK_CUSTOM_LOGGING"
+    private static let featureFlagKey = "sdk_custom_logging"
+    private static let featureFlagsKey = "sdk_remote_feature_flags"
+    private static let loggingConfigKey = "logging_config"
+    private static let logLevelKey = "log_level"
+
+    let isFeatureEnabled: Bool
+    let logLevel: String?
+    private let threshold: ONE_S_LOG_LEVEL?
```
nan-li: is there a reason we need to maintain both of these? You could do a computed property so that one is a getter to the other.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,425 @@
...
+            isListening = false
+        }
+        notificationTokens.forEach(notificationCenter.removeObserver)
+        notificationTokens.removeAll()
+        let activeRemoteLogger = remoteLogger
+        remoteLogger = nil
+        activeRemoteLogger?.shutdown()
+    }
+
+    private func registerLifecycleObservers() {
```
nan-li, line 318: There are several lifecycle things in this PR including a full listener setup but we also have OneSignalLifecycleObserver. Thoughts on using that or refactoring it to work with this feature?

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    static var current: OSRemoteLoggingConfiguration {
+        let params = OSRemoteParamController.shared().remoteParams as? [String: Any] ?? [:]
+        return OSRemoteLoggingConfiguration(remoteParams: params)
+    }
+
+    init(remoteParams: [String: Any]) {
+        let loggingConfig = remoteParams[Self.loggingConfigKey] as? [String: Any]
+        let enabledFlagNames = remoteParams[Self.featureFlagsKey] as? [String] ?? []
```
nan-li: Note this doesn't come from remote params which doesnt have "sdk_remote_feature_flags", we will have to modify it when we include featureflags / featuremanager

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -735,6 +738,7 @@ + (void)downloadIOSParamsWithAppId:(NSString *)appId {
         }
 
         [[OSRemoteParamController sharedController] saveRemoteParams:result];
+        [OSRemoteLoggingController configure];
```
nan-li, line 741: This means we will only start telemetry after remote params returns a successful response on a cold start

iOS_SDK/OneSignalSDK/OneSignal.xcodeproj/project.pbxproj
fadi-george: Fixed in 27d826ea. Replaced the synthetic ticket-shaped IDs with conventional opaque 24-character Xcode object IDs.

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalLog.h
fadi-george: Agreed. Removed the public event.message addition in 27d826ea and kept raw-body extraction internal to the remote logging controller. A client-facing raw message API can be proposed separately with release notes.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
fadi-george: Yes, it is currently needed to keep Mac Catalyst builds compiling because OneSignalKMP has no Catalyst slice. The protocol keeps the umbrella controller independent of KMP types, and this no-op implementation supplies that interface on Catalyst. SDK-4998 is the follow-up for Catalyst support; if it makes the real KMP logger available there, we can remove this stub and use the real implementation.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: Updated. `logLevel` is now a computed property derived from the parsed `threshold`, so the threshold remains the single source of truth.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: I considered reusing OneSignalLifecycleObserver, but it currently has no extension point for additional consumers and owns unrelated SDK side effects. Remote logging also needs dynamic teardown, background flush completion, and termination handling. I’d keep this isolated for this PR and handle a shared lifecycle-event abstraction as a separate refactor.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: Updated in 16fc15a0. I removed the `sdk_remote_feature_flags` lookup because that field is not exposed through iOS remote params. Remote logging is now enabled by a valid `logging_config.log_level`, and the feature-flag metadata remains empty until FeatureManager is integrated.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
fadi-george: Addressed in 110a56a6. The successful response now caches the app-scoped remote log level, and initialization restores that cached configuration before starting the next remote-params request. First-ever launch remains disabled until the backend returns a configuration, while subsequent cold starts no longer depend on a successful response.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
nan-li: ok

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
nan-li: ok

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
nan-li: ok

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -0,0 +1,451 @@
...
+            message: "OneSignal logging initialized: sdk=\(ONESIGNAL_VERSION), "
+                + "kmp=\(remoteLogger.kmpVersion), path=kmp, "
+                + "crash_dir=\(remoteLogger.crashStoragePath)"
+        )
+    }
+
+    private func makeProviders(configuration: OSRemoteLoggingConfiguration) -> OSRemoteLoggerProviders {
+        OSRemoteLoggerProviders(
+            installId: { Self.installId },
+            onesignalId: { OneSignalUserManagerImpl.sharedInstance.onesignalId },
```
nan-li: calling `OneSignalUserManagerImpl.sharedInstance.onesignalId` will spam warn logs when privacy consent is not provided since that method does a privacy consent check + WARN log. We could probably just drop the privacy check on it, which is what the OneSignalUserManagerImpl.sharedInstance.pushSubscriptionId currently does. We just have to make sure the public accessor to `OneSignal.User.onesignalId` blocks.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
nan-li: the setup still happens after the "launchOptions is set..." log, protected-data setup, migration, apns register. It feels quite late to be able to capture critical early processes

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalLog.h
nan-li: I think seeing the workaround to parse `entry`, I just realized that this logger is hooked into our public API which is why this API had to change / do a workaround. That's a bit fragile. 

Remote logging should not register as a public OSLogListener: that API is a customer-facing, presentation-oriented stream (entry includes console prefixes, no exception fields, always every level), so the exporter ends up parsing log strings, dropping stack traces, and sharing fate with app listeners instead of taking a private sink next to onesignal_Log the way Android does.

fadi-george: Addressed in 4ab4bbab. Remote logging no longer conforms to or registers as the public `OSLogListener`. `onesignal_Log` now emits a separate internal signal with the level and original raw message before notifying customer listeners, so telemetry no longer parses presentation-formatted `entry` values or shares the public listener path. Structured exception fields remain a separate concern because the current iOS logging primitive only accepts level and message.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
fadi-george: Addressed in d89c0248. Cached remote logging is now restored as the first step of `+[OneSignal init]`, before the initial launch log, protected-data setup, migration, and APNs registration. This captures those early operations on subsequent launches when cached configuration exists.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: Addressed in 4825109c. The telemetry provider now reads the internal cached OneSignal ID directly instead of invoking the public consent-gated `onesignalId` accessor for every event. This avoids repeated consent warnings while leaving `OneSignal.User.onesignalId` privacy behavior unchanged.

fadi-george:
@nan-li addressed current set of comments

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
```diff
@@ -362,7 +362,12 @@ final class OSRemoteLoggingController: NSObject {
     private func makeProviders(configuration: OSRemoteLoggingConfiguration) -> OSRemoteLoggerProviders {
         OSRemoteLoggerProviders(
             installId: { Self.installId },
-            onesignalId: { OneSignalUserManagerImpl.sharedInstance.onesignalId },
+            onesignalId: {
+                OneSignalUserDefaults.initShared().getSavedString(
+                    forKey: OS_SNAPSHOT_ONESIGNAL_ID,
```
nan-li: I recommend against a direct read and `OS_SNAPSHOT_ONESIGNAL_ID` it's going to give you the previous one

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalLog.m
```diff
@@ -113,6 +116,14 @@ void onesignal_Log(ONE_S_LOG_LEVEL logLevel, NSString* message) {
         [[OSDialogInstanceManager sharedInstance] presentDialogWithTitle:levelString withMessage:message withActions:nil cancelTitle:NSLocalizedString(@"Close", @"Close button") withActionCompletion:nil];
     }
 
+    [[NSNotificationCenter defaultCenter]
```
nan-li: posting the whole event to NSNotificationCenter on every log even when telemetry is off is probably not the most stable workaround, especially as we're going to attach like stack traces in addition to message etc

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingController.swift
fadi-george: Addressed. Telemetry now reads the live OneSignal ID from an internal `OneSignalUserManagerImpl` accessor backed by the current identity model, so it neither reads the previous snapshot nor triggers the public consent warning. Added coverage that switches the current user and verifies the ID updates.

iOS_SDK/OneSignalSDK/OneSignalCore/Source/OneSignalLog.m
fadi-george: Addressed. The `NSNotificationCenter` workaround is removed. Logging now uses a single private, lock-protected sink that is only installed while telemetry is active, and the sink contract carries the message plus structured exception type, message, and stack trace fields.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -247,6 +247,7 @@ + (void)initialize:(nonnull NSString*)newAppId withLaunchOptions:(nullable NSDic
 // TODO: For release, note this change in migration guide:
 // No longer reading appID from plist @"OneSignal_APPID" and @"GameThrive_APPID"
 + (void)setAppId:(nullable NSString*)newAppId {
+    [OSRemoteLoggingController configureFromCacheForAppId:newAppId ?: OneSignalIdentifiers.storedAppId];
```
nan-li: not a blocker, just a nit, since setAppId and setLaunchOptions are only called from `initialize:(nonnull NSString*)newAppId withLaunchOptions:(nullable NSDictionary*)launchOptions` we can move the 2 configure calls to the top of `initialize:withLaunchOptions`

nan-li (approved):
Looks great!

iOS_SDK/OneSignalSDK/Source/OneSignal.m
fadi-george: Addressed. Cached telemetry is now configured once at the top of `initialize:withLaunchOptions:` before either setter emits logs, and the stale wrapper-call comment was removed.

fadi-george: done

PR #1707, opened 2026-08-12
feat: [JWT-3] feature flags, JWT config, and the Identity Verification gate

nan-li:
# Description
## One Line Summary
Add the Identity Verification foundation: feature flags, three-state JWT requirement config, the IV gate service, and remote-params hydration with retry — nothing consumes the gate yet.

### Motivation
Identity Verification needs a known answer to “does this app require a token?” before user-scoped work can be sent. That answer is three-state (unknown / on / off), comes from remote params (`jwt_required`), and must survive a failed first-launch params download without stranding the session’s queued work forever.

### Scope
- `OSFeatureManager`, `OSUserJwtConfig`, `OSIdentityVerificationService`
- `OneSignal.m` params hydration from `jwt_required`, retry/backoff, cache clears for the new keys
- Replaces unread `require_user_id_auth` with `jwt_required` (the `requiresUserAuth` property itself is removed in a later PR)
- Unit tests for the three new types
- No request-pipeline or public JWT listener API yet

Targets `5.6-main` (already includes the project-file normalize and read-your-write consistency fixes).

# Testing
## Unit testing
- `OSFeatureManagerTests`
- `OSUserJwtConfigTests`
- `OSIdentityVerificationServiceTests`

## Manual testing
Built in sequence as part of the local JWT stack against an iOS Simulator.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -759,9 +785,29 @@ + (void)downloadIOSParamsWithAppId:(NSString *)appId {
...
+    // 5s, 10s, 20s, 40s across the remaining attempts
+    NSTimeInterval delay = DOWNLOAD_PARAMETERS_RETRY_BASE_SECONDS * (1 << MAX(_downloadParametersAttempts - 1, 0));
+    [OneSignalLog onesignalLog:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Retrying the iOS parameters download in %.0f seconds", delay]];
+
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delay * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+        // A session start or a re-init may have already succeeded or have one in flight
+        if (_downloadedParameters || _didCallDownloadParameters)
+            return;
+
+        [self downloadIOSParamsWithAppId:appId];
```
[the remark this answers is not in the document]
nan-li: This is a non-issue because the SDK doesn't actually support changing app ID.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -665,6 +683,9 @@ + (void)handleAppIdChange:(NSString*)appId {
         [sharedUserDefaults removeValueForKey:OSUD_RECEIVE_RECEIPTS_ENABLED];
         [sharedUserDefaults removeValueForKey:OS_PUSH_SUBSCRIPTION_MODEL_STORE_KEY];
 
+        [sharedUserDefaults removeValueForKey:OSUD_USE_IDENTITY_VERIFICATION];
+        [sharedUserDefaults removeValueForKey:OSUD_SDK_FEATURE_FLAGS];
```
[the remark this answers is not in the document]
nan-li: This is a non-issue because the SDK doesn't actually support changing app ID.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -710,15 +731,20 @@ + (void)setConsentGiven:(BOOL)granted {
...
 
-        if (result[IOS_REQUIRES_USER_ID_AUTHENTICATION]) {
-            OneSignalUserManagerImpl.sharedInstance.requiresUserAuth = [result[IOS_REQUIRES_USER_ID_AUTHENTICATION] boolValue];
+        // A response that omits the key means Identity Verification is off for this app; an empty
+        // response answers nothing, so the cached requirement stands
+        if (result != nil) {
+            id jwtRequired = result[IOS_JWT_REQUIRED];
+            BOOL requiresUserAuth = jwtRequired != (id)[NSNull null] && [jwtRequired boolValue];
+            [OSUserJwtConfig.shared hydrateWithRequiresUserAuth:requiresUserAuth];
         }
```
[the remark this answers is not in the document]
nan-li: minor, not addressing
- requires server to return success but no body
- and only affects on the first launch when IV=unknown
- only lasts the app process and will refetch next launch

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -759,9 +785,29 @@ + (void)downloadIOSParamsWithAppId:(NSString *)appId {
 
     } onFailure:^(OneSignalClientError *error) {
         _didCallDownloadParameters = false;
+        [self scheduleDownloadIOSParamsRetryWithAppId:appId];
```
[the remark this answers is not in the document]
nan-li: Addressed the gating part in 859489e80.

Outer retries now only run when `OSNetworkingUtils.getResponseStatusType` is `OSResponseStatusRetryable`, so classified 4xx are not re-armed for the session. Nested transport budget (client then outer) stays intentional for first-launch stranding; folding into one policy is out of scope here.

Not adding `OneSignal.m` unit coverage in this PR — leaf types are covered; the statics/`dispatch_after` path needs a dedicated test hook. App-id invalidation of a pending retry is a non-issue (SDK does not support changing app ID).

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Jwt/OSIdentityVerificationService.swift
```diff
@@ -0,0 +1,115 @@
...
+            } else {
+                jwtConfigHydratedHandlers.append((observer, handler))
+            }
+        }
+
+        let alreadyKnown = jwtConfig.requirement
+        guard alreadyKnown != .unknown else {
+            return
+        }
+        handler(alreadyKnown)
```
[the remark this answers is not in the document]
nan-li: Addressed in 859489e80.

`addOnJwtConfigHydratedHandler` now reads `jwtConfig.requirement` inside the same `handlerLock` critical section as registration, so catch-up cannot follow a fresher hydrate fire with a stale snapshot.

nan-li:
rebased on main, added a commit to reconcile the merged Feature Manager

fadi-george:
Potential issues:
- Empty successful params responses can leave JWT requirements unknown while disabling retries.
- App ID changes clear persisted JWT state but not the in-memory singleton.
- Retry state is unsynchronized and pending retries may use a stale app ID.
- Concurrent hydration and registration can duplicate or reorder callbacks.
- `refreshIfUnknown()` does not notify waiting observers.

fadi-george:
Maybe also address flakey tests.

abdulraqeeb33 (commented):
Thanks for splitting the gate out this cleanly. A couple of lifecycle holes around app-id change are worth closing before later PRs start trusting `requirement`.

Inline notes below, with a suggested fix for each.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -723,6 +741,8 @@ + (void)handleAppIdChange:(NSString*)appId {
         [sharedUserDefaults removeValueForKey:OSUD_RECEIVE_RECEIPTS_ENABLED];
         [sharedUserDefaults removeValueForKey:OS_PUSH_SUBSCRIPTION_MODEL_STORE_KEY];
 
+        [sharedUserDefaults removeValueForKey:OSUD_USE_IDENTITY_VERIFICATION];
```
abdulraqeeb33, line 744: Easy to miss because the disk key *does* get cleared.

`handleAppIdChange` removes `OSUD_USE_IDENTITY_VERIFICATION` and resets `OSFeatureManager`, but `OSUserJwtConfig.shared` keeps the previous app’s on/off. `refreshIfUnknown` only does anything while the value is still unknown, so it cannot repair this.

The bad direction is non-IV → IV: the SDK still thinks JWT is off and will create the new app’s user unsigned. Demo’s change-app-id path hits this.

I’d promote `resetRequirementToUnknownForTests` to a production reset (set `_requirement` to `.unknown`, clear the cache key) and call it next to `resetAndClearCachedFlags`.

A test that hydrates on, changes app id, then asserts `requirement == .unknown` would lock this in.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
```diff
@@ -818,9 +843,33 @@ + (void)downloadIOSParamsWithAppId:(NSString *)appId {
...
+    // 5s, 10s, 20s, 40s across the remaining attempts
+    NSTimeInterval delay = DOWNLOAD_PARAMETERS_RETRY_BASE_SECONDS * (1 << MAX(_downloadParametersAttempts - 1, 0));
+    [OneSignalLog onesignalLog:ONE_S_LL_DEBUG message:[NSString stringWithFormat:@"Retrying the iOS parameters download in %.0f seconds", delay]];
+
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delay * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+        // A session start or a re-init may have already succeeded or have one in flight
+        if (_downloadedParameters || _didCallDownloadParameters)
+            return;
+
+        [self downloadIOSParamsWithAppId:appId];
```
abdulraqeeb33, line 869: Same family as the in-memory reset above.

This timer captures `appId`. `handleAppIdChange` then flips `_downloadedParameters` / `_didCallDownloadParameters` back to false, so 5–40s later the old app’s params can hydrate the new process-wide `jwt_required`. The success handler at `hydrateWithRequiresUserAuth` has the same gap: it never checks that `appId` is still current.

In this `dispatch_after` *and* at the top of the params success handler, I’d bail when the captured `appId` is not `OneSignalIdentifiers.currentAppId`.

A test like “retry armed for A, switch to B, B’s policy must not become A’s” would cover both.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Jwt/OSUserJwtConfig.swift
```diff
@@ -0,0 +1,134 @@
...
+    @objc
+    public func hydrate(requiresUserAuth: Bool) {
+        let hydrated: OSRequiresUserAuth = requiresUserAuth ? .on : .off
+        // Keep the log and the handler out of the lock; either can re-enter and read the requirement.
+        let (previous, handler) = lock.withLock { () -> (OSRequiresUserAuth, ((OSRequiresUserAuth) -> Void)?) in
+            let previous = _requirement
+            _requirement = hydrated
+            // Written even when the value is unchanged, so a launch whose write was dropped by locked
+            // storage still ends up with the requirement on disk.
+            OneSignalUserDefaults.initShared().saveInteger(forKey: OSUD_USE_IDENTITY_VERIFICATION, withValue: hydrated.rawValue)
```
abdulraqeeb33, line 90: Smaller, same file family. `hydrate` keeps the `NSLock` across `saveInteger`, and that path calls `synchronize`. Readers of `requirement` are on the repo queue, IAM on main, and the executors, so they all wait on a disk flush.

You already kept the log and the handler *out* of the lock, which is the right instinct. I’d do the same for the write: update `_requirement` under the lock, then `saveInteger` after release. The “write even when unchanged” comment still holds.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
nan-li: Yeah, we have this app ID change stuff but it's really not supported after we shipped v5. I think something similar about app id changes was called out above and it was marked is a non-issue because the SDK doesn't actually support changing app ID. Linear ticket SDK-5142.

We do have a lot of stale code in this method that we either should clean up or actually handle app ID changes (which we have a linear ticket about that discussion a little while back). The latest conversation was that Aug 11 slack thread asking for this feature.

Followup: A developer changes the app id between launches, not in the middle of a run. On a new launch the in-memory config is built fresh from disk, and the PR already wipes the disk key before anything reads it. So the stale value they describe can only happen if the app id changes while the app is running, and that is the case SDK-5142 (aug 11 convo) will decide.

iOS_SDK/OneSignalSDK/Source/OneSignal.m
nan-li: same response, we never actually support app id change, and there's no real change to downloading remote params here except that the SDK will retry for a retriable error up to 4 times.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Jwt/OSUserJwtConfig.swift
nan-li: thanks, filed into a followup in SDK-5137

nan-li:
> Potential issues:
> 
> * Empty successful params responses can leave JWT requirements unknown while disabling retries.
> * App ID changes clear persisted JWT state but not the in-memory singleton.
> * Retry state is unsynchronized and pending retries may use a stale app ID.
> * Concurrent hydration and registration can duplicate or reorder callbacks.
> * `refreshIfUnknown()` does not notify waiting observers.

Thanks for your review, this is my thoughts on the issues surfaced:

* Empty successful params responses can leave JWT requirements unknown while disabling retries.
   * **Not addressing.** It needs the server to return 200/201/202 with no body, which the params endpoint never does. If it did, we keep the value on disk from the last successful fetch. It only matters when there is no cached value yet, so the requirement truly is unknown and leaving it unknown is correct. The download is fetched again on the next launch, so it lasts one process.
* App ID changes clear persisted JWT state but not the in-memory singleton.
   * **Not addressing, tracked in SDK-5142** Same answer as [my reply to Abdul](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1707#discussion_r3896590972). App ID changes are not supported since v5, and a change between launches is covered because the config is rebuilt from the cleared disk key on the next process. A mid-process change is what the ticket scopes.
* Retry state is unsynchronized and pending retries may use a stale app ID.
   * **Not addressing.** The race is extremely rare and harmless. It needs a foreground event to land in the microseconds the failure callback runs, and the outcome is one extra retry or a retry a step earlier or later than planned. Happy to hop that callback onto main in the follow-up if you want it closed. The stale app ID part is the same answer as above.
* Concurrent hydration and registration can duplicate or reorder callbacks.
   * **Not addressing.** Both are contained. Duplicate delivery is the normal path. On every launch after the first, each handler runs once at registration with the cached value and again when that session's params hydrate, so the three handlers in PR4 and PR6 are idempotent by design. Ordering can only slip if a hydrate overlaps a registration, and registration happens once, inside `start()`, so that is a first-launch response landing in the microseconds between three `addOnJwtConfigHydratedHandler` calls. The one ordering that matters (restore a silenced subscription before Create User goes out) only applies in a session that already cached `on`, and there every handler registered long before the response, so they run in order on one thread. Happy to track last-delivered per observer in the follow-up if you want it closed on paper.
* `refreshIfUnknown()` does not notify waiting observers.
   * **Follow-up, filed on SDK-5137** True in isolation. In practice the only caller runs during startup, before any observer exists, and anyone who registers later gets the current value immediately. So nothing is ever left waiting. The one-line fix is to fire the handler when refresh moves the value off unknown.

abdulraqeeb33 (approved):

fadi-george (approved):

PR #1722, opened 2026-08-21
feat: [SDK-5047] drive iOS logging from remote params

abdulraqeeb33:
## Summary

iOS already parsed `logging_config.log_level` from `ios_params`, but never used it to drive the logging lifecycle — so remote crash/OTel logging couldn't be enabled or adjusted server-side the way it can on Android. This ports Android's model and lifecycle, then fixes three pre-existing defects in the KMP adapter that surfaced once the backend actually started enabling logging.

Two commits, independently reviewable:

**`feat:` — config + lifecycle port.** Adds `OSRemoteLoggingConfiguration` (Android's `RemoteLoggingConfigModel`: persisted level + enabled flag derived from the server sending a valid level) and `OSRemoteLoggingConfigEvaluator` (Android's `OtelConfigEvaluator`: pure old-vs-new diff yielding Enable / Disable / UpdateLogLevel / NoChange). The controller applies those actions rather than only filtering log lines. The cache also stops round-tripping through a synthesized params-shaped dictionary, so it no longer depends on the params API response shape — relevant while SDK-5046 is changing that shape.

**`fix:` — KMP boundary.** Kotlin/Native only allows calling exported `suspend` functions from the main thread, but all three crossings were reached from the controller's serial queue and from URLSession callbacks; they're now marshalled to main. `telemetry.shutdown()`'s five-second `runBlocking` drain was reachable from `initialize:` and app-id changes via `stateQueue.sync` and could stall the UI, so the drain is deferred. And the HTTP exporter's diagnostics logged through `OneSignalLog`, whose sink is invoked unconditionally, so at verbose levels each export POST generated a new record and guaranteed another POST.

## Behavior changes worth a close look

- **`NONE` now means enabled-but-exporting-nothing**, matching Android's hydrate path, so the crash handler stays armed instead of collapsing to fully disabled.
- **A level change rebuilds the logger**, as Android does in `startLogging`. This isn't just parity: the platform provider handed to KMP is built alongside the logger, so reusing the instance would keep reporting the previous level into KMP and would never re-run the crash uploader after a level escalates away from `NONE`.
- **`start()` now rejects once shutdown has *begun***, not once it has finished. Required by the deferred drain — otherwise a logger told to shut down could still install a crash handler nothing would unregister, permanently blocking every later logger from installing its own.

## Verification

Verified end to end on the demo app against a real app id returning `log_level: VERBOSE`:

- Config parsed to VERBOSE/enabled; evaluator returned `Enable(VERBOSE)`; cache round-trip preserved both fields.
- Remote records confirmed arriving at log-ingestion-service, `/sdk/log` returning 202.
- Uncaught `NSException` captured by `OSLogCrashHandler`, persisted durably (7,861-byte `.otlp`), and uploaded by `LogCrashUploader` on the following launch. Confirmed server-side.

Tests: 18 `OSRemoteLoggingControllerTests` + 67 `OneSignalOSCoreTests`, all passing. Includes a verbatim `ios_params` payload fixture, the full evaluator transition matrix, and direct lifecycle tests for the start-after-shutdown guard.

This went through an adversarial multi-model review before opening; the critical finding (the `start()` guard) and three warnings were fixed as a result.

## Test plan

- [ ] CI green, including swiftlint and the Catalyst KMP logger host
- [ ] Confirm a mid-session level change (e.g. `ERROR` → `VERBOSE`) rebuilds the logger and the new level reaches KMP
- [ ] Confirm `NONE` keeps crash capture armed while exporting nothing
- [ ] Confirm cold start from cache arms the crash handler before the params call returns

## Follow-ups filed separately

- SDK-5058 — export batches near 128 KB rejected with 403 and silently dropped
- SDK-5059 — `isOneSignalAtFault` drops SDK crashes when the OneSignal frame is elided

## Not included

The KMP submodule pointer is deliberately left off this branch; local testing ran against a newer KMP commit than `main` pins, and CI will use `main`'s.

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
@@ -60,11 +60,14 @@ final class OSRemoteLoggerLifecycle {
     private var isShuttingDown = false
     private var isShutdown = false
 
-    var canStartUploader: Bool { canEmit }
+    var canStartUploader: Bool { isActive }
 
-    /// Records stop being accepted the moment shutdown begins, rather than once the
-    /// final drain finishes, so teardown never lets a late log escape.
-    var canEmit: Bool {
+    /// True while the transport is usable and teardown has not begun. Gates record
+    /// emission, uploader start, and explicit flushes alike: the drain inside
+    /// `shutdown()` also crosses into KMP, and two concurrent crossings are unsafe.
+    /// Keyed on shutdown *beginning* rather than finishing, because that drain is
+    /// asynchronous.
+    var isActive: Bool {
         lock.lock()
         defer { lock.unlock() }
         return isStarted && !isShuttingDown && !isShutdown
@@ -373,7 +376,7 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
         exceptionMessage: String?,
         exceptionStacktrace: String?
     ) {
-        guard lifecycle.canEmit else {
+        guard lifecycle.isActive else {
             return
         }
         let telemetry = self.telemetry
@@ -381,7 +384,7 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
         Self.onMain {
             // Re-checked here because the hop is asynchronous: teardown can begin
             // between the caller-side check above and the crossing into KMP.
-            guard lifecycle.canEmit else {
+            guard lifecycle.isActive else {
                 return
             }
             LogLoggingHelper.shared.log(
@@ -398,7 +401,16 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
 
     public func forceFlush(completion: @escaping () -> Void) {
         let telemetry = self.telemetry
+        let lifecycle = self.lifecycle
         Self.onMain {
+            // Skipped once teardown has begun, because `shutdown()`'s deferred drain
+            // flushes the same telemetry and both would cross into KMP at once. The
+            // completion still has to run either way: callers end a background task
+            // in it, and swallowing it would leak that task.
+            guard lifecycle.isActive else {
+                completion()
+                return
+            }
             telemetry.forceFlush(completionHandler: { _ in completion() })
         }
     }
--- a/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
@@ -54,13 +54,17 @@ public extension OSRemoteLoggerProtocol {
 
 @_implementationOnly import OneSignalKMP
 
-private final class OSRemoteLoggerLifecycle {
+final class OSRemoteLoggerLifecycle {
     private let lock = NSLock()
     private var isStarted = false
     private var isShuttingDown = false
     private var isShutdown = false
 
-    var canStartUploader: Bool {
+    var canStartUploader: Bool { canEmit }
+
+    /// Records stop being accepted the moment shutdown begins, rather than once the
+    /// final drain finishes, so teardown never lets a late log escape.
+    var canEmit: Bool {
         lock.lock()
         defer { lock.unlock() }
         return isStarted && !isShuttingDown && !isShutdown
@@ -76,10 +80,13 @@ private final class OSRemoteLoggerLifecycle {
         return true
     }
 
+    /// Rejects once shutdown has *begun*, not just once it has finished. The final
+    /// drain is asynchronous, so a logger told to shut down can otherwise still be
+    /// started afterwards and install a crash handler nothing will ever unregister.
     func start() -> Bool {
         lock.lock()
         defer { lock.unlock() }
-        guard !isStarted, !isShutdown else {
+        guard !isStarted, !isShuttingDown, !isShutdown else {
             return false
         }
         isStarted = true
@@ -170,6 +177,9 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
     private let lifecycleOperationLock = NSLock()
     private let uploaderOwner = UUID()
 
+    /// Serial so overlapping teardowns cannot stack several bounded drains at once.
+    private static let teardownQueue = DispatchQueue(label: "com.onesignal.logger.remote-teardown")
+
     public convenience init(
         installIdProvider: @escaping () -> String,
         onesignalIdProvider: @escaping () -> String?,
@@ -237,9 +247,12 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
         let crashLogger = OSCrashLogger()
         let lifecycle = OSRemoteLoggerLifecycle()
         let fileStore = FileLogStore(rootPath: provider.crashStoragePath)
+        // Console-only logger on purpose. Exporter diagnostics describe the POST that
+        // ships log records, so routing them through OneSignalLog would feed each POST
+        // back into the export queue as a new record and never settle.
         let httpSender = Self.makeHttpSender(
             requestSender: requestSenderOverride,
-            logger: logger,
+            logger: crashLogger,
             isDiagnosticsEnabled: exporterLoggingEnabledProvider,
             lifecycle: lifecycle
         )
@@ -309,16 +322,29 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
         let logger = self.logger
         let lifecycle = self.lifecycle
         OSCrashUploaderCoordinator.shared.enqueue(owner: owner) {
-            guard lifecycle.canStartUploader else {
-                OSCrashUploaderCoordinator.shared.finish(owner: owner)
-                return
-            }
-            crashUploader.start { error in
-                if let error {
-                    logger.error(message: "LogCrashUploader failed: \(error.localizedDescription)")
+            Self.onMain {
+                guard lifecycle.canStartUploader else {
+                    OSCrashUploaderCoordinator.shared.finish(owner: owner)
+                    return
                 }
-                OSCrashUploaderCoordinator.shared.finish(owner: owner)
+                crashUploader.start { error in
+                    if let error {
+                        logger.error(message: "LogCrashUploader failed: \(error.localizedDescription)")
+                    }
+                    OSCrashUploaderCoordinator.shared.finish(owner: owner)
+                }
             }
+        }
+    }
+
+    /// Kotlin/Native only supports calling exported `suspend` functions from the main
+    /// thread, so every crossing into KMP has to be marshalled here. Callers reach this
+    /// class from the logging controller's serial queue and from URLSession callbacks.
+    private static func onMain(_ work: @escaping () -> Void) {
+        if Thread.isMainThread {
+            work()
+        } else {
+            DispatchQueue.main.async(execute: work)
         }
     }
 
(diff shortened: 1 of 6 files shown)


abdulraqeeb33:
Thanks — all four addressed in 86b7363, plus the lint failure. Details and one thing I could only partly fix:

**`forceFlush()` entering KMP concurrently with async shutdown.** Gated now. `shutdown()`'s deferred drain flushes the same telemetry, so the explicit flush is skipped once teardown has begun. The completion still fires on every path, including the skipped one — `flushForLifecycle` ends a `UIBackgroundTask` in it, so swallowing it would leak the task. While doing this I renamed the predicate from `canEmit` to `isActive`, since it now gates emission, uploader start, and flushes alike; they all cross the same boundary, and having three names for one condition was inviting the next bug.

**Cached `is_enabled: true` with an invalid or missing level.** Normalized at the source rather than at the consumer: `init(cached:)` now falls back to `ERROR` when enabled without a parseable level, matching Android's `logLevel ?: ERROR`. That keeps the stored config self-consistent instead of relying on every reader to apply the same default. Covered by `testCachedEnabledWithUnparseableLevelFallsBackToError`, which also asserts the inverse — `is_enabled: false` must not manufacture a level.

**Active crash upload continuing after shutdown begins.** Partly fixed, and I want to be straight about the limit. Starting is now blocked: the uploader's enqueue closure checks `isActive`, which flips the moment `beginShutdown()` runs, so no upload begins after teardown starts and `finish()` won't kick off a queued one either. But an upload already in flight can't be interrupted — `LogCrashUploader` exposes no cancellation, and its suspend `start()` runs to completion inside KMP. Genuinely stopping it needs a cancellation API in the KMP module, which is submodule work. Happy to file that as a follow-up if you agree it's worth it; the current behavior is that an in-flight upload finishes and deletes the reports it sent, which is at least not lossy.

**License header artifacts.** Removed. Those were line-number gutter prefixes from a file read that leaked into the content when I wrote the file — good catch.

**Lint.** Root cause was unrelated to the review: this branch was one commit behind `main`, and that commit (#1719) replaced the inline `swiftlint` step with a stricter standalone job. Merged `main` in, then found the repo has exactly one error-severity violation and it was mine — my test class was 352 lines against `type_body_length`'s 350 error threshold. Split the configuration cases into `OSRemoteLoggingConfigurationTests`, which is better organization regardless. `swiftlint` now exits 0 locally. The remaining violations in these files are `file_length` warnings, which the linter tolerates — there are 296 such warnings repo-wide.

**On the `NONE` question** — you're right that it deserves an explicit decision, and it's a product call rather than mine. Today `NONE` means "enabled, crash handler installed process-wide, nothing exported," which matches Android's hydrate path. Android is actually inconsistent with itself here: its cached path uses `OtelIdResolver.resolveRemoteLoggingEnabled()`, which excludes `NONE` and therefore does *not* install the handler. So there's no single Android behavior to copy. If the intent of `NONE` is "we are not observing this app," installing a process-wide `NSUncaughtExceptionHandler` is arguably the wrong default and it should behave like disabled. If the intent is "stop shipping logs but stay ready to enable," current behavior is right. I'd rather you or @sherwin decide than pick silently — say which and I'll make it match, and align Android as a follow-up.

Tests: 19 in `UnitTests` (12 controller + 7 configuration) and 67 in `OneSignalOSCoreTests`, all passing.

abdulraqeeb33:
Good catch — the flag check was insufficient for exactly the reason you describe. Fixed in e900732 by tracking the operation rather than testing a flag.

`forceFlush` now claims a slot on the lifecycle (`beginFlush`) and releases it in the KMP completion (`endFlush`). `shutdown()` waits for outstanding slots before running its drain. Since `beginFlush` refuses new flushes once `beginShutdown()` has run, the set the drain waits on can only shrink, so there's no starvation path.

Two deliberate choices in there:

The wait is bounded at five seconds, matching the cap KMP already applies to its own drain in `LogTelemetryRemoteImpl.shutdown`. A wedged flush therefore delays teardown by no more than the drain itself already could, rather than wedging it permanently — which matters because teardown runs on app-id changes and at termination.

Blocking is safe where it happens: the wait runs on the teardown queue, while the KMP completion that releases the slot resumes on main. So the releasing thread is never the blocked one.

Covered by two new lifecycle tests — `testLifecycleShutdownWaitsForInFlightFlush` asserts the drain blocks while a slot is held and proceeds once released, and `testLifecycleRefusesNewFlushOnceShutdownBegins` asserts the shrinking-set property. 69 tests in `OneSignalOSCoreTests`, all passing, and `swiftlint` exits 0.

---

Separately, on the red **Build and Test**: I believe that's an unrelated flake rather than anything on this branch. The failure was `SubscriptionUpdateRaceTests.testPendingPrePermissionUpdateSendsLiveSubscribedPayloadAfterGrant`, which is user/subscription code with no path to the logging changes here. It passes locally (all four cases in that suite), the same job passed on this branch's previous run, and that specific file was modified by #1719 as part of "fix flakey tests" — which this branch only picked up when I merged `main` in to get the new lint job.

I tried re-running the failed job to confirm, but it errored in 19s with `Artifact not found for name: OneSignalKMP-XCFramework` — the artifacts from that three-day-old run had expired, so the rerun was inconclusive rather than informative. The push above has started a fresh run, which should settle it. If it fails again on the same test I'll dig in properly rather than assume.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
```diff
@@ -54,13 +54,22 @@ public extension OSRemoteLoggerProtocol {
...
-    private let lock = NSLock()
+final class OSRemoteLoggerLifecycle {
+    /// A condition rather than a plain lock so teardown can wait on in-flight flushes.
+    private let lock = NSCondition()
     private var isStarted = false
     private var isShuttingDown = false
     private var isShutdown = false
+    private var activeFlushes = 0
+
+    var canStartUploader: Bool { isActive }
```
nan-li: nit, do we need another alias for `isActive`? `canStartUploader` is only read in one place in `start` and here I think it makes as much sense to use `isActive` just as appropriately?

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
```diff
@@ -309,19 +363,32 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
...
                 }
-                OSCrashUploaderCoordinator.shared.finish(owner: owner)
             }
         }
     }
 
+    /// Kotlin/Native only supports calling exported `suspend` functions from the main
+    /// thread, so every crossing into KMP has to be marshalled here. Callers reach this
+    /// class from the logging controller's serial queue and from URLSession callbacks.
+    private static func onMain(_ work: @escaping () -> Void) {
```
nan-li: nan-todo: check https://github.com/OneSignal/OneSignal-KMP-SDK/pull/15

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
```diff
@@ -309,19 +363,32 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
...
+        if Thread.isMainThread {
+            work()
+        } else {
+            DispatchQueue.main.async(execute: work)
```
nan-li: nan-todo: check if these two implementations are different, running work() vs running work on async

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
```diff
@@ -347,32 +414,76 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
...
-            exceptionType: exceptionType,
-            exceptionMessage: exceptionMessage,
-            exceptionStacktrace: exceptionStacktrace,
-            completionHandler: { _ in }
-        )
+        guard lifecycle.isActive else {
+            return
+        }
+        let telemetry = self.telemetry
+        let lifecycle = self.lifecycle
```
nan-li: nan-todo: figure out why we needed to snapshot

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingConfiguration.swift
```diff
@@ -0,0 +1,192 @@
...
+        case "VERBOSE", "TRACE":
+            return .LL_VERBOSE
+        default:
+            return nil
+        }
+    }
+}
+
+/// Snapshot used by [OSRemoteLoggingConfigEvaluator], matching Android's `OtelConfig`.
+struct OSRemoteLoggingConfig: Equatable {
```
nan-li: do we need this snapshot of OSRemoteLoggingConfiguration? It can be confusing since it is named so similarly. The only usage I see of this is an intermediary is to turn OSRemoteLoggingConfiguration into  OSRemoteLoggingConfig to pass into OSRemoteLoggingConfigEvaluator.evaluate when the evaluate can just do on OSRemoteLoggingConfiguration directly?

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
nan-li: https://github.com/OneSignal/OneSignal-KMP-SDK/pull/15 made some changes for calls off the main thread. Do we still throw without this move to the main thread here, or can it be removed? If it can, the flush/shutdown races will also be remedied I believe.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
```diff
@@ -309,19 +363,32 @@ public final class OSRemoteLogger: OSRemoteLoggerProtocol {
...
+                        logger.error(message: "LogCrashUploader failed: \(error.localizedDescription)")
+                    }
+                    OSCrashUploaderCoordinator.shared.finish(owner: owner)
                 }
-                OSCrashUploaderCoordinator.shared.finish(owner: owner)
             }
         }
     }
 
+    /// Kotlin/Native only supports calling exported `suspend` functions from the main
```
nan-li: https://github.com/OneSignal/OneSignal-KMP-SDK/pull/15 made some changes for calls off the main thread. Do we still throw without this move to the main thread here, or can it be removed? If it can, the flush/shutdown races will also be remedied I believe.

abdulraqeeb33: Great catch, thank you — I think you're right, and I'd missed that PR entirely.

I dug into it: KMP #15 sets `kotlin.native.binary.objcExportSuspendFunctionLaunchThreadRestriction=none`, which is exactly the flag that lifts the restriction, and `72581f39` is an ancestor of both our pinned commit (`17dcaba`) and my local submodule (`d417111`). So the KMP we build against already has it disabled and this hop should be unnecessary.

I am running an experiment next and will report back before touching it.

iOS_SDK/OneSignalSDK/Source/OSRemoteLoggingConfiguration.swift
abdulraqeeb33: Agreed, and thank you — you're right that it earned nothing. Dropped in 020dbce.

`OSRemoteLoggingConfigEvaluator.evaluate` now takes `OSRemoteLoggingConfiguration` directly, which deletes the type, its `disabled` constant, and the conversion at the single call site. Net 15 lines lighter.

For context on why it was there: Android keeps `OtelConfig` separate from `RemoteLoggingConfigModel` because the model is persistence-backed and can't be compared directly, so it needs a plain snapshot to diff. I mirrored that structure without noticing the reason for it doesn't apply here — `OSRemoteLoggingConfiguration` is already an `Equatable` value type. I left a short comment on the evaluator noting that, so the next person porting from Android doesn't reintroduce it.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
abdulraqeeb33: Ran the experiment — you were right, and it's removed in 7771667.

The probe calls an exported suspend function from `DispatchQueue.global()` (asserting `Thread.isMainThread == false` first) and it completes cleanly, no exception. That generalizes, since `objcExportSuspendFunctionLaunchThreadRestriction` is a single global setting rather than per-function — the check lives in the coroutine-launch path every exported suspend function goes through. So all three hops are gone, along with the helper.

I kept the probe as `testKmpSuspendCallSucceedsOffMainThread` so this fails loudly if that gradle flag ever gets dropped, which is the failure mode that sent me down this path in the first place.

Two knock-on wins beyond the deletion: the check-then-cross window in `log()` disappears, because the lifecycle guard and the KMP call are now sequential on one thread, and we no longer pay a main-queue dispatch per exported record — that was adding up at VERBOSE.

One thing I did keep is the slot tracking on `forceFlush`. The flush completes asynchronously even when started inline, so shutdown's drain can still overlap it; that isn't a consequence of the hop. Happy to be talked out of it if you see it differently.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSRemoteLogger.swift
abdulraqeeb33: Agreed — dropped in 7771667. `canStartUploader` was just a second name for the same predicate with one reader, and `start` now reads `isActive` directly. I also reworded the doc comment, since it had justified the alias in terms of the main-thread marshalling that's now gone.

nan-li (approved):

PR #1724, opened 2026-08-25
chore: [SDK-5068] raise minimum iOS version to 12

fadi-george:
# Description
## One Line Summary
Raise the OneSignal iOS SDK minimum supported deployment target from iOS 11 to iOS 12.

### Motivation
Align the SDK's package metadata, Xcode targets, and support tooling with the documented iOS 12 minimum.

### Scope
Updates Swift Package Manager, CocoaPods, SDK and development-app Xcode deployment targets, and the GitHub bug report template. No runtime behavior or public APIs change.

### Other
Multi-model review completed with no blocking findings.

# Testing
## Unit testing
No unit tests were added because this change only updates deployment and package configuration.

## Manual testing
- `swift package dump-package`
- Ruby syntax validation for both podspecs
- `plutil` and `xcodebuild -list` validation for both Xcode projects
- Release build of the `OneSignalCore` scheme for a generic iOS Simulator destination

The targeted build passed with existing compiler deprecation and null-conversion warnings. Full CI will run on this PR. Device testing is not applicable to a configuration-only change.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] I have filled out all **REQUIRED** sections above
   - [x] PR does one thing
     - If it is hard to explain how any codes changes are related to each other then it most likely needs to be more than one PR
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [x] I have included test coverage for these changes, or explained why they are not needed
   - [x] All automated tests pass, or I explained why that is not possible
   - [x] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [x] Code is as readable as possible.
      - Simplify with less code, followed by splitting up code into well named functions and variables, followed by adding comments to the code.
   - [x] I have reviewed this PR myself, ensuring it meets each checklist item
      - WIP (Work In Progress) is ok, but explain what is still in progress and what you would like feedback on. Start the PR title with "WIP" to indicate this.

Made with [Cursor](https://cursor.com)

PR #1728, opened 2026-08-26
feat: [SDK-5106] report host app build toolchain on iOS logs

abdulraqeeb33:
## Summary

Linear: [SDK-5106](https://linear.app/onesignal/issue/SDK-5106/report-host-xcode-version-and-deployment-target-on-ios-remote-logs)

`OSLoggerPlatformProvider` shipped `swiftVersion`, `kotlinVersion`, and `additionalVersionAttributes` as placeholders (`nil` / Catalyst-only), so iOS logs carried nothing about the toolchain that built the host app. `LogFields` uses `putIfValueNotBlank`, so those attributes were silently absent from every record rather than failing loudly.

This reads Xcode's `DT*` keys and the declared deployment target from the **running executable's** `Info.plist` via `Bundle.main`.

**Why not a compile-time check.** `#if swift(>=6.0)` inside SDK source is evaluated when *we* build the SDK. Since we distribute a prebuilt XCFramework (podspecs vend `OneSignalCore.xcframework`, `Package.swift` uses binary targets), that value freezes at our build time and is identical for every customer — it would duplicate `ossdk.sdk_base_version` and tell you nothing about the app. Reading `Bundle.main` describes the customer's build and varies per app.

### Attributes added

Emitted as `ossdk.<key>` resource attributes (attached once per resource, not per record):

| Attribute | Source | Example |
| --- | --- | --- |
| `xcode_version` | `DTXcode`, decoded | `26.2` |
| `xcode_build` | `DTXcodeBuild` | `17C52` |
| `minimum_os_version` | `MinimumOSVersion` | `12.0` |
| `minimum_macos_version` | `LSMinimumSystemVersion` (Catalyst / macOS) | `13.1` |

`apple_platform` (`mac_catalyst`) is pre-existing and stays. `DTXcode` is packed (`"2620"` = 26.2) so it is decoded rather than shipped raw. Blank and missing keys are dropped rather than emitted empty.

`LSMinimumSystemVersion` is a macOS version, so it is not folded into `minimum_os_version` — a query for hosts below iOS 15 would otherwise include Catalyst apps pinned to macOS 13. Both keys can be present and are emitted independently.

Review trimmed `build_compiler`, `build_platform_*`, and `build_sdk_*` — they duplicated existing `device.*` / `os.*` fields or each other.

### Why `minimum_os_version` matters

These attributes exist to answer support questions, not for general telemetry, and this is the one that answers **"can we raise our deployment target?"**

It is deliberately distinct from `os.version`, which is already emitted. `os.version` is what a customer's users are *running*; `minimum_os_version` is what their build is *pinned to*. A customer can serve 100% iOS 18 traffic while still declaring min iOS 13 — and raising our deployment target breaks their build regardless of where their users are. #1724 just raised our minimum to iOS 12; this is the field that would show who that affects before the next bump.

## Why `swiftVersion` stays nil

An earlier revision of this PR approximated `swiftVersion` from the Xcode version via a lookup table. That has been dropped, and the field is left `nil`. Three reasons:

**1. It is redundant with `xcode_version`.** Xcode determines the Swift compiler 1:1 — there is no Xcode 16 shipping a Swift 5.9 compiler. Anything derived from the Xcode version is a lossy re-encoding of a field we already emit exactly, and can only ever be less accurate than the value sitting next to it.

**2. The one thing a distinct value could carry, this approach cannot capture.** A `swift_version` meaningfully different from the Xcode version would be the *language mode* — the per-target `SWIFT_VERSION` build setting, 5 vs 6. That would answer a real question ("if we adopt Swift 6 strict concurrency in our public API, do customers still building in Swift 5 mode break?"). But language mode is not in the host `Info.plist` at all, and an approximation derived from the compiler reports the wrong thing for exactly that case. A field that looks like it answers the question and does not is worse than an absent field.

**3. The iOS compatibility cliff is not shaped like Kotlin's.** On Android, `kotlin_version` gates a real hazard: Kotlin metadata is forward-incompatible, so a library compiled with Kotlin 2.x cannot be consumed by a 1.9 compiler, making the customer's Kotlin version a hard blocker on bumping ours. iOS has no equivalent, because we build every target with `BUILD_LIBRARY_FOR_DISTRIBUTION = YES` and therefore ship a module-stable `.swiftinterface`, which newer compilers consume fine. The practical gate is "customer's Xcode >= ours", and `xcode_version` answers that directly.

Worth noting *why* Android can do this and we cannot: `KotlinVersion.CURRENT` works because `kotlin-stdlib` is an ordinary dependency bundled into the APK, so the version travels with the app as readable data. Swift's runtime has been ABI-stable and OS-provided since Swift 5, so there is no app-bundled artifact carrying a language version — anything readable at runtime describes the OS's Swift runtime, which already tracks `os.version`.

`kotlinVersion` likewise stays `nil`: the protocol scopes it to a Kotlin *host app*, which iOS is not. The shared module's own provenance is already on `ossdk.kmp_version`.

## Test plan

- [x] `OneSignalOSCoreTests` green against `main` with KMP at the pinned `v0.3.0`
- [x] New `OSLoggerHostBuildAttributesTests` covering `DTXcode` digit unpacking, key mapping including `minimum_os_version` / `minimum_macos_version`, and omission of blank/missing keys
- [x] `swiftlint` clean on the changed source file
- [x] KMP XCFramework rebuilt at the pinned submodule commit
- [ ] Confirm the new `ossdk.*` attributes land in GCP Logs Explorer on a real device build

Verify on device with:

```
resource.type="k8s_container"
labels.log_source="sdk_log"
jsonPayload."ossdk.app_id"="<app id>"
```

Related: [SDK-5106](https://linear.app/onesignal/issue/SDK-5106/report-host-xcode-version-and-deployment-target-on-ios-remote-logs), [SDK-4999](https://linear.app/onesignal/issue/SDK-4999/add-kotlinswift-language-versions-to-remote-log-resource-attributes) (KMP contract), [SDK-5081](https://linear.app/onesignal/issue/SDK-5081/bug-bash-ios-remote-logging-turbine-feature-flags-kmp) (bug bash covering remote logging + feature flags).

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
@@ -51,6 +51,9 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
         static let deviceManufacturer = "Apple"
         static let unknown = "unknown"
         static let osBuildName = "kern.osversion"
+        static let xcodeVersionInfoKey = "DTXcode"
+        static let xcodeVersionAttribute = "xcode_version"
+        static let macCatalystAttribute = "apple_platform"
         static let disabledLogLevel = "NONE"
         static let crashDirectoryComponents = ["onesignal", "logger", "crashes"]
 
@@ -107,15 +110,20 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
     let osBuildId = OSLoggerPlatformProvider.systemValue(named: Constants.osBuildName)
     let sdkWrapper = OneSignalWrapper.sdkType
     let sdkWrapperVersion = OneSignalWrapper.sdkVersion
+    /// Nil on iOS: this describes the *host app's* Kotlin stack, and an iOS app is
+    /// not a Kotlin host. The shared module's own provenance rides on
+    /// `ossdk.kmp_version` instead.
     let kotlinVersion: String? = nil
-    let swiftVersion: String? = nil
-    let additionalVersionAttributes: [String: String] = {
-        #if targetEnvironment(macCatalyst)
-        return ["apple_platform": "mac_catalyst"]
-        #else
-        return [:]
-        #endif
-    }()
+
+    /// Approximated from the host app's Xcode version. Apple ships one Swift
+    /// toolchain per Xcode release and exposes no runtime API for the language
+    /// version, so this is the closest signal available. `xcode_version` is emitted
+    /// alongside and is exact, so a stale row in the lookup table stays recoverable.
+    let swiftVersion: String? = OSLoggerPlatformProvider.hostXcodeVersion()
+        .flatMap(OSLoggerPlatformProvider.approximateSwiftVersion(forXcodeVersion:))
+
+    let additionalVersionAttributes: [String: String] =
+        OSLoggerPlatformProvider.hostBuildAttributes()
     var enabledFeatureFlags: [String] {
         featureFlagsProvider()
     }
@@ -188,6 +196,89 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
     }
 
     let apiBaseUrl = OS_API_SERVER_URL.trimmingCharacters(in: CharacterSet(charactersIn: "/"))
+
+    /// Xcode stamps its own version, the compiler, and the SDK it built against
+    /// into the *host app's* `Info.plist`. Reading `Bundle.main` is deliberate:
+    /// this SDK ships as a prebuilt XCFramework, so a compile-time check here would
+    /// describe OneSignal's build machine and be identical for every customer.
+    ///
+    /// These are build-time facts. The OS actually running is reported separately
+    /// as `os.name` / `os.version` / `os.build_id`.
+    private static let buildMetadataKeys: [(infoKey: String, attribute: String)] = [
+        ("DTXcodeBuild", "xcode_build"),
+        ("DTCompiler", "build_compiler"),
+        ("DTPlatformName", "build_platform_name"),
+        ("DTPlatformVersion", "build_platform_version"),
+        ("DTPlatformBuild", "build_platform_build"),
+        ("DTSDKName", "build_sdk_name"),
+        ("DTSDKBuild", "build_sdk_build")
+    ]
+
+    /// Floor lookup, so an Xcode newer than every row resolves to the most recent
+    /// known Swift release rather than dropping the attribute. Add a row whenever
+    /// an Xcode release ships a new Swift version.
+    private static let swiftVersionByXcode: [(xcode: (major: Int, minor: Int), swift: String)] = [
+        ((14, 0), "5.7"),
+        ((14, 3), "5.8"),
+        ((15, 0), "5.9"),
+        ((15, 3), "5.10"),
+        ((16, 0), "6.0"),
+        ((16, 3), "6.1"),
+        ((26, 0), "6.2")
+    ]
+
+    static func hostBuildAttributes(
+        infoDictionary: [String: Any]? = Bundle.main.infoDictionary
+    ) -> [String: String] {
+        var attributes: [String: String] = [:]
+        #if targetEnvironment(macCatalyst)
+        attributes[Constants.macCatalystAttribute] = "mac_catalyst"
+        #endif
+        if let xcodeVersion = hostXcodeVersion(infoDictionary: infoDictionary) {
+            attributes[Constants.xcodeVersionAttribute] = xcodeVersion
+        }
+        for (infoKey, attribute) in buildMetadataKeys {
+            guard let value = infoDictionary?[infoKey] as? String, !value.isEmpty else {
+                continue
+            }
+            attributes[attribute] = value
+        }
+        return attributes
+    }
+
+    static func hostXcodeVersion(
+        infoDictionary: [String: Any]? = Bundle.main.infoDictionary
+    ) -> String? {
+        guard let raw = infoDictionary?[Constants.xcodeVersionInfoKey] as? String else {
+            return nil
+        }
+        return decodeXcodeVersion(raw)
+    }
+
+    /// `DTXcode` packs major/minor/patch into digits: `"2620"` is 26.2, `"0900"` is 9.0.
+    static func decodeXcodeVersion(_ raw: String) -> String? {
+        guard let packed = Int(raw.trimmingCharacters(in: .whitespaces)), packed > 0 else {
+            return nil
+        }
+        let major = packed / 100
+        let minor = (packed / 10) % 10
+        let patch = packed % 10
+        return patch == 0 ? "\(major).\(minor)" : "\(major).\(minor).\(patch)"
+    }
+
+    static func approximateSwiftVersion(forXcodeVersion version: String) -> String? {
+        let components = version.split(separator: ".").compactMap { Int($0) }
+        guard let major = components.first else {
+            return nil
+        }
+        let minor = components.count > 1 ? components[1] : 0
+        var resolved: String?
+        for entry in swiftVersionByXcode
+        where (entry.xcode.major, entry.xcode.minor) <= (major, minor) {
+            resolved = entry.swift
+        }
+        return resolved
+    }
 
     private static func systemValue(named name: String) -> String {
         var size = 0
--- a/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
@@ -55,10 +55,12 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
         static let xcodeVersionAttribute = "xcode_version"
         static let xcodeBuildInfoKey = "DTXcodeBuild"
         static let xcodeBuildAttribute = "xcode_build"
-        /// Catalyst bundles declare their floor as `LSMinimumSystemVersion`; iOS uses
-        /// `MinimumOSVersion`. First match wins.
-        static let minimumOSVersionInfoKeys = ["MinimumOSVersion", "LSMinimumSystemVersion"]
+        /// iOS / tvOS / watchOS deployment target. Distinct from the macOS floor:
+        /// `LSMinimumSystemVersion` is a macOS version and must not share this key.
+        static let minimumOSVersionInfoKey = "MinimumOSVersion"
         static let minimumOSVersionAttribute = "minimum_os_version"
+        static let minimumMacOSVersionInfoKey = "LSMinimumSystemVersion"
+        static let minimumMacOSVersionAttribute = "minimum_macos_version"
         static let applePlatformAttribute = "apple_platform"
         static let disabledLogLevel = "NONE"
         static let crashDirectoryComponents = ["onesignal", "logger", "crashes"]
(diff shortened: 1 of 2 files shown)


iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
```diff
@@ -107,15 +110,21 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
...
+    /// `ossdk.kmp_version` instead.
     let kotlinVersion: String? = nil
+
+    /// Nil on iOS. Apple exposes no runtime API for the Swift language version and
+    /// there is no `DTSwiftVersion` key, so the only derivable value would come from
+    /// the Xcode version — which maps to a Swift compiler 1:1 and is already emitted
+    /// exactly as `xcode_version`. The one thing a distinct value could carry is the
+    /// per-target language mode (`SWIFT_VERSION`, 5 vs 6), and that is not in the
+    /// host's `Info.plist` at all.
     let swiftVersion: String? = nil
```
abdulraqeeb33, line 125: i will remove this swiftVersion in another pr

fadi-george:
Lets see what nan says but maybe we can drop DTXcodeBuild?
build_sdk_name seems to just be build_platform_name + build_platform_version, could drop build_sdk_name or just build_platform_name + build_platform_version.
build_platform_build seems to most often same as build_sdk_build so maybe just keep build_sdk_build.

fadi-george:
Could we fall back to `LSMinimumSystemVersion` when `MinimumOSVersion` is absent? Catalyst bundles use that key, so `minimum_os_version` is currently omitted for Catalyst hosts.

```swift
for key in ["MinimumOSVersion", "LSMinimumSystemVersion"] {
    guard let value = infoDictionary?[key] as? String, !value.isEmpty else {
        continue
    }
    attributes["minimum_os_version"] = value
    break
}
```

fadi-george:
`Bundle.main` refers to the extension bundle when this runs inside an NSE, not the containing app. Should these attributes describe the current executable instead? If so, could we update the “host app” wording? Otherwise, we need to locate the enclosing `.app` bundle before reading its `Info.plist`.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
```diff
@@ -189,6 +198,67 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
...
+    /// the *host app's* `Info.plist`, alongside the deployment target the app
+    /// declares. Reading `Bundle.main` is deliberate: this SDK ships as a prebuilt
+    /// XCFramework, so a compile-time check here would describe OneSignal's build
+    /// machine and be identical for every customer.
+    ///
+    /// These are build-time facts and answer what the host *commits to* supporting.
+    /// The OS actually running is reported separately as `os.name` / `os.version` /
+    /// `os.build_id`, and the two diverge: an app can serve only iOS 18 users while
+    /// still declaring a much older `minimum_os_version`, which is what constrains
+    /// raising our own deployment target.
```
nan-li: nit do we need 10 lines to describe this?

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
```diff
@@ -51,6 +51,9 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
         static let deviceManufacturer = "Apple"
         static let unknown = "unknown"
         static let osBuildName = "kern.osversion"
+        static let xcodeVersionInfoKey = "DTXcode"
+        static let xcodeVersionAttribute = "xcode_version"
+        static let macCatalystAttribute = "apple_platform"
```
nan-li: the name of this variable is a bit confusing. Can we be worded or something?

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
```diff
@@ -189,6 +198,67 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
...
+    /// raising our own deployment target.
+    private static let buildMetadataKeys: [(infoKey: String, attribute: String)] = [
+        ("DTXcodeBuild", "xcode_build"),
+        ("DTCompiler", "build_compiler"),
+        ("DTPlatformName", "build_platform_name"),
+        ("DTPlatformVersion", "build_platform_version"),
+        ("DTPlatformBuild", "build_platform_build"),
+        ("DTSDKName", "build_sdk_name"),
+        ("DTSDKBuild", "build_sdk_build"),
+        ("MinimumOSVersion", "minimum_os_version")
```
nan-li: yeah, agree with Fadi on including LSMinimumSystemVersion for catalyst

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
```diff
@@ -189,6 +198,67 @@ final class OSLoggerPlatformProvider: ILoggerPlatformProvider {
...
+    /// machine and be identical for every customer.
+    ///
+    /// These are build-time facts and answer what the host *commits to* supporting.
+    /// The OS actually running is reported separately as `os.name` / `os.version` /
+    /// `os.build_id`, and the two diverge: an app can serve only iOS 18 users while
+    /// still declaring a much older `minimum_os_version`, which is what constrains
+    /// raising our own deployment target.
+    private static let buildMetadataKeys: [(infoKey: String, attribute: String)] = [
+        ("DTXcodeBuild", "xcode_build"),
+        ("DTCompiler", "build_compiler"),
```
nan-li: recommend to drop this field as `DTCompiler` has been the constant com.apple.compilers.llvm.clang.1_0 on effectively every app built since about 2011

nan-li:
> `Bundle.main` refers to the extension bundle when this runs inside an NSE, not the containing app. Should these attributes describe the current executable instead? If so, could we update the “host app” wording? Otherwise, we need to locate the enclosing `.app` bundle before reading its `Info.plist`.

This is ok, nothing in the extension or outcomes reference remote logging. Even if we did it would be okay because the same xcode builds everything in general

nan-li:
> Lets see what nan says but maybe we can drop DTXcodeBuild? build_sdk_name seems to just be build_platform_name + build_platform_version, could drop build_sdk_name or just build_platform_name + build_platform_version. build_platform_build seems to most often same as build_sdk_build so maybe just keep build_sdk_build.

I agree on trimming this down and recommend we just add 3 new fields as everything else is just noise based on looking at the full picture of labels we get below. For example, `ossdk.build_platform_name: "iphonesimulator"` when we already have `device.model.identifier: "Simulator iPhone"`


Attribute | Verdict | Why
-- | -- | --
xcode_version | keep | The human-filterable toolchain field.
xcode_build | keep | Exact toolchain; separates beta from release; determines all SDK values below.
minimum_os_version | keep | The one fact nothing else carries, and the PR's stated point. Needs the LSMinimumSystemVersion fallback for Catalyst.
apple_platform | keep | Pre-existing Catalyst marker.


Here's a snippet of the payload
```
device.manufacturer: "Apple"
device.model.identifier: "Simulator iPhone"
os.build_id: "25F80"
os.name: "iOS"
os.version: "26.5"
ossdk.build_platform_build: "23F81a"
ossdk.build_platform_name: "iphonesimulator"
ossdk.build_platform_version: "26.5"
ossdk.build_sdk_build: "23F81a"
ossdk.build_sdk_name: "iphonesimulator26.5"
ossdk.install_id: "xxxxxxxxxxxxxxx"
ossdk.kmp_version: "v0.3.0"
ossdk.minimum_os_version: "16.0"
ossdk.sdk_base: "ios"
ossdk.sdk_base_version: "050506"
ossdk.xcode_build: "17F113"
ossdk.xcode_version: "26.6"
```

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
abdulraqeeb33: Trimmed. Now four lines, keeping only the part that is not evident from the code (why this is read at runtime rather than with a compile-time check).

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
abdulraqeeb33: Renamed `macCatalystAttribute` -> `applePlatformAttribute` so the constant matches the attribute name it holds.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
abdulraqeeb33: Added the fallback, using Fadi's ordering — `MinimumOSVersion` first, then `LSMinimumSystemVersion`. Covered by two tests: the Catalyst-only fallback, and that `MinimumOSVersion` wins when both are present.

iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/Logging/OSLoggerPlatformProvider.swift
abdulraqeeb33: Dropped `build_compiler`, and took the rest of the trim too — `build_platform_name`, `build_platform_version`, `build_platform_build`, `build_sdk_name`, `build_sdk_build`. Left with `xcode_version`, `xcode_build`, `minimum_os_version`, plus the pre-existing `apple_platform`.

abdulraqeeb33:
Pushed `45a0eac8` — took the full trim you both landed on.

**Kept (3 new + 1 pre-existing):**

| Attribute | Example |
| --- | --- |
| `xcode_version` | `26.6` |
| `xcode_build` | `17F113` |
| `minimum_os_version` | `16.0` |
| `apple_platform` | `mac_catalyst` (pre-existing) |

**Dropped:** `build_compiler`, `build_platform_name`, `build_platform_version`, `build_platform_build`, `build_sdk_name`, `build_sdk_build`.

The payload sample made the redundancy hard to argue with — `build_platform_build` and `build_sdk_build` were both `23F81a`, and `build_sdk_name` was literally `build_platform_name` + `build_platform_version` concatenated. Agreed on `build_platform_name` vs `device.model.identifier` too.

**Catalyst fallback** (@fadi-george): added with your ordering, `MinimumOSVersion` then `LSMinimumSystemVersion`. Two tests cover it — the Catalyst-only case, and that `MinimumOSVersion` wins when both keys are present.

**NSE / `Bundle.main`** (@fadi-george, @nan-li): taking Nan's call that this is fine functionally, so no code change. I did fix the wording — the doc comment now says "the running executable" rather than "the host app", since that is accurate in either context.

Also renamed `macCatalystAttribute` -> `applePlatformAttribute`, and cut the doc comments down.

`swiftVersion` stays `nil` for now; @abdulraqeeb33 is removing the property in a separate PR since it is a KMP protocol requirement rather than something iOS can drop unilaterally.

Tests: 107 passing, 0 failures, against the pinned KMP submodule. Net -47/+61 across the provider and its tests.

fadi-george:
Maybe a create a linear ticket for this and add to pr title.

abdulraqeeb33:
Created [SDK-5106](https://linear.app/onesignal/issue/SDK-5106/report-host-xcode-version-and-deployment-target-on-ios-remote-logs) and put it In Review. Title updated to match.

fadi-george:
`LSMinimumSystemVersion` is a macOS version, while `minimum_os_version` otherwise contains an iOS version. Could Catalyst use a separate `minimum_macos_version` attribute to avoid mixing version namespaces?

abdulraqeeb33:
Agreed — `LSMinimumSystemVersion` is a macOS version, so stuffing it into `minimum_os_version` would make "who is below iOS 15" queries include Catalyst apps pinned to macOS 13.

Pushed `70952b7d`: Catalyst (and any host with that plist key) now emits `ossdk.minimum_macos_version`. `minimum_os_version` is only `MinimumOSVersion`. Both can be present and are written independently.

`OSLoggerHostBuildAttributesTests` covers the Catalyst-only case and the both-keys case. [SDK-5106](https://linear.app/onesignal/issue/SDK-5106/report-host-xcode-version-and-deployment-target-on-ios-remote-logs) updated to match.

fadi-george (approved):

nan-li (approved):

PR #1731, opened 2026-09-01
fix: [SDK-5099] respect REST API-disabled push subscriptions

nan-li:
# Description
## One Line Summary
Stop re-enabling push subscriptions that the app owner disabled remotely, from the dashboard (`notification_types` -22) or through the REST API (-31).

### Motivation
Customers who suppress users by disabling subscriptions via the REST API see them come back subscribed. The SDK never hydrated the server's disable onto an existing push subscription (`parseFetchUserResponse` is gated on `subscriptionId == nil`), and every Create User and PATCH body recomputed `enabled` from device state, so a login or a routine device-metadata update silently re-enabled the subscription. A dashboard unsubscribe writes a different code and was undone the same way. Tracked internally as SDK-5099.

### Scope
Push subscriptions only; email/SMS hydration is unchanged. The model gains a server-owned `remoteDisabledReason` field that mirrors the server's `notification_types`, and `OSRemoteDisable` names the two codes the app owner sets remotely, -22 for a dashboard unsubscribe and -31 for a REST API disable. A response reporting either one records it verbatim, so the outgoing payload echoes back the code the server sent rather than collapsing both into one. Any other reported value clears it, and it also clears when the subscription ID resets (the record it describes is gone) or on an explicit `optIn()`, whose clear outranks stale in-flight hydration until the server confirms. That opt-in guard arms only when the `optIn()` changed something, either it cleared a recorded disable or lifted the user's own opt-out; a call that changes nothing sends nothing and must not blind the SDK to a later disable. While a code is recorded, Create User and PATCH payloads send `enabled: false` with that code instead of device values; both builders share one snapshot-based body (`updateParams()` / `outboundNotificationTypes`). Recording and clearing log at DEBUG.

Public `pushSubscription.optedIn` now reports false while a disable is recorded, and its doc comment says so. No signature changes. A remote disable suppresses delivery as surely as a missing permission or an opt-out, and it is the only one of the three an app has no other way to see, so a preference center no longer shows "subscribed" on a device the app owner turned off. Hydration therefore fires the push subscription observer with the new state, and suppresses the `enabled` delta a local change of the same shape would generate, since sending that would tell the server what it just told us.

### For reviewers
Two product calls, both matched by the companion PR in OneSignal-Android-SDK. A remote disable now flips the public `optedIn` to false, which reverses the earlier decision on this branch to keep that property to preference plus permission. And `optIn()` still overrides the suppression, on the reasoning that a deliberate user action beats an exclusion list.

# Testing
## Unit testing
Model-level tests cover code recognition, verbatim recording of each code and switching between them, mirror clearing, ID-reset clearing, `optIn()` precedence over stale hydration when it changed something and none when it didn't, email and SMS hydration staying off the push path, the reset event's previous state, payload override, NSCoding round-trip, device-state refresh survival, `optedIn` reporting false for both codes, `optedIn` ignoring device-recoverable codes, and hydration firing the observer without enqueueing an `enabled` delta. An integration test reproduces the customer scenario end to end: fetch reports the disable, login to a different external ID, and the Create User body still sends `enabled: false` with that code.

## Manual testing
Device tested before and after with REST API calls.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
@@ -114,6 +114,10 @@ class OSSubscriptionModel: OSModel {
         var deviceModel: String?
         var appVersion: String?
         var netType: Int?
+        var restApiDisabledReason: Int?
+        // Not persisted; an optIn() clear outranks stale hydration until the server reports
+        // the subscription in another state.
+        var restApiDisableClearedByUser = false
     }
 
     /**
@@ -182,6 +186,11 @@ class OSSubscriptionModel: OSModel {
                 return
             }
 
+            // The disable code describes a specific server record; the record is gone when the ID resets.
+            if newValue == nil {
+                restApiDisabledReason = nil
+            }
+
             // Cache the subscriptionId to UserDefaults for routine reads, and the OSResilientStorage mirror 
             OneSignalUserDefaults.initShared().saveString(forKey: OSUD_PUSH_SUBSCRIPTION_ID, withValue: newValue)
             OSResilientStorage.setString(newValue ?? "", forKey: OSResilientStorage.keySubscriptionId)
@@ -194,7 +203,12 @@ class OSSubscriptionModel: OSModel {
     var enabled: Bool { // Does not consider subscription_id in the calculation
         get {
             let state = snapshot()
-            return calculateIsEnabled(address: state.address, reachable: state.reachable, isDisabled: state.isDisabled)
+            return calculateIsEnabled(
+                address: state.address,
+                reachable: state.reachable,
+                isDisabled: state.isDisabled,
+                restApiDisabledReason: state.restApiDisabledReason
+            )
         }
     }
 
@@ -258,6 +272,28 @@ class OSSubscriptionModel: OSModel {
             }
             firePushSubscriptionChanged(.isDisabled(oldValue))
             notificationTypes = -2
+        }
+    }
+
+    /// The notification_types value for a REST API disable, the only server-owned code; other
+    /// negative codes are device or delivery errors the device recovers by re-asserting its state.
+    static let restApiDisabledNotificationType = -31
+
+    /**
+     The server's REST API disable code, or nil when the server has not disabled this subscription.
+     Hydrated from responses, never derived from device state, and echoed back in payloads so routine
+     updates and logins don't re-enable a suppressed subscription. Cleared when a response reports
+     any other state, when the subscription ID resets, or by `optIn()`.
+     */
+    var restApiDisabledReason: Int? {
+        get { stateLock.withLock { state.restApiDisabledReason } }
+        set {
+            let oldValue = swapValue(\.restApiDisabledReason, to: newValue)
+            guard newValue != oldValue else {
+                return
+            }
+            // Mirrors server state rather than a local change, so persist without generating a delta.
+            self.set(property: "restApiDisabledReason", newValue: newValue, preventServerUpdate: true)
         }
     }
 
@@ -371,7 +407,9 @@ class OSSubscriptionModel: OSModel {
             sdk: ONESIGNAL_VERSION,
             deviceModel: OSDeviceUtils.getDeviceVariant(),
             appVersion: Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String,
-            netType: OSNetworkingUtils.getNetType() as? Int
+            netType: OSNetworkingUtils.getNetType() as? Int,
+            restApiDisabledReason: nil,
+            restApiDisableClearedByUser: false
         )
 
         super.init(changeNotifier: changeNotifier)
@@ -393,6 +431,7 @@ class OSSubscriptionModel: OSModel {
         coder.encode(state.deviceModel, forKey: "deviceModel")
         coder.encode(state.appVersion, forKey: "appVersion")
         coder.encode(state.netType, forKey: "netType")
+        coder.encode(state.restApiDisabledReason, forKey: "restApiDisabledReason")
     }
 
     required init?(coder: NSCoder) {
@@ -415,7 +454,9 @@ class OSSubscriptionModel: OSModel {
             sdk: coder.decodeObject(forKey: "sdk") as? String ?? ONESIGNAL_VERSION,
             deviceModel: coder.decodeObject(forKey: "deviceModel") as? String,
             appVersion: coder.decodeObject(forKey: "appVersion") as? String,
-            netType: coder.decodeObject(forKey: "netType") as? Int
+            netType: coder.decodeObject(forKey: "netType") as? Int,
+            restApiDisabledReason: coder.decodeObject(forKey: "restApiDisabledReason") as? Int,
+            restApiDisableClearedByUser: false
         )
 
         super.init(coder: coder)
@@ -436,18 +477,43 @@ class OSSubscriptionModel: OSModel {
                 // self.address = property.value as? String
             case "enabled":
                 if let enabled = property.value as? Bool {
-                    if self.enabled != enabled { // TODO: Is this right?
-                        _isDisabled = !enabled
-                    }
+                    hydrateEnabled(enabled, response: response)
                 }
             case "notification_types":
                 if let notificationTypes = property.value as? Int {
-                    self.notificationTypes = notificationTypes
+                    hydrateNotificationTypes(notificationTypes)
                 }
             default:
                 OneSignalLog.onesignalLog(.LL_DEBUG, message: "Unused property on subscription model")
             }
         }
+    }
+
+    /// Applies a hydrated `enabled`. A REST API disable is server-owned, not a user opt-out,
+    /// so it must not flip `_isDisabled`; the notification_types hydration records it instead.
+    private func hydrateEnabled(_ enabled: Bool, response: [String: Any]) {
+        guard !isRestApiDisable(response) else {
+            return
+        }
+        if self.enabled != enabled { // TODO: Is this right?
+            _isDisabled = !enabled
+        }
+    }
+
+    /// Routes a hydrated notification_types: -31 records the server's disable; any other value
+    /// clears it and becomes the device value.
+    private func hydrateNotificationTypes(_ value: Int) {
+        if value == Self.restApiDisabledNotificationType {
+            recordRestApiDisable(value)
+        } else {
+            acceptServerNonDisabledState()
+            self.notificationTypes = value
+        }
+    }
+
+    /// True when the response's notification_types carries a REST API disable code.
+    private func isRestApiDisable(_ response: [String: Any]) -> Bool {
+        return response["notification_types"] as? Int == Self.restApiDisabledNotificationType
     }
 
     // Using snake_case so we can use this in request bodies
@@ -457,19 +523,22 @@ class OSSubscriptionModel: OSModel {
         json["id"] = state.subscriptionId
         json["type"] = state.type.rawValue
         json["token"] = state.address
-        json["enabled"] = calculateIsEnabled(address: state.address, reachable: state.reachable, isDisabled: state.isDisabled)
+        json["enabled"] = calculateIsEnabled(
+            address: state.address,
+            reachable: state.reachable,
+            isDisabled: state.isDisabled,
+            restApiDisabledReason: state.restApiDisabledReason
+        )
         json["test_type"] = state.testType
         json["device_os"] = state.deviceOs
         json["sdk"] = state.sdk
         json["device_model"] = state.deviceModel
         json["app_version"] = state.appVersion
         json["net_type"] = state.netType
-        // notificationTypes defaults to -1 instead of nil, don't send if it's -1
-        if state.notificationTypes != -1 {
-            json["notification_types"] = state.notificationTypes
-        }
+        json["notification_types"] = outboundNotificationTypes(state)
         return json
     }
+
 }
 
 // Push Subscription related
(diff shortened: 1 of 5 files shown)


iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
```diff
@@ -491,14 +560,87 @@ extension OSSubscriptionModel {
...
+     */
+    func clearRestApiDisable() {
+        let oldValue = swapValue(\.restApiDisabledReason, to: nil)
+        guard oldValue != nil else {
+            return
+        }
+        stateLock.withLock { state.restApiDisableClearedByUser = true }
+        self.set(property: "restApiDisabledReason", newValue: nil as Int?, preventServerUpdate: true)
+        firePushSubscriptionChanged(.restApiDisabledReason(oldValue))
+    }
```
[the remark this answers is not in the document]
nan-li: **Declining**. Many apps call optIn on every launch. If every call counted as the user choosing to be on, those apps would ignore every disable the customer sends from the REST API, which is the exact bug we're fixing. So opt-in only counts as a choice when it actually changes something on the device.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
```diff
@@ -491,14 +560,87 @@ extension OSSubscriptionModel {
...
 
+    /// Records the server's disable code unless `optIn()` cleared one and the server has not yet
+    /// reported the subscription in another state; the user's explicit intent wins that race.
+    private func recordRestApiDisable(_ code: Int) {
+        let clearedByUser = stateLock.withLock { state.restApiDisableClearedByUser }
+        guard !clearedByUser else {
+            return
+        }
+        restApiDisabledReason = code
+    }
```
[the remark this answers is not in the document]
nan-li: Fair, fixed in 0de021e0f. Each change now checks and writes inside one lock, then fires its events after letting go of it.

nan-li:
> ## Multi-model review (Opus 5, GPT 5.6 Sol, Grok 4.6)
> The mirror + snapshot `updateParams()` / `jsonRepresentation()` is the right shape for SDK-5099. `optedIn` staying device-owned matches the stated product call.
> 
> **Act on** (3/3)
> 
> * `clearRestApiDisable()` returns before setting `restApiDisableClearedByUser` when the mirror is already nil. `optIn()` then loses the race against a first `-31` fetch (cold start after an operator disable, or after a subscription-ID reset). Tests only cover opt-in after the disable was already recorded.
> * `recordRestApiDisable` / `clearRestApiDisable` read and write the latch and reason across separate lock acquisitions, so a concurrent hydrate can still write `-31` after opt-in.
> 
> **Consider** (2/3)
> 
> * Cached `start()` / `startNewSession()` flush a subscription PATCH _before_ fetch hydrates `-31`. The first session after a REST disable can still send `enabled: true` — the customer path this PR names.
> * Existing-ID hydration matches only `id`. A miss plus new-session “sub gone” (`subscriptionId = nil` clears the mirror and POSTs a new enabled sub) can re-enable.
> * The latch is process-lifetime and generation-less: a later legitimate `-31` can be ignored until restart, and a stale `-31` after a newer non-disable can re-record.
> 
> **Dismissed:** `optedIn` not flipping is the stated product choice. NSCoding `decodeObject as? Int` matches `testType`/`netType` and is covered. Assigning `nil` `notification_types` omits the key rather than inserting `NSNull`.
> 
> [Open in Web](https://cursor.com/agents/bc-199f6429-ceba-411a-8d6e-cc166dadc72c?cursor_ref=pr_footer&cursor_cta=open_in_web) [View Automation](https://cursor.com/automations/d8c15061-9ff2-11f1-b532-320a589b8025) 
> Sent by Cursor Automation: PR Reviews

On the first launch after upgrading, if the SDK sends a routine device update before it has asked the server about the user, it can turn the subscription back on once. We accept that. It happens at most once and the next daily check fixes it. If the server's subscription ID doesn't match ours, we treat the old record as gone and make a new enabled one, which is intended. The "user opted in" marker is wiped as soon as the server reports anything other than a disable, so it can't stick around long enough to cause the problem described.

nan-li:
This PR fixes the coordination between "the server switched this subscription off through the API" and local SDK state. Opting in and out can happen on the SDK and enabling and disabling can happen on the REST API.

Almost every review comment is about a moment where this server state is missing, stale, or ignored. The sync between SDK and server is a fundamental inherent problem regardless of this PR. It will never be completely right for all. Consider that if someone adds a tag or calls optIn() and there is no connection. 30 minutes later, they update something via the REST API. Later, the SDK gets a connection 8 hours later, it will send those updates even though it is technically stale.

fadi-george:
Thanks. Items 3 and 4 look resolved. I think 1 and 2 still need changes:

1. `settleOptInGuard()` clears the flag after any completed subscription write, not specifically the opt-in write. An older in-flight update could clear a newly armed guard before the opt-in PATCH executes.

2. The NORMAL correction does not address the actual session-start ordering. OperationRepo executes the metadata PATCH before RefreshUser, so the GET sees the subscription already enabled and never generates the correction. The test manually constructs a stale-plus-corrective batch, but that batch is not produced on this path.

Could we add an OperationRepo-level test covering metadata update followed by RefreshUser and address that ordering directly?

nan-li:
> Thanks. Items 3 and 4 look resolved. I think 1 and 2 still need changes:
> 
> 1. `settleOptInGuard()` clears the flag after any completed subscription write, not specifically the opt-in write. An older in-flight update could clear a newly armed guard before the opt-in PATCH executes.
> 2. The NORMAL correction does not address the actual session-start ordering. OperationRepo executes the metadata PATCH before RefreshUser, so the GET sees the subscription already enabled and never generates the correction. The test manually constructs a stale-plus-corrective batch, but that batch is not produced on this path.
> 
> Could we add an OperationRepo-level test covering metadata update followed by RefreshUser and address that ordering directly?

I think this is meant for Android PR right? I answered it there: https://github.com/OneSignal/OneSignal-Android-SDK/pull/2728#issuecomment-5625816202

fadi-george:
Potential issues:
- `OSSubscriptionModel.hydrateNotificationTypes` handles `-22` and `-31` without checking `type == .push`. Email and SMS models are also hydrated through this path, so one of these values can set `remoteDisabledReason` and fire `onPushSubscriptionDidChange` with an email address or phone number as the push token. Please restrict remote-disable handling to push models and add an email/SMS regression test.

- `clearRemoteDisable()` sets `remoteDisableClearedByUser` on every `optIn()`, including idempotent calls when no disable is recorded. The flag is only cleared after fetching a non-disabled state. If an operator applies a legitimate remote disable before that happens, hydration ignores it and a later PATCH can re-enable the subscription. Please scope the guard to server responses that were already in flight when `optIn()` occurred, such as with a request generation or timestamp, and test a remote disable arriving after `optIn()`.

- (non-blocking) When `subscriptionId` becomes nil, `remoteDisabledReason` is cleared before `firePushSubscriptionChanged(.subscriptionId(...))` reconstructs the previous state. A remotely disabled subscription therefore emits `previous.optedIn == true` instead of false. Please preserve the old disable reason when constructing the observer event and test the previous/current values.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
```diff
@@ -436,40 +503,68 @@ class OSSubscriptionModel: OSModel {
...
+            return
+        }
+        if self.enabled != enabled { // TODO: Is this right?
+            _isDisabled = !enabled
+        }
+    }
+
+    /// Routes a hydrated notification_types: -22 and -31 record the server's disable verbatim, so
+    /// the recorded code stays distinguishable; any other value clears it and becomes the device value.
+    private func hydrateNotificationTypes(_ value: Int) {
```
abdulraqeeb33: `hydrateRemoteDisableState` is push-only. `hydrateModel` is not, and email/SMS still go through it. A `-22`/`-31` on an email row records `remoteDisabledReason` and fires the push observer with the email id and address. Every other mutator on this model already returns early for non-push. Guard `type == .push` here too.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
```diff
@@ -479,26 +574,141 @@ extension OSSubscriptionModel {
...
+        } else {
+            acceptServerNonDisabledState()
+        }
+    }
+
+    /**
+     Clears a remote disable and enqueues an enabled-change delta so the server re-enables the
+     subscription. Called from `optIn()`, where a deliberate user action overrides the suppression.
+     */
+    func clearRemoteDisable() {
```
abdulraqeeb33, line 710: This arms the guard on every `optIn()` and only disarms when a fetch reports a non-disable. Android settles when the write goes out. As written, `optIn()` on launch ignores `-22`/`-31` for the rest of the process, which is the customer path this PR is fixing.

abdulraqeeb33 (changes requested):
Two issues on the hydration / opt-in path.

1. Email/SMS hydrate records a remote disable and fires the push observer.
2. The opt-in guard never settles, so `optIn()` on launch ignores `-22`/`-31` for the process.

nan-li:
> Potential issues:
> 
> * `OSSubscriptionModel.hydrateNotificationTypes` handles `-22` and `-31` without checking `type == .push`. Email and SMS models are also hydrated through this path, so one of these values can set `remoteDisabledReason` and fire `onPushSubscriptionDidChange` with an email address or phone number as the push token. Please restrict remote-disable handling to push models and add an email/SMS regression test.
> * `clearRemoteDisable()` sets `remoteDisableClearedByUser` on every `optIn()`, including idempotent calls when no disable is recorded. The flag is only cleared after fetching a non-disabled state. If an operator applies a legitimate remote disable before that happens, hydration ignores it and a later PATCH can re-enable the subscription. Please scope the guard to server responses that were already in flight when `optIn()` occurred, such as with a request generation or timestamp, and test a remote disable arriving after `optIn()`.
> * (non-blocking) When `subscriptionId` becomes nil, `remoteDisabledReason` is cleared before `firePushSubscriptionChanged(.subscriptionId(...))` reconstructs the previous state. A remotely disabled subscription therefore emits `previous.optedIn == true` instead of false. Please preserve the old disable reason when constructing the observer event and test the previous/current values.


thanks, All three fixed in b31ea63c0.
1. Remote disable handling is push-only now, so email and SMS keep their old enabled mapping. Test covers both types with both codes and checks push observers hear nothing.
2. The flag is set only when optIn() actually changed something, since a call that changes nothing sends nothing and setting it anyway would ignore every later disable until the process died. Tests cover an opted-out user opting in during a fetch that reports the first disable, and a no-op optIn() followed by a real disable.
3. The old code now travels with the reset event, so previous reads as disabled and current as opted in. The reset no longer queues an enabled delta either, since the create request carries the new record's state.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
nan-li: thanks, this is the same callout made by Fadi, so fixed in b31ea63c0

iOS_SDK/OneSignalSDK/OneSignalUser/Source/OSSubscriptionModel.swift
nan-li: thanks, this is the same callout made by Fadi, so fixed in https://github.com/OneSignal/OneSignal-iOS-SDK/commit/b31ea63c0f20afc3fc59e61c9e459fc55c432e31

fadi-george:
Still some issues:
- The opt-in guard is still open-ended after a real opt-in. `remoteDisableClearedByUser` remains set until a fetch reports a non-disabled state, and PATCH success does not clear it. If an operator disables the subscription after the opt-in but before that fetch, the legitimate `-22/-31` is ignored and a later routine PATCH can re-enable it. The guard should apply only to fetch responses already in flight when `optIn()` was called.

- The new condition only treats `_isDisabled == true` as a meaningful opt-in. If `_isDisabled == false` but notification permission is unavailable, `optIn()` still initiates a real permission recovery. The guard is not armed, so a stale fetch can record the remote disable before permission succeeds and prevent the explicit opt-in from enabling the subscription. Please include this state in the guard logic and regression tests.

abdulraqeeb33 (approved):
Both comments addressed in b31ea63c0.

1. Email/SMS hydrate is push-only.
2. No-op `optIn()` no longer arms the guard.

Remaining: the guard still only drops on a non-disable fetch, and `userWasOptedOut` can arm with no write when there is no token or permission. Neither blocks.

nan-li:
> Still some issues:
> 
> * The opt-in guard is still open-ended after a real opt-in. `remoteDisableClearedByUser` remains set until a fetch reports a non-disabled state, and PATCH success does not clear it. If an operator disables the subscription after the opt-in but before that fetch, the legitimate `-22/-31` is ignored and a later routine PATCH can re-enable it. The guard should apply only to fetch responses already in flight when `optIn()` was called.
> * The new condition only treats `_isDisabled == true` as a meaningful opt-in. If `_isDisabled == false` but notification permission is unavailable, `optIn()` still initiates a real permission recovery. The guard is not armed, so a stale fetch can record the remote disable before permission succeeds and prevent the explicit opt-in from enabling the subscription. Please include this state in the guard logic and regression tests.

1. This one I think to leave as is. It needs three things in one process: a real opt-in, an operator disable landing before the next fetch, and a device change after that fetch. Cold starts clear the flag, so that last one is rare. Same answer as on Android, and I'd rather keep both platforms on one rule than stamp every fetch and update for that window. Do you feel like this one is blocking?

2. This second one is fair so fixed in 85764c4f0. A missing permission now counts as a real opt-in, since the prompt's grant is the write that re-enables. Test added.

fadi-george:
Yeah #1 is not really blocking.

fadi-george (approved):

PR #1733, opened 2026-09-04
chore: pin the KMP submodule and update

nan-li:
## One Line Summary
Pins the KMP submodule to KMP `main` (`bfc679a`, the merged crash-retention work) and adopts the two `CrashRetention` signatures that came with it.

## Testing
OneSignalOSCore test bundle (135 tests) passes against the rebuilt framework; the `OneSignalFramework` scheme builds.


nan-li:
closed for newer #1734

PR #1726, opened 2026-08-25
chore: [SDK-5084] remove unsupported iOS compatibility code

fadi-george:
# Description
## One Line Summary
Remove obsolete compatibility code for iOS versions below the SDK's iOS 12 minimum.

### Motivation
The SDK no longer supports iOS versions below 12, leaving runtime checks, fallbacks, selectors, constants, tests, and comments that cannot apply to supported devices.

### Scope
- Removes unreachable notification, IAM, location, jailbreak-detection, and app-termination compatibility paths.
- Removes unused iOS-version helpers, a no-op notification callback, and a deprecated Notification Service Extension overload.
- Removes an unreferenced legacy SwiftPM header while retaining the active SwiftPM package and supported NSE APIs.
- Flattens safe-area handling that is always available on supported versions.
- Stabilizes user-operation tests by explicitly controlling queue flushing and waiting for asynchronous executor cleanup.
- Retains legacy notification payload parsing because it is server-payload compatibility, not OS-version compatibility.

### Other
The removed Objective-C selectors were publicly visible but undocumented and unused internally. Distribution XCFramework headers are regenerated by the release process rather than edited manually.

# Testing
## Unit testing
Updated existing operation batching and subscription race tests to remove timing assumptions and prevent asynchronous work from leaking between tests.

## Manual testing
- Rebuilt the required OneSignalKMP XCFramework.
- Built the `UnitTestApp` test bundle for an iOS simulator.
- Ran `UnitTestApp_TestPlan_Reduced` successfully on an iOS simulator.
- Stress-ran the affected flaky tests successfully.
- Confirmed removed symbols have no remaining source references.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [x] Sessions
   - [x] In-App Messaging
   - [ ] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] I have filled out all **REQUIRED** sections above
   - [x] PR does one thing
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [x] I have included test coverage for these changes, or explained why they are not needed
   - [x] All automated tests pass, or I explained why that is not possible
   - [x] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [x] Code is as readable as possible.
   - [x] I have reviewed this PR myself, ensuring it meets each checklist item

Made with [Cursor](https://cursor.com)

PR #934, opened 2021-05-27
Randomly delay notification received calls to control traffic volume

emawby:
Notifications sent to a large number of subscribers can result in a very large number of confirmed deliveries hitting our backend at the same time. This PR randomly delays the confirmed delivery between 0 and 25 seconds. In order to do this we need have a blocking task on the NSE process until the confirmed delivery is sent. We don't want to delay the actual receipt of the notification, so this requires that OneSignal is passed the contentHandler from the NSE and calls it before waiting for the confirmed delivery to finish.

To achieve this I have created a new public method and marked the previous method as deprecated.
`+ (UNMutableNotificationContent*)didReceiveNotificationExtensionRequest:(UNNotificationRequest* _Nonnull)request withMutableNotificationContent:(UNMutableNotificationContent* _Nullable)replacementContent withContentHandler:(void (^)(UNNotificationContent * _Nullable))contentHandler;`

OneSignal will now call `contentHandler(replacementContent)` to trigger the notification display so this is no longer needed in the customer's NotificationService file. Any additional changes to the notification content will need to take place BEFORE calling OneSignal's `didReceiveNotificationExtensionRequest` method.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-ios-sdk/934)

PR #1737, opened 2026-09-08
fix: restore the deprecated 2-arg NSE selector

nan-li:
# Description
## One Line Summary
Put `didReceiveNotificationExtensionRequest:withMutableNotificationContent:` back on `OneSignal` and `OneSignalExtension` as a deprecated public wrapper.

### Motivation
5.6.0 removed this selector in [#1726](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1726) as unsupported iOS 10 code. The "iOS 10 only" comment meant NSE's floor is iOS 10, not that the method died on 11+. Wrappers and old `NotificationService` files still call the 2-arg form, so 5.6.0 is a compile break in a minor.

It was deprecated in 3.5.0 ([#934](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/934)) because confirmed-delivery spreading needs `contentHandler` to keep the NSE alive after the banner shows. The 2-arg path is still the sync contract (`nil` handler, no receipt delay). Deprecation is not the same as removal. Drop it in 6.0, not 5.6.x.

### Scope
- Public headers and `.m` wrappers on `OneSignal` and `OneSignalExtension` only
- Still deprecated, same message as before
- Forwards to the existing handler 2-arg method, which already calls the 3-arg path with `nil`
- No change to attachments, action buttons, or the contentHandler path
- XCFramework binaries are rebuilt at release

# Testing
## Unit testing
None added. The internal handler never lost the 2-arg method, and that is what these wrappers call. Existing NSE handler tests cover that path.

## Manual testing
Not device-tested. This restores a public selector so NSE files that still call it compile against 5.6.x. The 3-arg path is unchanged.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [x] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalExtension/OneSignalExtension.h
+++ b/iOS_SDK/OneSignalSDK/OneSignalExtension/OneSignalExtension.h
@@ -37,7 +37,6 @@
 #pragma clang diagnostic ignored "-Wnullability-completeness"
 // Call from a UNNotificationServiceExtension on iOS 10 and later.
 // Processes media attachments and action buttons.
-// The 2-arg form is the sync path (nil handler). Prefer the contentHandler overload.
 + (UNMutableNotificationContent*)didReceiveNotificationExtensionRequest:(UNNotificationRequest* _Nonnull)request withMutableNotificationContent:(UNMutableNotificationContent* _Nullable)replacementContent __deprecated_msg("Please use didReceiveNotificationExtensionRequest:withMutableNotificationContent:withContentHandler: instead.");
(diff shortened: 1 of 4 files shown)


fadi-george:
could add tests

fadi-george (approved):

PR #1739, opened 2026-09-10
feat: [SDK-5193] let an app opt out of subscription ID copy in Info.plist

nan-li:
# Description
## One Line Summary
Lets an app turn off subscription ID copy with an Info.plist key.

### Motivation
The gesture in #1730 is on by default, and the only switch is OneSignal's remote kill switch. An app that doesn't want a pasteboard write in production should be able to say so itself, in its own Info.plist, without a support ticket.

Ticket SDK-5193. Follows #1730, which merged as c23c6f99b; this PR targets main.

### Scope

`OneSignal_disable_subscription_id_copy` set to `YES` in Info.plist

- read like the SDK's older Info.plist switches (`OneSignal_disable_swizzling`, `OneSignal_suppress_launch_urls`, `OneSignal_disable_badge_clearing`): a Boolean or Number counts by value and a String through `NSString.boolValue`, so `<true/>` and `<string>YES</string>` both opt out. That matters because Xcode's plist editor stores a new row as a String unless its type is changed. A missing or mistyped value leaves the gesture on instead of crashing on `boolValue` the way the Objective-C reads would.
- checked before the observers register. With the key set the detector registers nothing, so nothing is counted, written or recorded.
- That is stricter than the remote kill switch, which still records `disabled` so we can count attempts; an app that opted out has said no.
- One INFO line at start says the key was honoured, so a developer can confirm it took. The started latch still flips, so the recovery path's second `start()` stays quiet.
- The key names the outcome, not the gesture, so the docs can call the feature one thing.
- No public API, so the wrapper SDKs expose it with no change.

### Other
The docs page and release notes need to say the feature is on by default and name the key.

# Testing
## Unit testing
`OSDeviceGestureDetectorTests` injects the opt-out and asserts, through the observer-counting center, that no observers register, that six cycles write nothing and record nothing, and that a second start registers nothing either. `OSDeviceGestureDetectorInfoPlistTests` deserialises each Info.plist value shape from its literal XML and checks the parse: `<true/>`, `<integer>1</integer>`, `<real>1.0</real>` and the strings `YES`, `yes`, `true` and `1` opt out; `<false/>`, `<integer>0</integer>` and the strings `NO`, `false`, `0` and empty do not; nil, array, dict, date and data values leave the gesture on. The parse tests fail against the previous `NSNumber`-only read. 187 OneSignalOSCore tests pass serially, SwiftLint clean apart from the pre-existing file length warning on the test file.

## Manual testing
Not run on a device. The read is the same `Bundle.main` lookup the other Info.plist switches use.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSDeviceGestureDetector.swift
+++ b/iOS_SDK/OneSignalSDK/OneSignalOSCore/Source/OSDeviceGestureDetector.swift
@@ -42,7 +42,9 @@ import UIKit
 /// at round trips of five seconds or faster.
 ///
 /// The `killSwitchKey` catalog flag turns the gesture off. Absent means enabled, so a device
-/// that has never fetched flags still has it.
+/// that has never fetched flags still has it. An app can also opt out for good with the
+/// Info.plist key `disableInfoPlistKey`; then the detector never starts, so nothing is counted
+/// or recorded.
 ///
 /// Every recognised gesture also records `OSObservabilityEvent.deviceGesture`, with its outcome
 /// and the copied ID, so the gesture's usage can be measured.
@@ -55,6 +57,7 @@ public final class OSDeviceGestureDetector: NSObject {
     static let minBackgroundDwellSeconds: TimeInterval = 0.25
 
     static let killSwitchKey = "sdk_device_gesture_disabled"
+    static let disableInfoPlistKey = "OneSignal_disable_subscription_id_copy"
 
     /// The copied ID expires after five minutes. Whatever it replaced is not restored.
     static let pasteboardExpirySeconds: TimeInterval = 300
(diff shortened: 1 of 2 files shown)


fadi-george:
Potential issues:
- The Info.plist value is cast only to `NSNumber`, so string values like `YES` silently fail and leave copying enabled. Could we support `NSString.boolValue` too and add coverage for string-backed values?

nan-li:
> Potential issues:
> 
> * The Info.plist value is cast only to `NSNumber`, so string values like `YES` silently fail and leave copying enabled. Could we support `NSString.boolValue` too and add coverage for string-backed values?

good find, in [1c7d2d4](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1739/commits/1c7d2d47fd8d0a7878b38a217f2016bf68a9cd6f) String now goes through boolValue like the other keys, with the plist-XML-driven tests in OSDeviceGestureDetectorInfoPlistTests.

fadi-george (approved):

fadi-george:
My bad, I accidentally merged.

nan-li:
> My bad, I accidentally merged.

No probs! thanks for approving

PR #1743, opened 2026-09-16
fix: [SDK-5137] Identity Verification config and token request follow-ups from the JWT review

nan-li:
# Description
## One Line Summary
Four follow-ups from the Identity Verification review, covering how the requirement is cached and how the app is asked for a token. Tracked on SDK-5137.

### Motivation
Review feedback on the Identity Verification stack was collected on [SDK-5137](https://linear.app/onesignal/issue/SDK-5137/ios-address-jwt-identity-verification-pr-review-feedback) instead of amending the stacked branches. This PR takes the items around the cached requirement and the token request to the app. #1740 covers the request queue items.

### Scope
Each fix is its own commit.

- Readers of the Identity Verification requirement no longer wait on a disk write. The cached value was written while the lock was held, so every reader across the SDK waited on that write. It now happens after the lock is released.
- Re-reading the cached requirement at startup now tells listeners about it. That re-read moved the requirement from unknown to the cached value silently, while the normal path runs a handler. Not a live bug today, since nothing listens at that point, but a trap for any later caller.
- The list of token listeners is created once at startup instead of on first use. Two threads could each create one, and the app's listener could land in the copy that was thrown away. A token request that no listener hears now logs a warning naming the user, and the public listener methods document that the listener is held weakly.
- Logging in again as the same user after a logout asks for a token again. The SDK asks once per user per session, and only a stored token cleared that. A user who was asked, never answered, logged out, and logged back in without a token was never asked again, so every request for that user stayed queued until the next launch. Each login that builds a new identity for the user now gets a fresh ask.

Follow-ups from review, each its own commit:

- Three comments that still described the token request as once per session now say that logging in as that user asks again. Comments only.
- The token request tests wait for the mock client to go idle before the next test starts, the same guard the lifecycle tests use. Test code only.
- The token request is cleared only when a login brings no usable token; a stored token already clears it, so a login with a token no longer clears it twice.
- The warning for a token request nobody hears fires only after a listener was registered. Before that it is a debug line, since registering a listener replays the request.
- The token request tests read only the log lines they are about, under a lock. Test code only.

Not changed: the public API surface, request signing, in-app messaging.

# Testing
## Unit testing
One or more tests per fix except the first, next to the code they cover. For the second and fourth fixes, the new test was run against the previous commit and failed there, then passed with the fix. The first moves a disk write out of a lock, which no test can observe deterministically, so it has none. The third has no old behavior a test can observe, so its tests cover the new behavior only. The token request tests live in a new file, `UserJwtAskTests`, because the existing lifecycle test class is at SwiftLint's size limit. The follow-ups add three tests there: a login with an empty token asks again, an unheard request before any listener is registered logs no warning, and one after the registered listener was released does. The second fails with the warning forced on.

## Manual testing
Not run on a device. The full unit test plan was run locally on an iPhone 17 Pro simulator.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


abdulraqeeb33 (approved):
Approve. Config and ask follow-ups match.

On merge with 1740, keep 1740's `parkedForToken`, not this PR's `awaitsToken`. Take this PR's `park` comment (rearm on login).

fadi-george:
Potential issues (not blocking):
- **P3:** `OneSignalUserManagerImpl+Jwt.swift:44` sets `everRegistered` before adding the observer. A concurrent JWT request between these operations logs a false warning claiming the listener was released. Add the observer before setting `everRegistered`, or synchronize both operations.
Not blocking. It only risks a rare misleading warning; JWT requests are still replayed and not lost.

nan-li:
> Potential issues (not blocking):
> 
> * **P3:** `OneSignalUserManagerImpl+Jwt.swift:44` sets `everRegistered` before adding the observer. A concurrent JWT request between these operations logs a false warning claiming the listener was released. Add the observer before setting `everRegistered`, or synchronize both operations.
>   Not blocking. It only risks a rare misleading warning; JWT requests are still replayed and not lost.

done

PR #1748, opened 2026-09-18
fix: [SDK-5181] accept a create-subscription success with no subscription object

nan-li:
# Description
## One Line Summary
Stop logging an error when the server answers a create-subscription request with a success that has no subscription object.

### Motivation
The server answers this way when the email or SMS being added already belongs to the user. The SDK logged it as an error on every occurrence even though nothing went wrong.

### Scope
- A success response without a subscription object is logged at INFO and the request is finished, not retried. The local model is left unchanged.
- The `ryw_token` and `ryw_delay` fields are read from the top level of the response, where the server puts them, instead of from inside the subscription object where they never were.
- Nothing sent to the server changes.

# Testing
## Unit testing
Two tests in `SubscriptionCreateResponseTests`: a success without a subscription object completes once without changing the model, and a success with one sets the subscription id and records the token and delay.

## Manual testing
Both tests plus `SubscriptionUpdateRaceTests` run locally on an iPhone 17 Pro simulator. 

Ran on device and added an existing email and got a 200 response and the SDK logged "INFO: Create Email subscription response has no subscription object to hydrate"

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


iOS_SDK/OneSignalSDK/OneSignalUserTests/Executors/SubscriptionCreateResponseTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        OneSignalLog.setLogLevel(.LL_VERBOSE)
+    }
+
+    override func tearDownWithError() throws { }
+
+    /**
+     The server answers a create for a subscription that already exists on the user with a 2xx and no
+     subscription object. The request is finished, not retried, and the model is left as it was.
+     */
+    func testResponseWithoutSubscription_completesWithoutHydratingOrRetrying() throws {
```
abdulraqeeb33: This still passes if the executor hunk is reverted. Parent already dequeued before the parse guard and left `subscriptionId` nil. Assert a top-level `ryw_token` is recorded (or `resolveConditions` when it is absent) so revert goes red.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSSubscriptionOperationExecutor.swift
```diff
@@ -315,7 +308,13 @@ class OSSubscriptionOperationExecutor: OSOperationExecutor {
                     }
                 }
 
-                request.subscriptionModel.hydrate(response)
+                // A 2xx with no subscription object is the server's no-op for a subscription that already exists on this user.
+                if let subscription = response?["subscription"] as? [String: Any] {
+                    request.subscriptionModel.hydrate(subscription)
+                } else {
+                    let type = request.subscriptionModel.type.rawValue
+                    OneSignalLog.onesignalLog(.LL_INFO, message: "Create \(type) subscription response has no subscription object to hydrate")
```
abdulraqeeb33, line 316: INFO is right for a missing `subscription` key. This also fires when the value is present but not a dict, and default remote logging will not export INFO. WARN/ERROR for that shape would keep a real payload bug visible.

abdulraqeeb33 (approved):
Approve with nits.

1. The no-subscription test still passes if the executor hunk is reverted. Assert RYW record/resolve so revert goes red.
2. INFO is right for a missing `subscription` key. Same line also fires when the value is present but not a dict, and default remote logging will not export INFO.

iOS_SDK/OneSignalSDK/OneSignalUserTests/Executors/SubscriptionCreateResponseTests.swift
nan-li: Done in a8296701a. The test now registers a never-met waiter on the IAM fetch condition before the request and asserts the executor's resolve releases it. With the executor reverted to main the test times out; with the fix it passes.

iOS_SDK/OneSignalSDK/OneSignalUser/Source/Executors/OSSubscriptionOperationExecutor.swift
nan-li: Skipping this one. The server either omits `subscription` or sends an object, so a present but non-object value cannot come from our API, and the point of the PR is to stop error logging this response. A missing key stays INFO.

PR #1749, opened 2026-09-18
chore: [SDK-5280] raise minimum iOS version to 15

nan-li:
# Description
## One Line Summary
Raise the minimum iOS version from 12 to 15. SDK-5280.

### Motivation
Xcode 27 refuses to build anything with a deployment target below iOS 15. That broke this project for local development and broke CocoaPods customers on Xcode 27, because CocoaPods copies the podspec's platform version onto the pod target. Swift Package Manager customers were not affected.

### Scope
- Every target in the SDK project and the example app now builds for iOS 15.
- Both podspecs and Package.swift declare iOS 15. Apps that target iOS 12 to 14 can no longer use the SDK, so the next release needs at least a minor version bump and a release note.
- README, migration guide and the bug report template say iOS 15.

# Testing
## Unit testing
No test changes.

## Manual testing
Built the test app on Xcode 27.0 with no command-line override and ran the OSCore tests and the subscription race tests on an iPhone 17 Pro simulator. 187 tests, all passing.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


fadi-george:
so we should update the wrappers too @nan-li , also is there no legacy code to remove 12-14?

fadi-george (approved):

abdulraqeeb33 (approved):
Do we have to update the documentation as well?

Note: Minimum iOS is different from DEPLOYMENT_TARGET

nan-li:
> so we should update the wrappers too @nan-li , also is there no legacy code to remove 12-14?

> Do we have to update the documentation as well?

Good point, I added a ticket to look at removing code and including wrappers too if there is code there.
Yes we will need to update wrappers to have minor bump and readme changes as part of that release. So it is clearer in history, I'll make those each PRs so they are easily seen when that cut is made.
And docs need updating.
Apps **must** target 15+ when submitting to app store starting April 2027 so if we have to keep supporting ios 5.6.x, it wont be indefinite.

> Note: Minimum iOS is different from DEPLOYMENT_TARGET

@abdulraqeeb33 Good call, we should be clearer with language. Unles you meant to call out something else, or just we should use consistent terminology?

abdulraqeeb33:
> @abdulraqeeb33 Good call, we should be clearer with language. Unles you meant to call out something else, or just we should use consistent terminology?

Just the language - I think when we update the docs, please keep that in mind.

